PR #45, opened 2022-07-16
set up CI

noot:
attempting to set up a CI for building + testing (and maybe fmt-ing :P)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      run: bash ./scripts/build.sh
+    - name: Run tests
+      run: bash ./scripts/test.sh
+
+  fmt:
+    name: fmt
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - run: rustup component add rustfmt
```
kayabaNerve: nightly fmt

scripts/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+./coins/monero/c/monero/build/release/bin/monerod --detach --regtest --offline --fixed-difficulty=1 --rpc-bind-ip 127.0.0.1 --rpc-bind-port 18081
+cargo test
```
kayabaNerve: `--all-features` ?

scripts/build.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+BASEDIR=$(pwd)
+git submodule update --init --recursive
+sudo apt-get install libgtest-dev && cd /usr/src/gtest && sudo cmake . && sudo make
```
kayabaNerve: I don't think this is necessary for Monero, just requested. We can slim this down a good bit (notably the following apt) by removing optionals.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,38 @@
+name: Tests
+
+on: [pull_request]
+
+jobs:
```
kayabaNerve, line 9: Also needs a cache.

https://github.com/MerosCrypto/asmr/blob/master/.github/workflows/ci.yaml#L54-L84 for an example

kayabaNerve (changes requested):

.github/workflows/tests.yml
noot: yep added

scripts/build.sh
noot: that would be nice, i just copy-pasted this from the Monero readme lol

silverpill: This is the list of dependencies I used for building the project on Debian:

```
apt install cmake libzmq3-dev libboost-dev libboost-system-dev libboost-filesystem-dev libboost-thread-dev libboost-date-time-dev libboost-chrono-dev libboost-regex-dev libboost-serialization-dev libboost-program-options-dev libboost-locale-dev
```

May be sufficient for Ubuntu too since it's Debian-based.

Also, last time I checked Serai required Rust 1.60 stable.

kayabaNerve: Thanks for chiming in with that :)

I'm actually not sure what version of Rust we require + why. It may be good to go through each crate, working that, and adding `msrv`s accordingly, working on finding earlier versions which are trivial to move to in order to bring wide support.

silverpill: I checked out the latest commit and there were 458 new crates to download :exploding_head: 
- `librocksdb-sys` requires `apt install clang`
- `ethereum-serai` requires solc binary (I think it can be installed from PPA on Ubuntu)
- `serai-runtime` requires rust-nightly (in build script)
- `monero-serai` fails with ```no method named `len` found for type `usize` in the current scope``` <del>(I'm not sure if it has anything to do with rust version)</del> https://github.com/serai-dex/serai/pull/51
- Apparently the rest of dependency tree still builds with 1.60, though I didn't get to the end of it

Rust 1.60 requirement probably comes from namespaced features in [some manifests](https://github.com/serai-dex/serai/blob/develop/crypto/transcript/Cargo.toml#L19) - that was the error I encountered when I tried to build project with 1.56 initially.

kayabaNerve: The new crates would be the mess that is Substrate.

solc should be from solc-select, which is a pip3.

monero-serai is absolutely my fault and dates back to when I cleaned up CLSAG a bit due to cargo fmt's opinion on it.

Anything wasm under Substrate will require nightly :/

Thanks for clarifying there. I forgot Rust only recently got sane feature naming...

scripts/test.sh
kayabaNerve: Looks like this isn't feasible because try-runtime doesn't work for the current Substrate runtime. One error seems to be a macro not realizing what it's doing, the others... I'm not sure about. We could remove try-runtime for now?

.github/workflows/tests.yml
```diff
@@ -10,6 +10,8 @@ jobs:
     - uses: actions/checkout@v1
     - name: Install nightly
       run: rustup toolchain install nightly
+    - name: Install WASM toolchain
+      run: rustup target add wasm32-unknown-unknown
```
kayabaNerve: stable rust, nightly wasm

Issue #52, opened 2022-07-18
Substrate tracked multisig


kayabaNerve:
Right now, there's an ink contract on https://github.com/serai-dex/serai/tree/ink-multisig offering confirmation for a multisig's validity and able to signal when to move it.

... this has no reason to be an ink! smart contract, other than to keep things consolidated in ink!. It's an incredibly simple piece of code though solely used to obtain BFT agreement on which multisig to move forward with and exactly when. I'd like to see this rewritten as a pallet.

kayabaNerve:
The validator-sets branch now implements voting on keys.

kayabaNerve:
Closable since #187.

PR #53, opened 2022-07-18
Validator Sets + ink! implementation of a multisig tracker

kayabaNerve:
https://github.com/serai-dex/serai/issues/52 is relevant as this is already potentially deprecated, yet I do believe this is still a solid first step.

Does document how the processor should react to this, and is tested on the success route.

cc @noot who wanted to leave comments on a PR.

cc @TheArchitect108 as it seems we may want this to be multi-multisig from the get go (just asking for your quick feedback on if we should do that here and now or...).

### DIFF
--- a/contracts/extension/src/lib.rs
+++ b/contracts/extension/src/lib.rs
@@ -3,22 +3,32 @@
 use ink_lang as ink;
 use ink_env::{Environment, DefaultEnvironment, AccountId};
 
+pub type Curve = u16;
+pub type Coin = u32;
+pub type GlobalValidatorSetId = u32;
+pub type ValidatorSetIndex = u8;
+pub type Key = Vec<u8>;
+
 #[ink::chain_extension]
 pub trait SeraiExtension {
   type ErrorCode = ();
 
-  /// Returns the amount of active validators on the current chain.
+  /// Returns the ID for the current global validator set.
   #[ink(extension = 0, handle_status = false, returns_result = false)]
-  fn active_validators_len() -> u16;
+  fn global_validator_set_id() -> GlobalValidatorSetId;
 
-  /// Returns the ID for the current validator set for the current chain.
-  // TODO: Decide if this should be an increasing unsigned integer instead of a hash.
+  /// Returns the amount of active validator sets within the global validator set.
   #[ink(extension = 1, handle_status = false, returns_result = false)]
-  fn validator_set_id() -> [u8; 32];
+  fn validator_sets() -> u8;
 
-  /// Returns if the specified account is an active validator for the current chain.
+  /// Returns the amount of key shares used within the specified validator set.
   #[ink(extension = 2, handle_status = false, returns_result = false)]
-  fn is_active_validator(account: &AccountId) -> bool;
+  fn validator_set_shares(set: ValidatorSetIndex) -> u16;
+
+  /// Returns the validator set the specified account is in, along with their amount of shares in
+  /// that validator set, if they are in a current validator
+  #[ink(extension = 3, handle_status = false, returns_result = false)]
+  fn active_validator(account: &AccountId) -> Option<(ValidatorSetIndex, u16)>;
 }
 
 pub struct SeraiEnvironment;
@@ -45,41 +55,57 @@ pub fn test_validators() -> Vec<AccountId> {
 }
 
 pub fn test_register() {
-  struct ExtensionLen;
-  impl ink_env::test::ChainExtension for ExtensionLen {
-    fn func_id(&self) -> u32 {
-      0
-    }
-
-    fn call(&mut self, _: &[u8], output: &mut Vec<u8>) -> u32 {
-      scale::Encode::encode_to(&u16::try_from(test_validators().len()).unwrap(), output);
-      0
-    }
-  }
-  ink_env::test::register_chain_extension(ExtensionLen);
-
   struct ExtensionId;
   impl ink_env::test::ChainExtension for ExtensionId {
     fn func_id(&self) -> u32 {
-      1
+      0
     }
 
     fn call(&mut self, _: &[u8], output: &mut Vec<u8>) -> u32 {
-      scale::Encode::encode_to(&[0xffu8; 32], output);
+      // Non-0 global validator set ID
+      scale::Encode::encode_to(&1u32, output);
       0
     }
   }
   ink_env::test::register_chain_extension(ExtensionId);
 
-  struct ExtensionActive;
-  impl ink_env::test::ChainExtension for ExtensionActive {
+  struct ExtensionSets;
+  impl ink_env::test::ChainExtension for ExtensionSets {
+    fn func_id(&self) -> u32 {
+      1
+    }
+
+    fn call(&mut self, _: &[u8], output: &mut Vec<u8>) -> u32 {
+      // 1 validator set
+      scale::Encode::encode_to(&1u8, output);
+      0
+    }
+  }
+  ink_env::test::register_chain_extension(ExtensionSets);
+
+  struct ExtensionShares;
+  impl ink_env::test::ChainExtension for ExtensionShares {
     fn func_id(&self) -> u32 {
       2
     }
 
+    fn call(&mut self, _: &[u8], output: &mut Vec<u8>) -> u32 {
+      // 1 key share per validator
+      scale::Encode::encode_to(&u16::try_from(test_validators().len()).unwrap(), output);
+      0
+    }
+  }
+  ink_env::test::register_chain_extension(ExtensionShares);
+
+  struct ExtensionActive;
+  impl ink_env::test::ChainExtension for ExtensionActive {
+    fn func_id(&self) -> u32 {
+      3
+    }
+
     fn call(&mut self, input: &[u8], output: &mut Vec<u8>) -> u32 {
       use scale::Decode;
-      let potential = AccountId::decode(&mut &input[1 ..]).unwrap(); // TODO: Why is this 1 ..?
+      let potential = AccountId::decode(&mut &input[1 ..]).unwrap(); // TODO: Why is this [1 ..]?
 
       let mut presence = false;
       for validator in test_validators() {
@@ -87,7 +113,8 @@ pub fn test_register() {
           presence = true;
         }
       }
-      scale::Encode::encode_to(&presence, output);
+      // Validator set 0, 1 key share
+      scale::Encode::encode_to(&Some((0u8, 1u16)).filter(|_| presence), output);
       0
     }
   }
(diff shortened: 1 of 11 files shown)


TheArchitect108:
Yeah might as well do it as a part of this PR.

kayabaNerve:
This then largely becomes merged with the pallet move, as we need to coordinate which validator is in which multisig which will be non-ink dependent. Immediately, to keep progress, it can be shimmed by adding yet another key + exposed function but...

kayabaNerve:
I'm running the benchmarks overnight (single-threaded, hence overnight). I believe 500 nodes would be just over a minute to key gen, and ~5s to sign (each) based on my initial math. Accordingly, a single multisig should be fine? The remaining question would be if/when we move specific coins to specific validators to scale better there. That's not directly incompatible with this design though, as BTC could be Secp256k1-0 while ETH would be Secp256k1-1. Regardless, I'd like to drop multiple multisigs from consideration overall for now, once my benchmark finishes and confirms my rough math.

kayabaNerve:
```
running 1 test
test tests::literal::kp256::secp256k1_non_ietf has been running for over 60 seconds
KeyGen: 7s per participant for 500 participants
Sign: 0s per participant for 500 participants
test tests::literal::kp256::secp256k1_non_ietf ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 3724.77s
```

I vote we drop all multi-multisig conversations until after launch.

EDIT: I divved the sign time by 500. This only had 334 parties signing as that was the threshold. Sign was still <0.7s per participant though.

contracts/extension/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+  type ErrorCode = ();
+
+  /// Returns the amount of active validators on the current chain.
+  #[ink(extension = 0, handle_status = false, returns_result = false)]
+  fn active_validators_len() -> u16;
+
+  /// Returns the ID for the current validator set for the current chain.
+  // TODO: Decide if this should be an increasing unsigned integer instead of a hash.
+  #[ink(extension = 1, handle_status = false, returns_result = false)]
+  fn validator_set_id() -> [u8; 32];
```
noot: idk if it matters here, but validator set ID in substrate is an incrementing uint

contracts/multisig/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+  /// A contract which tracks the current multisig keys.
+  #[ink(storage)]
+  #[derive(SpreadAllocate)]
+  pub struct Multisig {
+    /// Validator set currently holding the multisig.
+    validator_set: [u8; 32],
+    /// Mapping from a curve's index to the multisig's current public key for it.
+    // This is a mapping due to ink's eager loading. Considering we're right now only considering
+    // secp256k1 and Ed25519, it may be notably more efficient to use a Vec here.
+    keys: Mapping<u8, Vec<u8>>,
```
noot: where/how are the curves indexed?

contracts/multisig/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+    /// Voter + Keys -> Voted already or not
+    voted: Mapping<(AccountId, [u8; 32]), ()>,
+    /// Validator Set + Keys -> Vote Count
+    votes: Mapping<([u8; 32], [u8; 32]), u16>,
```
noot: what are the `keys` representing here? is it the multisig key?

contracts/multisig/lib.rs
```diff
@@ -0,0 +1,303 @@
...
+
+      let validator = self.env().caller();
+      if !self.env().extension().is_active_validator(&validator) {
+        Err(Error::NotValidator)?;
+      }
+
+      let validator_set = self.env().extension().validator_set_id();
+      if self.validator_set == validator_set {
+        Err(Error::AlreadyGeneratedKeys)?;
+      }
```
noot: I'm a bit confused as to what this is checking :p

edit: it's basically checking that we're not voting to enact the current set ID I think?

contracts/extension/src/lib.rs
kayabaNerve: It does. Thanks :) Will update.

contracts/multisig/lib.rs
kayabaNerve: Off-chain by the processor.

contracts/multisig/lib.rs
kayabaNerve: The hash of all multisig keys being proposed. Will add a comment clarifying.

contracts/multisig/lib.rs
kayabaNerve: It's checking that a validator set can't have multiple keys provided. If a keygen supposedly fails by one vote, and a new keygen succeeds, it shouldn't be possible for a single party to tell the chain the old one succeeded as well by submitting their vote AFTER having already performed a keygen.

docs/protocol/Constants.md
```diff
@@ -0,0 +1,54 @@
...
+| Ed25519   | 1  |
+
+### Networks
+
+Every network connected to Serai operates over a specific curve. While the
+processor generates keys for curves, these keys are bound to specific networks
+via an additive offset created by hashing the network's name (among other
+things). The network's key is used for all coins on that network.
+
+Networks are not acknowledged by the Serai network, solely by the processor.
```
kayabaNerve, line 36: This could be moved under Serai, with the benefit that validator sets would be defined over Networks, not over Coins. This is a level of indirection though, when handling Oraclizations, despite slightly cleaning the validator set definition (and ensuring that property now), So... pros and cons? It may be worth to move it on chain, have that literal impl, it may be worth it to leave it off chain, as it has no direct reason to be on?

docs/protocol/Consensus.md
```diff
@@ -0,0 +1,25 @@
+# Oraclization (message)
```
kayabaNerve, line 1: This is an ugly as hell document which is out of scope for this PR. I just moved it out of a document I did correct under this PR.

docs/protocol/Multisig.md
```diff
@@ -0,0 +1,35 @@
...
+successfully generated shares, actually creating a t-of-t multisig in practice,
+which is not BFT. This does mean a single node can delay a churn, which is
+expected to be handled via a combination of slashing, and if necessary, removal.
+
+Validators are allowed to vote multiple times across sets of keys, with the
+first set to be confirmed becoming the set of keys for that validator set. These
+keys remain valid for the validator set until it is changed. If a validator set
+remains consistent despite the global validator set updating, their keys carry.
+If a validator set adds a new member, and then loses them, their historical keys
+are not reused.
```
kayabaNerve, line 23: This is decently annoying to implement. We can't index from H(validators) to keys due to this last property, which was intended to let nodes move to another node without their history and obligations (at least, for this specific issue). We also can't index by global set ID, validator set index as the validator set may be unchanged across global set ID changes. Considering the multisig contract just tracks the most recent keys until updated, we could simply have the processor NOP in this case, handling it off chain? Reduce dev work and consensus obligations?

contracts/multisig/lib.rs
kayabaNerve: Document has been added specifying Curves/Networks/Coins.

contracts/extension/src/lib.rs
kayabaNerve: Properly defined in the new Validator Sets document (or at least, close enough, as it's defined as a u32 in that doc).

contracts/multisig/lib.rs
kayabaNerve: Used type aliases, so this should be clear now.

kayabaNerve:
This PR did end up defining Validator Sets, as this ink! contract needed to be built with those considerations in mind. Bit of scope creep, yet I'm fine just moving forward there.

docs/protocol/Validator Sets.md
```diff
@@ -0,0 +1,28 @@
...
+- `bond`  (Amount):    Amount of bond per key-share of this validator set.
+- `coins` (Vec<Coin>): Coins managed by this validator set.
+
+At launch, there will solely be validator set 0, managing Bitcoin, Ethereum,
+USDC, DAI, and Monero.
+
+### Multisig Management
+
+Every validator set is expected to form a multisig per curve required by its
+coins. This multisig is secure to hold funds up to 67% of the validator set's
```
TheArchitect108: Where will we define this "up to 67%"?

docs/protocol/Validator Sets.md
```diff
@@ -0,0 +1,28 @@
+# Validator Sets
+
+Validator Sets are defined at the protocol level, with the following parameters:
+
+- `index` (VS):        Validator set index, a global key atomically increasing
+from 0.
+- `bond`  (Amount):    Amount of bond per key-share of this validator set.
+- `coins` (Vec<Coin>): Coins managed by this validator set.
+
```
TheArchitect108, line 9: Maybe a status for the set?

docs/protocol/Validators.md
```diff
@@ -1,41 +1,48 @@
...
-`Oraclization` messages are published by the current block producer and communicate an external event being communicated to the native chain. This is presumably some other cryptocurrency, such as BTC, being sent to the multisig wallet, triggering a privileged call enabling relevant actions.
+  - `delegator` (signer):  Address removing delegated funds from `validator`.
+  - `validator` (address): Registered validator no longer being delegated to.
+  - `amount`    (Amount):  Amount of funds no longer being delegated to
+`validator`.
 
-# Report (message)
+If a sufficient amount of funds are queued, the `validator`'s operation
+continues normally, shifting in queued funds. If the `validator` falls below a
+multiple of the validator set's bond, they will lose a key share at the next
```
TheArchitect108: Seems we could take queued bond into account when distributing rewards and selecting validators to churn in while funds stay protected by unit bond. Thoughts?

kayabaNerve: The entire queue system was removed as part of #55. If/when it is re-introduced, I'd rather not consider it for rewards in order to have people actively work towards making it active, rotating it as needed.

docs/protocol/Validator Sets.md
kayabaNerve: It's intended to be based on the `t` decided for the `t-of-n` multisig. This doesn't note the `t-of-n` multisig formation however, nor `t` selection which is intended to provide this property. Will update accordingly :+1:

docs/protocol/Validator Sets.md
kayabaNerve: I'd rather simply append in hard forks, and if we ever remove for some reason, set `coins.len` to `0`. I don't see a benefit for an additional status as a protocol level, especially since rewards would be based on their managed coins (encouraging validators to provide bond where users need it), so no coins would cause it to have no effect. Seems more like an RPC comment. Feel free to clarify why though :)

PR #59, opened 2022-07-21
Document Instructions and various network integrations

### LINKED ISSUE #57: Define Architecture/Add "Instructions"
Instructions are DEX commands included within the data fields of each supported currency. Instructions provide the necessary context so that the protocol understands intent and destination.

Define instructions for:

- [ ] Add Bond
- [ ] Withdraw Bond
- [ ] Resign
- [ ] Swap
- [ ] Add Liquidity
- [ ] Withdraw Liquidity
- [ ] Donate

kayabaNerve:
Closes https://github.com/serai-dex/serai/issues/57.

Does use terminology from #53.

### DIFF
--- a/docs/integrations/Bitcoin.md
+++ b/docs/integrations/Bitcoin.md
@@ -1,0 +1,23 @@
+# Bitcoin
+
+### Addresses
+
+Bitcoin addresses are an enum, defined as follows:
+
+  - `p2pkh`:  20-byte hash.
+  - `p2sh`:   20-byte hash.
+  - `p2wpkh`: 20-byte hash.
+  - `p2wsh`:  32-byte hash.
+  - `p2tr`:   32-byte key.
+
+### In Instructions
+
+Bitcoin In Instructions are present via the transaction's last output in the
+form of `OP_RETURN`, and accordingly limited to 80 bytes. `origin` is
+automatically set to the transaction's first input's address, if recognized.
+If it's not recognized, an address of the multisig's current Bitcoin address is
+used, causing any failure to become a donation.
+
+### Out Instructions
+
+Out Instructions ignore `data`.
--- a/docs/integrations/Ethereum.md
+++ b/docs/integrations/Ethereum.md
@@ -1,0 +1,16 @@
+# Ethereum
+
+### Addresses
+
+Ethereum addresses are 20-byte hashes.
+
+### In Instructions
+
+Ethereum In Instructions are present via being appended to the calldata
+transferring funds to Serai. `origin` is automatically set to the party from
+which funds are being transferred. For an ERC20, this is `from`. For ETH, this
+is the caller. `data` is limited to 255 bytes.
+
+### Out Instructions
+
+`data` is limited to 255 bytes.
--- a/docs/integrations/Ethereum.md
+++ b/docs/integrations/Ethereum.md
@@ -14,3 +14,25 @@ is the caller. `data` is limited to 255 bytes.
 ### Out Instructions
 
 `data` is limited to 255 bytes.
+
+If `data` is provided, the Ethereum Router will call a contract-calling child
+contract in order to sandbox it. The first byte of `data` designates which child
+child contract to call. After this byte is read, `data` is solely considered as
+`data`, post its first byte. The child contract is sent the funds before this
+call is performed.
+
+##### Child Contract 0
+
+This contract is intended to enable connecting with other protocols, and should
+be used to convert withdrawn assets to other assets on Ethereum.
+
+  1) Transfers the asset to `destination`.
+  2) Calls `destination` with `data`.
+
+##### Child Contract 1
+
+This contract is intended to enable authenticated calls from Serai.
+
+  1) Transfers the asset to `destination`.
+  2) Calls `destination` with `data[.. 4], serai_address, data[4 ..]`, where
+`serai_address` is the address which triggered this Out Instruction.
--- a/docs/integrations/Instructions.md
+++ b/docs/integrations/Instructions.md
@@ -1,0 +1,153 @@
+# Instructions
+
+Instructions have two forms, In and Out. For a transaction originating on a
+connected network, an In Instruction must be provided, which may embed an Out
+Instruction. For a transaction originating on Serai, only an Out Instruction is
+allowed. Additionally, short hand forms are provided to minimize representations
+on connected networks.
+
+Instructions are interpreted according to their non-Serai network. Addresses
+have no validation performed, beyond being a valid enum entry (when applicable)
+of the correct length, unless otherwise noted. If the processor is instructed to
+act on invalid data, or send to itself, it will drop the entire instruction.
+
+### Serialization
+
+  - Numbers are exclusively unsigned and encoded as compact integers under
+SCALE. If omitted, `0`.
+  - Enums are prefixed by an ordinal byte of their type, followed by their
+actual values.
+  - Vectors are prefixed by their length. If omitted, `vec![]`.
+  - Instruction fields are numbered and sequentially encoded, each prefixed by
+an ordinal byte. All other fields are sequentially encoded with no markers.
+
+Certain fields may be omitted depending on the network in question.
+
+### In Instructions
+
+  - `origin` (Address): Address from the network of origin which sent funds in.
+  - `target` (Address): The ink! contract to transfer the incoming funds to.
+  - `data`   (Vec<u8>): The data to call the target with.
+
+Networks may automatically provide `origin`. If they do, the instruction may
+still provide `origin`, overriding the automatically provided value. If no
+`origin` is provided, the instruction is dropped.
+
+Upon receiving funds, the respective Serai Asset contract is called, minting the
+appropriate amount of coins, and transferring them to the specified target with
+the attached data.
+
+If the transaction fails, funds are scheduled to be returned to `origin`.
+
+### Out Instructions
+
+  - `destination` (Network, Address): Address to receive funds to.
+  - `data`        (Vec<u8>): The data to call the target with.
+
+If the network is Serai, this is a transfer. Else, it's a withdrawal to the
+specified address with the specified data. Asset contracts perform no validation
+on these fields.
+
+### Shorthand
+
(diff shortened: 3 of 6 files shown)


docs/integrations/Bitcoin.md
```diff
@@ -0,0 +1,23 @@
...
+
+### Addresses
+
+Bitcoin addresses are an enum, defined as follows:
+
+  - `p2pkh`:  20-byte hash.
+  - `p2sh`:   20-byte hash.
+  - `p2wpkh`: 20-byte hash.
+  - `p2wsh`:  32-byte hash.
+  - `p2tr`:   32-byte key.
```
kayabaNerve, line 11: Bitcoin does not validate P2TR keys, so this is fine.

docs/integrations/Bitcoin.md
```diff
@@ -0,0 +1,23 @@
...
+
+### In Instructions
+
+Bitcoin In Instructions are present via the transaction's last output in the
+form of `OP_RETURN`, and accordingly limited to 80 bytes. `origin` is
+automatically set to the transaction's first input's address, if recognized.
```
kayabaNerve, line 17: I also debated the change output. There's guaranteed to be a input though, and there's not guaranteed to be a change output. It will be more annoying to recover the input's address though.

docs/integrations/Ethereum.md
```diff
@@ -0,0 +1,16 @@
...
+### Addresses
+
+Ethereum addresses are 20-byte hashes.
+
+### In Instructions
+
+Ethereum In Instructions are present via being appended to the calldata
+transferring funds to Serai. `origin` is automatically set to the party from
+which funds are being transferred. For an ERC20, this is `from`. For ETH, this
+is the caller. `data` is limited to 255 bytes.
```
kayabaNerve, line 12: Matches Monero, fits in a single byte. A Curve trade takes ~120 bytes. Uv3 are massive at ~580 bytes, and likely best with an interim providing that expansion. No further considerations were made.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Serialization
+
+  - Numbers are exclusively unsigned and encoded as compact integers under
+SCALE. If omitted, `0`.
+  - Enums are prefixed by an ordinal byte of their type, followed by their
+actual values.
+  - Vectors are prefixed by their length. If omitted, `vec![]`.
+  - Instruction fields are numbered and sequentially encoded, each prefixed by
+an ordinal byte. All other fields are sequentially encoded with no markers.
```
kayabaNerve: Instruction fields are more frequently omitted, hence why we solely mention the ones included.

Shorthand fields are more frequently included, hence why we don't bother numbering them.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+Upon receiving funds, the respective Serai Asset contract is called, minting the
+appropriate amount of coins, and transferring them to the specified target with
+the attached data.
+
+If the transaction fails, funds are scheduled to be returned to `origin`.
+
+### Out Instructions
+
+  - `destination` (Enum { Native(Address), Serai(Address) }): Address to receive
+funds to.
```
kayabaNerve, line 59: We *can* convert BTC origin addresses to ETH addresses, and can convert used ETH EOA accounts to BTC addresses. I don't believe any wallet would dare touch such functionality, and while it's interesting, and would get memos to 10 bytes, is best not implemented (not to mention the scope).

kayabaNerve: This was originally (Network, Address). Sending XMR to an ETH address isn't valid though. The authenticated calling API could be used to build a wrapped token system, and is the intended path for any potential bridge. It'd be appropriately sandboxed and doesn't raise the question of XMR -> BTC (SLP tokens? I guess?), which would be a possible combination for simply (Network, Address).

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+```
+
+where `swap` is a function which:
+
+  1) Swaps the incoming funds for SRI.
+  2) Swaps the SRI for `coin`.
+  3) Checks the amount of `coin` received is greater than `minimum`.
+  4) Executes `out` with the amount of `coin` received.
+
+For a Bitcoin to Monero swap, Swap Shorthand is expected to generally take:
```
kayabaNerve: This may be better left to the PR comments, yet this does have to be less than 80 bytes. This math shows it is.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+##### Pong Example
+
+```
+In Instruction {
+  target: Bitcoin Asset Contract,
+  data:   Withdraw(Out Instruction { destination: Native(Bitcoin Sender) })
+}
+```
+
+would cause the created seraiBTC to be transferred to the Bitcoin Asset Contract
```
kayabaNerve: sBTC is synth BTC, which is not what we have. sriBTC isn't optimal. I do prefer the long form seraiBTC, with a token symbol of simply BTC on Serai.

docs/integrations/Monero.md
```diff
@@ -0,0 +1,24 @@
...
+This definition of Featured Addresses is non-standard given the flags are
+intended to be a VarInt, yet as of now, only half of the bits are used, with no
+further planned features. Accordingly, it should be fine to fix its length,
+which makes it comply with expectations present here. If needed, another enum
+entry for a 2-byte flags Featured Address could be added.
+
+### In Instructions
+
+Monero In Instructions are present via `tx.extra`, specifically via inclusion
+in a `TX_EXTRA_TAG_PADDING` tag, and accordingly limited to 255 bytes.
```
kayabaNerve, line 20: Monero wallets just error when scanning anything larger. We *could* ignore this without issue, yet we have no incentive to at this time, and it may introduce incompatibilities with wallets sending to us.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+In Instruction {
+  origin,
+  target: Router,
+  data:   swap_and_add_liquidity(Incoming Asset, address, minimum, gas)
+}
+```
+
+where `swap_and_add_liquidity` is a function which:
+
+  1) Swaps half of the incoming funds for SRI.
```
kayabaNerve, line 120: This actually leaves a potentially notable amount of potential liquidity on the table, as it doesn't consider slippage at all. It's just also trivial to calculate, and the amounts aren't notable at all when slippage isn't notable due to the size of the liquidity pool.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+  - 1 byte to not override `origin`.
+  - 1 byte for `coin`.
+  - 4 bytes for `minimum`.
+  - 1 byte for `out`'s `destination` field label.
+  - 1 byte for `out`'s `destination`'s ordinal byte.
+  - 65 bytes for `out`'s `destination`'s address.
+
+Or 74 bytes.
+
+##### Add Liquidity
```
kayabaNerve, line 100: Ideally, Add Liq would solely be from Serai. The issue is the Monero UX.

Monero can:

- Swap and add (1 XMR TX, 108 bytes)
- Swap for Serai + gas, wrap, add liquidity (1 XMR TX, 107 bytes, wait 20 minutes, 1 XMR TX, ~98 bytes, 1 Serai TX)
- Swap for Serai while also wrapping (1 XMR TX, 111 bytes, 1 Serai TX)

So I believe we have to accept this exposure and 108 bytes on Monero, not including the 1 byte label included on Monero (and tx.extra VarInt length size increase, +1 byte there too).

docs/integrations/Instructions.md
kayabaNerve: To elaborate, making any bridge a first class citizen doesn't really save on complexity, especially due to the scope creep. It would decrease gas costs on ETH, yet with increased first-party complexity when those are already a large burden, and this keeps thing modular. There is another discussion a bridge still needs to be a first class citizen to enable redemptions (/ we need a new API exposed for that), yet we're not tackling any bridge discussions yet (not even as a tracking issue, and I'd prefer to keep it that way). We're solely acknowledging API flexibility.

docs/integrations/Bitcoin.md
TheArchitect108: Less = more in this case

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+Instructions have two forms, In and Out. For a transaction originating on a
+connected network, an In Instruction must be provided, which may embed an Out
+Instruction. For a transaction originating on Serai, only an Out Instruction is
+allowed. Additionally, short hand forms are provided to minimize representations
+on connected networks.
+
+Instructions are interpreted according to their non-Serai network. Addresses
+have no validation performed, beyond being a valid enum entry (when applicable)
+of the correct length, unless otherwise noted. If the processor is instructed to
+act on invalid data, or send to itself, it will drop the entire instruction.
```
TheArchitect108, line 24: We should document refunds.

docs/integrations/Instructions.md
```diff
@@ -0,0 +1,160 @@
...
+
+Networks may automatically provide `origin`. If they do, the instruction may
+still provide `origin`, overriding the automatically provided value. If no
+`origin` is provided, the instruction is dropped.
+
+Upon receiving funds, the respective Serai Asset contract is called, minting the
+appropriate amount of coins, and transferring them to the specified target with
+the attached data.
+
+If the transaction fails, funds are scheduled to be returned to `origin`.
```
TheArchitect108: transaction OR instruction?

TheArchitect108 (approved):

docs/integrations/Instructions.md
kayabaNerve: It's an equivalence here. Keeping things in terms of "instruction" is likely optimal though :)

docs/integrations/Instructions.md
kayabaNerve: Also fee logic, gas limits, how we charge fees for USDC transfers out... yet out of scope for this immediate.

Issue #32, opened 2022-07-08
Implement Monero hash_to_point in Rust


kayabaNerve:
I believe we'd want to manually implement the Ed25519 FieldElement field as a PrimeField under dalek-ff-group, as dalek doesn't ecpose it, and then build the hash_to_point on top of that.

Related https://github.com/zkcrypto/group/issues/30.

Issue #31, opened 2022-07-07
Implement Bulletproofs in Rust


kayabaNerve:
Doing so would enable removing `-z muldefs` and make our only reliance on C the hash_to_point function.

kayabaNerve:
Technically, we use CLSAG verify and communicate over the FFI via serialization. We already have an experimental CLSAG verify, which I trust without issue to be *at least as strict* *except* it doesn't reject a variety of 0 cases that Monero does. Accordingly, that can be dropped.

That leaves hash_to_point and BP prove. I'll create a new issue for hash_to_point.

kayabaNerve:
`-z muldefs` was removed with #30 and I no longer feel this is as pressing as the Monero code is appropriately minimal. Accordingly, we shouldn't rush to remove the C CLSAG verify and so on.

dan-da:
You probably are aware of, but just in case:
https://crates.io/crates/bulletproofs

kayabaNerve:
I am! The distinction is we specifically need Monero Bulletproofs. While the existing dalek code could be forked to be Monero Bulletproofs, they aren't right now.

Beyond moving from Ristretto to Ed25519, there's going to be a completely different transcript format. I assume, beyond the Monero INV_8/8 preprocessing, the actual math is the same? Yet there may be a few edge cases there.

Thanks for chiming in though :)

dan-da:
there is also this, which supports bulletproofs+ and ed25519.  but maybe not so fast?
https://crates.io/crates/bulletproof-kzen

kayabaNerve:
We can't legally use that (as it's GPL-licensed and the Monero lib is MIT), and again, we explicitly need Monero Bulletproofs.

It also uses "curv" instead of ff/group? Weird.

Issue #30, opened 2022-07-07
Decrease amount of libs statically linked with Monero


kayabaNerve:
hidapi, which I don't know how to ignore on the Monero end, is an optional lib currently linked by the monero crate. Furthermore, epee/randomx/libunbound are linked, as is boost/sodium/easylogging. Ideally, all of these would be removed.

kayabaNerve:
It will probably be best to bypass the library system and directly link various object files as needed.

Issue #13, opened 2022-06-01
Project licensing


kayabaNerve:
Currently, all of the code in the project is MIT licensed. This keeps all our infrastructure accessible, while letting the community build on it and iterate, hopefully contributing back. To be more specific, on the notable pieces of of code currently in this project:

crypto/
- The multiexp crate is beneficial to a large amount of parties and could be argued as core infrastructure for ff/group.
- The FROST library is likely to be of interest to projects with this same task, as well as of independent interest to a variety of parties.

coins/
- The Monero library offers transaction signing abilities, an API I personally prefer to monero-rs, and a solution to https://github.com/monero-project/research-lab/issues/103, making it notable even outside of Rust.

There are three further planned pieces (beyond further coins, which will largely be predicated on existing libraries):
- Processor, which has already started, which will provide a unified coin API. This will be dependent on our substrate chain for coordination and explicitly handle multisig keys though, so there's not necessarily independent value in it (using it as the basis of a Rust multi-coin wallet). It could be segmented into a coin API, and files making coins match that API, as one library, with the multisig wallet + Substrate chain handling left in the Processor itself.
- Consensus, as documented in the docs folder, which will be a modification of Aura or BABE + GRANDPA. It should be noted these are "GPL-3.0-or-later WITH Classpath-exception-2.0" already, forcing any modified implementations to also be GPL-3.
- Daemon, which will be a Substrate chain using our consensus with the DEX functionality.

The immediate plan is to update the processor's license to AGPL, and use AGPL for the daemon (along with consensus if possible).

Pros:
- Decreased odds of flippant forks which don't provide additional merit.
- Forks must provide their source code to all users.
- Since this is only being applied to binaries, these projects aren't expected to built on top of, not hampering development, and these licenses shouldn't notably affect anyone who doesn't deploy a modified version (network-accessible instances must only declare themselves when modified).

Cons:
- The AGPL is a heavy handed license which may intimidate developers or make them not prefer to contribute.
- A multi-coin Rust wallet can't use our unified coin type and existing integrations without being AGPL (without separation, which I'm not against). This is the sole non-fork reason I can think of which would be impacted by this.
- Decreased odds of forks, which can be considered a con despite being listed as a pro.

[...]

PR #63, opened 2022-07-22
Experimental Dockerfile support

yatima1460:
`docker build --tag serai .`

For now running it just prints the version, needs arguments and ports!

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,8 @@
+.vscode
+README.md
+AGPL-3.0
+docs
+Dockerfile
+.dockerignore
+target
+Cargo.lock
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,6 @@
 # serai Docker image
 #
 # ex: docker build --tag serai .
-# 
-# takes about 10 minutes to build on a Ryzen 5900HX
 
 FROM debian:bullseye-slim as build
 
@@ -57,6 +55,7 @@ WORKDIR /serai
 
 # Finally build!
 #TODO: maybe ARCH and nproc as ARG parameters?
+# Takes about 15 minutes to build on a Ryzen 5900HX with 32GB of RAM
 RUN ARCH=default cargo build --all-features --release --verbose -j$(nproc)
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,73 @@
+#
+# serai Docker image
+#
+# ex: docker build --tag serai .
+# 
+# takes about 10 minutes to build on a Ryzen 5900HX
+
+FROM debian:bullseye-slim as build
+
+
+## BUILD ENVIRONMENT ##
+
+RUN apt-get update
+
+# Install solc
+RUN apt-get install -y python3-pip
+RUN pip3 install solc-select
+RUN solc-select install 0.8.9
+RUN solc-select use 0.8.9
+
+# Install Monero dependencies
+RUN apt-get install -y \
+    build-essential \
+    cmake \
+    pkg-config \
+    libboost-all-dev \
+    libssl-dev \
+    libzmq3-dev \
+    libpgm-dev \
+    libunbound-dev \
+    libsodium-dev \
+    ccache
+
+# Install Rust
+RUN apt-get install -y curl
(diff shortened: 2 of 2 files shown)


kayabaNerve:
Hey! Great to see you again :D Thanks for submitting this.

I know https://github.com/substrate-developer-hub/substrate-node-template offers some level of Docker code, which may be relevant.

I did just remove the requirements on building Monero. It's something I missed when doing #30 which should help here with both amount of dependencies and build time. I unfortunately think most of that is from Substrate though, which isn't something we can do anything about.

But yes, it still requires git in order to populate monero/. Specifically, as of right now, it uses a submodule and I'm not even sure that'll work if packaged under Cargo. I believe the current submodule will need to replaced with a git clone for that, and I don't have a better idea for packaging it at this time. This requirement will likely stand until we replace all C functions with Rust, as noted in #31 and #32.

We should work with stable Rust, only using nightly for wasm.

I... didn't know about the clang requirement? We don't grab clang for the CI? I'm not saying it isn't needed, as it may come shipped with Ubuntu/Ubuntu's build-essential and not Debian, but that'd be very surprising. Not an issue to simply include it here and call it a day though.

Regarding the `.dockerignore`... does this count as a source/binary distribution when packaged? It seems like the Dockerfile grabs this source distribution (the repo itself) to create a binary distribution it fully packages. That should preserve the AGPL license and we'd also have to include at least the GPL 3.0 due to Substrate's licensing. Then we'd have to review Apache 2.0/MIT requirements (if we ever actually published the built Docker image).

It's bullshit like that why I just stripped Docker for now, as we don't have any nodes being deployed for testing. Protonet will have that though, and this is a solid base we can use then, so thank you for it :)

yatima1460:
Yes GitHub Actions already comes with Clang preinstalled!

Here you can check the lists!

https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md

kayabaNerve:
That'd explain it :) Thanks for clarifying.

yatima1460:
The best way probably would be to build Monero Docker image and then push it to GitHub Container Registry and use that as a cached artifact for other images

yatima1460:
Generally I never use the GitHub Actions jobs as "raw" unless creating a Docker image is prohibitive 

Because they are never portable and replicating the build process on a local machine becomes painful

yatima1460:
Also another good advantage of having a perfect Docker image is that it behaves like build instructions and a dependencies documentation 😄

kayabaNerve:
If we're discussing Monero packaging, my consideration was about the Rust crate. Grabbing a Docker image doesn't help when that needs source :p

If we're discussing Monero daemon packaging, for testing purposes, this would matter if we moved the CI over to Docker. I... wouldn't be against that? Except I did just manage to get GH Actions setup lol. So I'd be curious to see the benefit.

If we're discussing Monero packaging for release purposes, where Serai validators need a Serai node and a Monero node, we'll likely want a Kubernetes + Docker structure there.

I definitely appreciate having a dependencies/build document available, which these do serve as a literal version of. I don't think I even comment a regtest daemon is needed to run the Monero tests at this point in time, when I definitely should... Currently working on getting more Serai protocol docs up though.

Want to reach out on Matrix/Discord so we can talk more there? :D I'll add links to the README since apparently I've forgotten to.

kayabaNerve:
@TheArchitect108 brought up merging this. To highlight the steps required:

- .dockerignore canonical to .gitignore
- Including licensing as needed, related to #13 (for now, just not excluding the AGPL)

- Correct #TODO for # TODO, which is minor and will be in #77 
- newline before EOF, which will be in #77 
- No double lines, which is in #77 

- Use stable Rust
- Remove git and the Monero dependencies (which were required when this PR was made)

- Replace FIXME with TODO, for specifying a clang version, though I don't actually care to nitpick to such a degree and rather remove that note entirely. While it may be nice to have a firm knowledge of the required clang version, that's a RocksDB decision where we follow their lead
- Actually run Serai, I guess in the dev mode for now

TheArchitect108:
Thanks for the contribution, closing this out as it was covered in #114. Happy to coordinate on further work, feel free to ping me on discord.

PR #74, opened 2022-07-31
Bp verify

DangerousFreedom1984:
Hello KayabaNerve,

Here is a basic implementation of the verify function for bulletproofs. It does not work for batches nor aggregated proofs yet. I will improve it during this week ;)


kayabaNerve:
Closed by #75.

PR #78, opened 2022-08-12
Add constant `EDWARDS_D` in dalek-ff-group

AiyionPrime:
The value is working in my implementation, a test is implemented, that d equals `-121665/121666`.


AiyionPrime:
Got it.
```bash
rustup update
rustup update nightly
rustup target add wasm32-unknown-unknown --toolchain nightly
```

AiyionPrime:
Pull request is ready for review,
`cargo test` has a new working one,
`cargo fmt` and `cargo clippy` are silent on my end, at least for `dalek-ff-group`.

In one of your other modules appears to be a match with unnecessary brackets, that will likely throw your github action of.

kayabaNerve:
clippy recently updated, which flagged on prior code. It's not your fault ;)

Thanks for this :D Definitely appreciated.

PR #79, opened 2022-08-12
Add implementation for `is_odd()`

AiyionPrime:
> in dalek-ff-group

A number is considered odd, if the low bit is 1.

Tests are implemented,
`is_negative`s output matches that of other implementations,
`cargo fmt` is silent, as is `clippy`.

This should be ready for review.


kayabaNerve:
Thanks for these! If you want to offer conditional_negate via ConditionalNegate, I'd be happy to merge it as well :) Let me know if you want help with sqrt_ratio_i!

Issue #83, opened 2022-08-13
Offer a dedicated Schnorr crate


kayabaNerve:
We currently offer:

- Schnorr in DLEq (quality transcript, batch verification on a passed in batch verifier)
- Schnorr in FROST (similar properties, yet manages the batch verifier itself)
- A modular HRAm in FROST for its modularity

Ideally we have one Schnorr crate supporting:

- A modular HRAm
- R, s Schnorr Signatures
- c, s Schnorr signatures via re-calculation of c (I don't care to offer an enum for the Schnorr signature nor variadic return types)
- Batch verification against an existing batch verifier
- Batch verification with its own batch verifier (sugar)
- Sugar for Schnorr over an existing transcript (using a quality transcript) and sugar for Schnorr as an isolate (with a classical HRAm)

With DLEq and FROST both deferring to it.

kayabaNerve:
We need custom generator Schnorr .-.  I'd like to move this after #84.

PR #86, opened 2022-08-14
Curve lib

### LINKED ISSUE #84: Curve crate
A `Curve` trait, similar to ff/group (and defined over ff/group), yet offering a variety of provided methods (\*\_len, read\_\*) and requiring Zeroize would be appreciated.

Offering a hash_to_F would also be appreciated, yet the issue is how hash_to_F varies across curves. Ed25519 does it raw, secp256k1 does ExpandMessageXMD. While an IETF hash_to_curve implementation composable for anything implementing ff/group would be appreciated, it still needs to be selected per-curve, where-as I'd like to keep the Curve trait incredibly minimal (ideally `impl Curve for Group`).

Perhaps:
- Curve
- ZeroizeCurve
- HashToCurve
- ZeroizeCurve + HashToCurve (SeraiCurve? FeaturedCurve? CompleteCurve?)

Would be quite minimal, yet de-duplicates code as does #83.

kayabaNerve:
Provides common sugar between DLEq and FROST. Quirks the API a bit, while resolving it in other ways at the same time.

I'd like to bring #83 under this to fully validate it before merging, yet wanted this open now.

Closes #84.

TODO:
- [ ] HashToCurve usage in frost
- [x] Move Curve's serialize to std, enabling nostd like DLEq

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4548,6 +4548,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "modular-schnorr"
+version = "0.1.0"
+dependencies = [
+ "blake2",
+ "composable-curve",
+ "dalek-ff-group",
+ "multiexp",
+ "rand_core 0.6.3",
+]
+
+[[package]]
 name = "monero"
 version = "0.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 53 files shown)


kayabaNerve:
I'm not exceptionally happy with this PR. Merging in FROST should be possible, though merging in DLEq doesn't look like it. The Curve lib also makes Rust lose focus on types occasionally. Not really as good as I wanted it to be... But probably still worth it overall.

kayabaNerve:
Additional note the read_G bans identity. That's a FROST rule, not a curve-intended rule.

PR #87, opened 2022-08-15
Add implementation for `sqrt_ratio_i()`

AiyionPrime:
> in dalek-ff-group

`cargo fmt` and `clippy` are silent,
tests are fine.

Had no need for `conditional_negate` so I left it out for now, might add it later.

> Shouldn't that be a CtOption though?

_Originally posted by @kayabaNerve in https://github.com/dalek-cryptography/curve25519-dalek/issues/389#issuecomment-1212682725_

This matches the function definition of curve25519-dalek, so it returns `(Choice, FieldElement)`.

I haven't seen a conflicting Trait that specifies it to be `CtOption`.

### DIFF
--- a/crypto/dalek-ff-group/src/field.rs
+++ b/crypto/dalek-ff-group/src/field.rs
@@ -156,6 +156,30 @@ impl FieldElement {
     }
     res
   }
+
+  pub fn sqrt_ratio_i(u: FieldElement, v: FieldElement) -> (Choice, FieldElement) {
+    let v3 = v.square() * v;
+    let v7 = v3.square() * v;
+    let mut r = (u * v3) *
+      (u * v7).pow((-FieldElement::from(5u8)) * FieldElement::from(8u8).invert().unwrap());
+    let check = (v) * r.square();
+    let i = SQRT_M1;
+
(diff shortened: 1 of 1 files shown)


kayabaNerve:
Hey! Sorry for the delay in reviewing this.

Was there a reason to take it in by borrow? It's copied around a bunch regardless and should be sufficiently small we can perform that lint as a nicety.

Algorithm looks great :+1: Thanks for the submission.

AiyionPrime:
> Was there a reason to take it in by borrow?

Not from functionalities perspective, just to match the *exact* fn-definition curve25519-dalek has, but I can change that.

kayabaNerve:
Would be appreciated, unless that specific property is desired, yet I'm happy to do lint myself too :p Don't mean to force you into it ;)

AiyionPrime:
No problem, will do.
I'll get to it in the afternoon.

AiyionPrime:
resolved in b95dd5eb41c3323561135136245a7bd005221a5c

AiyionPrime:
Tests, fmt and clippy are good.

kayabaNerve:
Thanks :D Definitely good to have.

Issue #82, opened 2022-08-13
chrono has a security issue in 0.4.19 and 0.4.20/0.4.21 break misc compatibility


kayabaNerve:
https://github.com/chronotope/chrono/issues/674 for the original security issue, which pinged for us and still does (though it may not apply).

https://github.com/chronotope/chrono/issues/755 and https://github.com/chronotope/chrono/issues/770 for the .20/.21 issues.

Waiting for .22 would likely be optimal to reduce amount of changes, despite .21 seeming fine if we don't use SGX (we do not). Just a minor tracking issue here.

Labelled substrate yet technically applies to ethers as well. `cargo update` should resolve all when the time comes.

kayabaNerve:
It looks like this advisory was actually withdrawn, and I do not mean to draw negative attention to chrono, yet solely keep things as pristine as possible over here. If GH didn't ping everyone all the time, I wouldn't have...

Issue #81, opened 2022-08-13
kvdb-rocksdb uses an old rocksdb, leading to two separate security issues


kayabaNerve:
https://github.com/paritytech/parity-common/issues/659 for its usage of owning_ref, which is also used by prometheus-client (https://github.com/prometheus/client_rust/issues/77).

https://github.com/paritytech/parity-common/issues/437 was closed 6 months ago as wontfix due to how rocksdb would be "deprecated" for paritydb, despite there still being no timeline on when and a lack of further reasoning directly stated (which would've been appreciated, albeit may just be available elsewhere and accordingly fine).

kvdb-rocksdb also uses rocksdb 0.18, which has a vulnerability in a certain API. While that may not be relevant, I'd appreciate using 0.19 to close out the advisory.

We'd have to fork parity-common to resolve this while staying on rocksdb, and we're reliant on prometheus's update anyways so there's no reason to yet. They thankfully are handling this properly.

Relevant to #28.

kayabaNerve:
prometheus-client has updated. libp2p has yet to update to the new prometheus client, and I don't care to fork libp2p to do so.

```
fixed-hash = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
kvdb = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
kvdb-memorydb = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
kvdb-rocksdb = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
rlp = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
uint = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
parity-util-mem = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
primitive-types = { git = "https://github.com/paritytech/parity-common", rev = "26d712dbc401f6065250baca41663c0c8298e60a" }
```

The following patch DOES use the new code. `substrate/client/db/src/upgrade.rs` has conflicts though. We'd have to update at the substrate level as well, or wait a bit longer as we already have to for libp2p.

[...]

PR #89, opened 2022-08-17
Add a cargo deny workflow

kayabaNerve:
Also trims out a pointless submodule checkout (we have none).

Relevant to #81 and #82.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -34,10 +34,10 @@ allow = [
   "OpenSSL",
 
   # Non-invasive copyleft
+  "MPL-2.0",
   "Apache-2.0",
   "Apache-2.0 WITH LLVM-exception",
-  "GPL-3.0-only WITH Classpath-exception-2.0",
-  "GPL-3.0-or-later WITH Classpath-exception-2.0",
+  "GPL-3.0 WITH Classpath-exception-2.0",
 ]
 
 copyleft = "warn"
--- a/deny.toml
+++ b/deny.toml
@@ -45,6 +45,15 @@ allow-osi-fsf-free = "neither"
 default = "deny"
 
 exceptions = [
+  { allow = ["AGPL-3.0"], name = "ethereum-serai" },
+  { allow = ["AGPL-3.0"], name = "serai-processor" },
+
+  { allow = ["AGPL-3.0"], name = "serai-extension" },
+  { allow = ["AGPL-3.0"], name = "serai-multisig" },
+
+  { allow = ["AGPL-3.0"], name = "serai-runtime" },
+  { allow = ["AGPL-3.0"], name = "serai-consensus" },
+  { allow = ["AGPL-3.0"], name = "serai-node" },
 ]
 
 [[licenses.clarify]]
--- a/deny.toml
+++ b/deny.toml
@@ -75,4 +75,5 @@ unknown-git = "deny"
 allow-registry = ["https://github.com/rust-lang/crates.io-index"]
 allow-git = [
   "https://github.com/serai-dex/substrate",
+  "https://github.com/hack-ink/array-bytes"
 ]
--- a/deny.toml
+++ b/deny.toml
@@ -40,7 +40,7 @@ allow = [
   "GPL-3.0 WITH Classpath-exception-2.0",
 ]
 
-copyleft = "warn"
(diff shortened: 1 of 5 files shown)


deny.toml
```diff
@@ -0,0 +1,73 @@
+[advisories]
+db-path = "~/.cargo/advisory-db"
+db-urls = ["https://github.com/rustsec/advisory-db"]
+
+vulnerability = "deny"
```
TheArchitect108, line 5: We should add a security threshold so this doesn't trigger on every little thing. Medium?

Threshold for security vulnerabilities, any vulnerability with a CVSS score lower than the range specified will be ignored. Note that ignored advisories will still output a note when they are encountered.
* None - CVSS Score 0.0
* Low - CVSS Score 0.1 - 3.9
* Medium - CVSS Score 4.0 - 6.9
* High - CVSS Score 7.0 - 8.9
* Critical - CVSS Score 9.0 - 10.0
`severity-threshold = `

deny.toml
```diff
@@ -0,0 +1,73 @@
...
+
+  # Non-invasive copyleft
+  "MPL-2.0",
+  "Apache-2.0",
+  "Apache-2.0 WITH LLVM-exception",
+  "GPL-3.0 WITH Classpath-exception-2.0",
+]
+
+copyleft = "deny"
+allow-osi-fsf-free = "neither"
```
TheArchitect108, line 38: did you simply deny the categories because you provide an explicit list?

deny.toml
```diff
@@ -0,0 +1,73 @@
...
+vulnerability = "deny"
+yanked = "deny"
+notice = "warn"
+unmaintained = "warn"
+
+ignore = [
+  "RUSTSEC-2020-0071", # https://github.com/chronotope/chrono/issues/602
+]
+
+[licenses]
```
TheArchitect108, line 14: do we want `private = { ignore = true }`?

deny.toml
```diff
@@ -0,0 +1,73 @@
...
+  "GPL-3.0 WITH Classpath-exception-2.0",
+]
+
+copyleft = "deny"
+allow-osi-fsf-free = "neither"
+default = "deny"
+
+exceptions = [
+  { allow = ["AGPL-3.0"], name = "ethereum-serai" },
+  { allow = ["AGPL-3.0"], name = "serai-processor" },
```
TheArchitect108, line 43: I think the `private = { ignore = true }` would allow us to skip all this?

TheArchitect108 (commented):
I think we need to ADD this TOML in the dockerfile for Serai so that it copies the deny and for other services as well. Maybe concern here is only running the action workflow, but we want it to trigger for local builds too.
ADD deny.toml .

TheArchitect108:
Also this does not do deep scanning and relies on the author to be honest about licensing. If the lib uses embedded libs, we also rely on the author to do the appropriate thing for supplying the license. May not be relevant at this point, but if we want deeper scanning in the future we should look at https://github.com/EmbarkStudios/cargo-about.

kayabaNerve:
Placing this in the Dockerfile isn't helpful, unless the Dockerfile then runs `cargo deny`. I don't feel a need to have the Dockerfile run `cargo deny` just as it shouldn't run `cargo test`.

We have clarifications available in cargo deny, and there's some level of LICENSE detection, but you're right it can misrepresented. I'm not sure `cargo about` does anything deeper. It seems to just generate a complete report/citation list.

deny.toml
kayabaNerve: No. That skips the entire tree for every unpublished crate. That means we wouldn't review our crates, nor Substrate, nor Substrate's depends.

deny.toml
kayabaNerve: Correct. If this is warn/allow, then they pass without erroring. They should not pass.

deny.toml
kayabaNerve: I'm not against this but I'd rather have the heads up so we can review it ourselves and then make a new commit explicitly allowing it. I will note the CI doesn't even run advisory checks right now for this reason, but I currently lean towards having it check everything and doing the aforementioned solution, having given it a second thought. Will reach out off GitHub to discuss.

PR #90, opened 2022-08-17
Conditional negate

AiyionPrime:
Now this is a mere draft. ~I think I might be missing something like the part around `ConditionallySelectable` in `lib.rs`~ Apparently I'm fine?.

This will look cleaner once #87 is merged.

@kayabaNerve If you'd like to take a look at this implicit approach fulfilling
```rust
impl<'a> Neg for &'a FieldElement
```
from https://docs.rs/subtle/2.4.1/subtle/trait.ConditionallyNegatable.html#impl-ConditionallyNegatable ?

I do not know whether this is better than explicitly implementing it like eg. this:

```rust
fn conditional_negate(&mut self, negate: Choice) {
    self.conditional_assign(&(-(*self)), negate);
}
```

Your choice.

### DIFF
--- a/crypto/dalek-ff-group/src/field.rs
+++ b/crypto/dalek-ff-group/src/field.rs
@@ -183,6 +183,14 @@ impl FieldElement {
 }
 
 #[test]
+fn test_edwards_d() {
+  let a = -FieldElement(U256::from_u32(121665));
+  let b = FieldElement(U256::from_u32(121666));
+
+  assert_eq!(EDWARDS_D, a * b.invert().unwrap());
+}
+
+#[test]
 fn test_is_odd() {
   assert_eq!(0, FieldElement::zero().is_odd().unwrap_u8());
   assert_eq!(1, FieldElement::one().is_odd().unwrap_u8());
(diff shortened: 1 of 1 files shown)


AiyionPrime:
Looks like I can use ConditionallyNegatable just fine, not sure how the lack of that in `lib.rs` affects users.

AiyionPrime:
rebased onto develop

AiyionPrime:
@kayabaNerve ex- or implicit, which do you want?

kayabaNerve:
Sorry, have had a few things. I'd assume you didn't have to implement ConditionalNegate because it was de-ref'ing to the underlying, which does, or `ConditionalNegate` is just implemented for everything satisfying certain traits (such as &'a Neg). Let me check this real quick.

kayabaNerve:
This seems to offer ConditionalNegate and therefore LGTM. I don't have an opinion beyond that. I don't believe it'll have problems regarding exposure, yet will check in a moment as I move the code in monero over ;)

Thanks again for all your help with these utils!

AiyionPrime:
Thanks for the ridiculous fast response times :D
Really enjoyed contributing.

Any idea, when the next minor release is up?

kayabaNerve:
Ha. Least I can do for a volunteer ;)

I can do one now :)

kayabaNerve:
Published 0.1.3.

AiyionPrime:
Awesome, nice.
Looking forward to contribute again some time.

kayabaNerve:
Happy to hear! I don't plan on entirely kidnapping you and making you the next Serai dev... BUT we do have a cryptography tag if you ever want to look there ;)

https://github.com/serai-dex/serai/issues?q=is%3Aopen+is%3Aissue+label%3Acryptography

Regarding dalek-ff-group specifically, I think it's largely done. The only other thing I'm looking for is better performance which I'm hoping to obtain when https://github.com/RustCrypto/crypto-bigint/issues/66 is resolved.

... but I do still want some Ed25519/Ristretto code :eyes:

AiyionPrime:
I'll keep that in mind, thanks.

PR #77, opened 2022-08-12
Add a basic CONTRIBUTING file

kayabaNerve:

PR #91, opened 2022-08-17
Simple getting started doc

TheArchitect108:
A basic document to clearly outline the steps necessary to build and run Serai. This document should be extended to include instructions for docker, docker compose, and processor related services.

### DIFF
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -6,7 +6,7 @@ git clone https://github.com/serai-dex/serai.git
 
 # Build and Run Serai
 
-- Ubuntu 18.04 native or WSL2 w/ Docker (Other distributions with docker may also work)
+- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
     - Windows Guide: [https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43](https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43)
     - [https://docs.docker.com/engine/install/ubuntu/](https://docs.docker.com/engine/install/ubuntu/)
 - Setup Rust & Dependencies (Assuming Linux or WSL2)
@@ -14,9 +14,9 @@ git clone https://github.com/serai-dex/serai.git
   
 ## Install Solidity Compiler
 ```
-     sudo add-apt-repository ppa:ethereum/ethereum \
-     sudo apt-get update \
-     sudo apt-get install solc
+     sudo pip3 install solc-select
+     solc-select install 0.8.16
+     solc-select use 0.8.16
 ```
 ## Install Other Dependencies
 ```sudo apt-get install -y \
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,43 +1,59 @@
-# Clone Serai
+# Getting Started
+
+## Clone Serai
+```
+    git clone https://github.com/serai-dex/serai.git
+```
+
+## Build and Run Serai
+
+### Install Rust
+
+#### Linux
+```
+    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+```
+
+#### Windows
+
+Use WSL2 and Linux instructions.
+
+
+#### Mac
 
 ```
-git clone https://github.com/serai-dex/serai.git
+    brew install rustup
 ```
 
-# Build and Run Serai
-
-- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
-    - Windows Guide: [https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43](https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43)
-    - [https://docs.docker.com/engine/install/ubuntu/](https://docs.docker.com/engine/install/ubuntu/)
-- Setup Rust & Dependencies (Assuming Linux or WSL2)
-  - Install Rust  `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
-  
-## Install Solidity Compiler
+### Install Solidity Compiler
 ```
-     sudo pip3 install solc-select
-     solc-select install 0.8.16
-     solc-select use 0.8.16
+    sudo pip3 install solc-select
+    solc-select install 0.8.16
+    solc-select use 0.8.16
 ```
-## Install Other Dependencies
-```sudo apt-get install -y \
+### Install Other Dependencies
+```
+    sudo apt-get install -y \
     cmake \
     libboost-all-dev \
     pkg-config \
     libssl-dev
 ```
 
-## Build
+### Build
 ```
     cd serai
     cargo build --release
 ```
 
-## Run
+### Run
 ```
-./target/release/serai-node --chain dev
+    ./target/release/serai-node --chain dev
+    OR
+    ./target/release/serai-node --dev
 ```
 
-## Help
+### Help
 ```
-./target/release/serai-node --help
+    ./target/release/serai-node --help
 ```
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -2,7 +2,7 @@
 
 ## Clone Serai
 ```
-    git clone https://github.com/serai-dex/serai.git
+git clone https://github.com/serai-dex/serai.git
 ```
 
 ## Build and Run Serai
@@ -11,7 +11,7 @@
 
 #### Linux
 ```
-    curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
 ```
 
 #### Windows
@@ -22,38 +22,38 @@ Use WSL2 and Linux instructions.
 #### Mac
 
 ```
-    brew install rustup
+brew install rustup
 ```
 
 ### Install Solidity Compiler
 ```
-    sudo pip3 install solc-select
-    solc-select install 0.8.16
-    solc-select use 0.8.16
+sudo pip3 install solc-select
+solc-select install 0.8.16
+solc-select use 0.8.16
 ```
 ### Install Other Dependencies
 ```
-    sudo apt-get install -y \
-    cmake \
-    libboost-all-dev \
-    pkg-config \
-    libssl-dev
+sudo apt-get install -y \
+cmake \
+libboost-all-dev \
+pkg-config \
+libssl-dev
 ```
 
 ### Build
 ```
-    cd serai
-    cargo build --release
+cd serai
+cargo build --release
 ```
 
 ### Run
 ```
-    ./target/release/serai-node --chain dev
-    OR
-    ./target/release/serai-node --dev
+./target/release/serai-node --chain dev
+OR
+./target/release/serai-node --dev
 ```
 
 ### Help
 ```
-    ./target/release/serai-node --help
+./target/release/serai-node --help
 ```
--- a/docs/Getting Started.md
+++ b/docs/Getting Started.md
@@ -1,0 +1,43 @@
+# Clone Serai
+
+```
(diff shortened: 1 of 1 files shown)


docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+    - [https://docs.docker.com/engine/install/ubuntu/](https://docs.docker.com/engine/install/ubuntu/)
+- Setup Rust & Dependencies (Assuming Linux or WSL2)
+  - Install Rust  `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+  
+## Install Solidity Compiler
+```
+     sudo add-apt-repository ppa:ethereum/ethereum \
+     sudo apt-get update \
+     sudo apt-get install solc
+```
```
noot: I like `solc-select` for installing solidity as you can easily switch between versions, eg for solidity 0.8.16:
```
pip3 install solc-select
solc-select install 0.8.16
solc-select use 0.8.16
```

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
+# Clone Serai
+
+```
+git clone https://github.com/serai-dex/serai.git
+```
+
+# Build and Run Serai
+
+- Ubuntu 18.04 native or WSL2 w/ Docker (Other distributions with docker may also work)
```
noot: ubuntu 18.04?? a bit of an older distro lol

TheArchitect108: Calling me old?? Swapped to LTS.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
+# Clone Serai
```
kayabaNerve: \# Getting Started

\### Section

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
+# Clone Serai
+
+```
+git clone https://github.com/serai-dex/serai.git
+```
+
+# Build and Run Serai
+
+- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
```
kayabaNerve: I'd like to either do Debian, pushing users from there Ubuntu, or Debian + Ubuntu, officially acknowledging Debian (as it is largely preferential) yet still supporting Ubuntu (as it's what's used). I also want to officially ack Arch as not only is what I use, but a sufficiently large amount of Linux users in general.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+git clone https://github.com/serai-dex/serai.git
+```
+
+# Build and Run Serai
+
+- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
+    - Windows Guide: [https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43](https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43)
+    - [https://docs.docker.com/engine/install/ubuntu/](https://docs.docker.com/engine/install/ubuntu/)
+- Setup Rust & Dependencies (Assuming Linux or WSL2)
+  - Install Rust  `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
```
kayabaNerve: This is only correct for Linux (which is part of your "A"ssuming, sure), but it doesn't install dependencies. Just the rustup tool itself. You'd need to additionally install the Rust toolchains in use.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+  
+## Install Solidity Compiler
+```
+     sudo pip3 install solc-select
+     solc-select install 0.8.16
+     solc-select use 0.8.16
+```
```
kayabaNerve, line 35: Missing newline after header (on purpose, yet breaking with existing) and this block (typo).

Also 4-space idented.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+## Install Other Dependencies
+```sudo apt-get install -y \
+    cmake \
+    libboost-all-dev \
+    pkg-config \
+    libssl-dev
+```
```
kayabaNerve, line 41: I don't believe these are required. We no longer build Monero so they shouldn't be. We may still need clang though... (but I believe that's shipped with Ubuntu)

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
+# Clone Serai
+
+```
+git clone https://github.com/serai-dex/serai.git
+```
+
+# Build and Run Serai
+
+- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
+    - Windows Guide: [https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43](https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43)
```
kayabaNerve: Do we want to link to a personal post? I'd rather let people fend for themselves when they have poor environments than officially link to this and carry its burden.

docs/Getting Started.md
kayabaNerve: Also, WSL2 isn't necessarily Ubuntu. Technically, it'd be Ubuntu via WSL2. Actual question, can we just drop WSL and assume anyone using Windows for dev who deals with this stuff already knows about WSL?

kayabaNerve: As one final comment on OSs, we could have one OS-specific section (Ubuntu, Debian, Arch) and then a universal set of steps.

- Install Dependencies (OS-specific)
- Install Rust (universal if Linux)
- Install solc (universal under pip3)
- Build (universal)

kayabaNerve: Is Docker a build requirement, which it hasn't been so far, a run requirement, which isn't exemplified here when the final comments are on running Serai directly, or a test requirement? If the latter, I'd prefer distinguishing it, along with noting the necessity for regtest Monero and forge/anvil.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+- Ubuntu LTS or WSL2 w/ Docker (Other distributions may also work or docker can be used)
+    - Windows Guide: [https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43](https://dev.to/davidkou/running-docker-in-windows-subsystem-for-linux-wsl2-1k43)
+    - [https://docs.docker.com/engine/install/ubuntu/](https://docs.docker.com/engine/install/ubuntu/)
+- Setup Rust & Dependencies (Assuming Linux or WSL2)
+  - Install Rust  `curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh`
+  
+## Install Solidity Compiler
+```
+     sudo pip3 install solc-select
+     solc-select install 0.8.16
```
kayabaNerve: We need to decide on a solc version. I'll make a tracking issue.

docs/Getting Started.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Build
+```
+    cd serai
+    cargo build --release
+```
+
+## Run
+```
+./target/release/serai-node --chain dev
```
kayabaNerve, line 51: Why not `--dev` directly? Is there a difference?

docs/Getting Started.md
kayabaNerve: This is also not on a newline within the code block, yet then the last ones here don't have any ident at all :thinking: I'd be fine with dropping all identing and just using code on its own lines.

kayabaNerve (changes requested):

kayabaNerve:
Possibly related https://github.com/serai-dex/serai/pull/77.

docs/Getting Started.md
TheArchitect108: Will put it in the docker guide instead of here, but yes, we should link any guides that make it easier for devs to get started.

docs/Getting Started.md
TheArchitect108: now we have OR --dev

docs/Getting Started.md
TheArchitect108: Pretty sure build errors led to me adding these dependencies versus me keeping them for no reason. Can check again on a fresh install? Otherwise let's keep them and get feedback from devs as they go through the process.

docs/Getting Started.md
TheArchitect108: I think the refactor resolves this and commented about moving docker info to another document when complete then linking it in this guide.

docs/Getting Started.md
kayabaNerve: It could be the GitHub CI provides them, but I do recognize these as Monero depends. The Dockerfile PR had the report that it needed clang, not previously specified, due to GH CI provision yet you also didn't include that lol. So not sure.

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
...
+brew install rustup
+```
+
+### Install Solidity Compiler
+```
+sudo pip3 install solc-select
+solc-select install 0.8.16
+solc-select use 0.8.16
+```
+### Install Other Dependencies
```
kayabaNerve, line 34: These are Debian/Ubuntu specific other-dependencies (if actually required, as discussed in another issue).

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
...
+### Build
+```
+cd serai
+cargo build --release
+```
+
+### Run
+```
+./target/release/serai-node --chain dev
+OR
```
kayabaNerve, line 52: ... I mean, is there a need to write a command two different ways? It does demonstrate the chain flag, I guess. Use two different code blocks and ACK.

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
+# Getting Started
+
+## Clone Serai
```
kayabaNerve, line 3: Clone Serai should be after dependencies so we can delinerate environment steps from Serai steps.

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
+# Getting Started
+
+## Clone Serai
+```
+git clone https://github.com/serai-dex/serai.git
+```
+
+## Build and Run Serai
```
kayabaNerve, line 8: This isn't a helpful header when it then includes:

- Rust
- Solidity
- Random system packages
- Build
- Run
- Help

Which is the entire document.

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+#### Mac
+
+```
+brew install rustup
+```
+
+### Install Solidity Compiler
+```
```
kayabaNerve, line 29: No new-lines before code block, as with some other sections, yet also no new-line after :/

docs/Getting Started.md
```diff
@@ -0,0 +1,59 @@
...
+### Install Rust
+
+#### Linux
+```
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
+```
+
+#### Windows
+
+Use WSL2 and Linux instructions.
```
kayabaNerve, line 19: Do we care to even mention Windows at this point? We can just officially support Linux and if someone says "But I run Windows" say "WSL2 at your own risk". I know, rolling stone from my previous comments. I just truly don't care about Windows development for server-software.

kayabaNerve (changes requested):

kayabaNerve:
I just have nitpicks on formatting tbh. Thanks for doing this, and yes, it looks good overall :) Let me know if you'd rather I merge and lint myself instead.

TheNavigator22:
OpenSuse installation instruction
add ethereum client => https://snapcraft.io/install/go-eth/opensuse
add cmake => https://snapcraft.io/cmake

kayabaNerve:
What was geth needed for? I remember something about it having some tool yet it's not used in the CI and I can't remember off the top of my head.

Also, is the official way to install things on OpenSuse via Snap? It doesn't have a native package manager?

TheNavigator22:
no it is not the official, but for those dependencies, this is what I found.
I think the go-eth is actually not needed.
Regarding cmake I used that installation, but it is also available through the native package manager yast.

I suggest to remove my previous OpenSuse remark for the moment.

PR #92, opened 2022-08-17
implement Router.sol and associated functions

### LINKED ISSUE #49: Ethereum Router - Out
Thanks to @noot, we now have a Ethereum Schnorr contract. Our final contract still needs two features:

- Ability to receive funds in.
- Ability to send transactions out.

This issue is specifically for the latter, with a follow up issue being for the former.

By taking in a series of address, calldata pairs, we can execute an arbitrary list of transactions. We can only check a single signature for all transactions if either:
- We accumulate a hash, where It's `accum = keccak256(abi.encodePacked(accum, address, data))`.
- We do `keccak256(abi.encode(addresses, datas))`.
- We use Yul's `calldatasize` and `calldatacopy` to avoid `abi.encode` entirely. This has a slight degree of malleability, as data could be put after the calldata and included here in this hash, yet then it'd be seeking a hash collision so this shouldn't be an issue.

I assume the last one will be cheapest, yet wanted to note the ideas I had so we can ensure this has minimal costs.

Due to using a Schnorr signature, someone must also publish this transaction. Accordingly, they should either:
A) Get a refund for doing so, from the contract. This would have the Schnorr signature also specify a gas price, verifying it's the one actually used by `gasprice`. This would create a leaderless scenario.
B) This can be left out-of-contract to some sort of DAO, yet I'd like for it to be in the protocol (under A).

noot:
- implement `Router.sol` which allows for a group public key to be set/updated and for a list of arbitrary transactions to be executed
- implement related rust code for deploying/calling contract functions 
- add unit tests
closes #49


coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        owner = msg.sender;
+    }
+
+    modifier onlyOwner() {
+        require(msg.sender == owner, "must be called by the contract owner");
+        _;
+    }
+
+    modifier noReentrant() {
+        require(locked == 0, "no re-entrancy");
```
kayabaNerve: This is sub-optimal IIRC. The gas refund from deleting locked is < the gas refund from restoring the original value when the initial set additionally allocates the memory. OZ's ReentrancyGuard discusses this.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+            (bool success, ) = transactions[i].to.call{value: transactions[i].value, gas: transactions[i].gas}(
+                transactions[i].data
+            );
+            emit Executed(nonce, i, success);
+            allOk = success && allOk;
+        }
+        nonce++;
+        return allOk;
+    }
+}
```
kayabaNerve: No newline :(

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
+//SPDX-License-Identifier: AGPLv3
+pragma solidity ^0.8.0;
+
+import "./Schnorr.sol";
+
+contract Router is Schnorr {
+    // contract owner
+    address owner;
```
kayabaNerve: ... do we want to replace this with just a PublicKey? I understand why the address form exists. It's just a variable we immediately kill off yet have to deal with FUD for (it has an owner!!!). I'd kinda rather overload a PublicKey variable (magic parity which will be considered invalid, yet signals px is an address for setPublicKey?). This also would save a minute amount of gas.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+
+    // nonce is incremented for each batch of transactions executed
+    uint256 nonce; 
+
+    // prevents re-entrancy
+    uint8 internal locked;
+
+    struct PublicKey {
+        uint8 parity;
```
kayabaNerve: What are the gas savings if this is always even/odd? Asking as Bitcoin already requires a fixed parity (would have to check which). Carrying that to Ethereum wouldn't be an issue, and it saves passing around a word.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+pragma solidity ^0.8.0;
+
+import "./Schnorr.sol";
+
+contract Router is Schnorr {
+    // contract owner
+    address owner;
+
+    // nonce is incremented for each batch of transactions executed
+    uint256 nonce;
```
kayabaNerve: public

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+    // prevents re-entrancy
+    uint8 internal locked;
+
+    struct PublicKey {
+        uint8 parity;
+        bytes32 px;
+    }
+
+    // current aggregated validator public key 
+    PublicKey publicKey;
```
kayabaNerve: public, unless Solidity can't do public structs

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    modifier noReentrant() {
+        require(locked == 0, "no re-entrancy");
+        locked = 1;
+        _;
+        locked = 0;
+    }
+
+    function getNonce() external view returns (uint256) {
```
kayabaNerve: \- for public

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        return nonce;
+    }
+
+    // setPublicKey can be called by the contract owner to set the current public key,
+    // only if the public key has not been set.
+    function setPublicKey(
+        PublicKey memory _publicKey
+    ) public onlyOwner {
+        require(publicKey.px == 0, "public key has already been set");
+        publicKey.parity = _publicKey.parity;
```
kayabaNerve: Can we not assign the struct as a whole?

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        publicKey.px = _publicKey.px;
+    }
+
+    // updatePublicKey validates the given Schnorr signature against the current public key,
+    // and if successful, updates the contract's public key to the given one.
+    function updatePublicKey(
+        PublicKey memory _publicKey,
+        Signature memory sig
+    ) public {
+        bytes32 message = keccak256(abi.encode(_publicKey.parity, _publicKey.px));
```
kayabaNerve: Add a DST, please. Should also be secure with encodePacked?

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        require(verify(publicKey.parity, publicKey.px, message, sig.s, sig.e), "failed to verify signature");
```
kayabaNerve: Modern Solidity errors should be cheaper to work with.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        uint8 parity;
+        bytes32 px;
+    }
+
+    // current aggregated validator public key 
+    PublicKey publicKey;
+
+    struct Transaction {
+        address to;
+        uint256 value;
```
kayabaNerve: uint64, which is what Substrate will use, would minimize memory yet require an additional multiplication to re-expand. Should still be cheaper.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        uint256 gas;
```
kayabaNerve: This can be a uint32 to create a complete word with the above. A 4 billion gas limit also shouldn't be an issue.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+    // execute accepts a list of transactions to execute as well as a Schnorr signature.
+    // if signature verification passes, the given transactions are executed.
+    // if signature verification fails, this function will revert.
+    // if any of the executed transactions fail, this function will return false but *not* revert.
+    // if all the executed transactions succeed, this function returns true.
+    function execute(
+        Transaction[] calldata transactions, 
+        Signature memory sig
+    ) public noReentrant returns (bool) {
+        bytes32 message = keccak256(abi.encode(nonce, transactions));
```
kayabaNerve: Preferably DST'd, either additionally encode'd OR encodePacked'd with the nonce (not sure which will be cheaper).

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        require(verify(publicKey.parity, publicKey.px, message, sig.s, sig.e), "failed to verify signature");
```
kayabaNerve: s, e? e, s, please.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+        bool allOk = true;
+        for(uint256 i = 0; i < transactions.length; i++) {
+            (bool success, ) = transactions[i].to.call{value: transactions[i].value, gas: transactions[i].gas}(
+                transactions[i].data
+            );
+            emit Executed(nonce, i, success);
```
kayabaNerve: Events aren't cheap. A single event after the loop of nonce with a bitvec of statuses would be much better.

coins/ethereum/contracts/Router.sol
```diff
@@ -0,0 +1,101 @@
...
+            allOk = success && allOk;
+        }
+        nonce++;
+        return allOk;
```
kayabaNerve: There's no reason to return this as this will likely be a top level call.

As a side note, the safest way to implement gas refunds for the publisher, without them calling other SCs using Serai's ETH, is to mandate EOA (though there's an EIP open which may change this and force gas calculation regardless...).

coins/ethereum/tests/router_contract.rs
```diff
@@ -0,0 +1,78 @@
...
+  let client = Arc::new(SignerMiddleware::new(provider, wallet));
+
+  // deploy and set public key
+  let contract = deploy_router_contract(client.clone()).await.unwrap();
+  let public_key = PublicKey::new(&group_key);
+  router_set_public_key(&contract, &public_key).await.unwrap();
+
+  let to = H160([0u8; 20]);
+  let value = U256([0u64; 4]);
+  let gas = U256([1, 0, 0, 0]); // arbitrary
```
kayabaNerve: U256::from(x). Please don't access its raw limbs :(

kayabaNerve (changes requested):
Looks good for an initial router :D Thanks. Mainly focused on discussion and some nits/minor opts.

coins/ethereum/contracts/Router.sol
noot: updated to use OZ ReentrancyGuard

coins/ethereum/contracts/Router.sol
noot: right, so do you mean replace this type with PublicKey or just remove it entirely and set the `PublicKey publicKey` in the constructor? I tried setting it in the constructor, but due to limitations with ethers-rs I was unable to actually call the constructor and pass in the keys :// might be able to now, since I believe they made some updates, can try again

coins/ethereum/contracts/Router.sol
noot: I suppose we'd save an SSTORE there - could modify to force it to be fixed if you'd like. but then I suppose we'd have to force the group key to always have the right parity somehow?

coins/ethereum/contracts/Router.sol
noot: why public vs external? it's not called within this contract

coins/ethereum/contracts/Router.sol
noot: it doesn't seem to get set if I just set the struct (maybe bc it's declared as memory)? seems weird though, I can retest

coins/ethereum/contracts/Router.sol
noot: DST? but yes will change to encodePacked

coins/ethereum/contracts/Router.sol
noot: ether-rs didn't have a uint64 type last I checked :/ was forced to use uint256

noot: https://docs.rs/ethers-core/latest/ethers_core/abi/token/enum.Token.html

coins/ethereum/contracts/ReentrancyGuard.sol
```diff
@@ -0,0 +1,69 @@
...
+        // Any calls to nonReentrant after this point will fail
+        _status = _ENTERED;
+    }
+
+    function _nonReentrantAfter() private {
+        // By storing the original value once again, a refund is triggered (see
+        // https://eips.ethereum.org/EIPS/eip-2200)
+        _status = _NOT_ENTERED;
+    }
+}
```
kayabaNerve: Missing newline, which may be from OZ. This does state it's MIT and gives credit? But I think it also needs the full license text.

I'd hate to suggest using npm to version this, yet a submodule may make sense?

coins/ethereum/contracts/Router.sol
kayabaNerve: Right. We can offset by one until it has parity, which we're doing for BTC. It's incredibly minor on our end and incredibly minor here as well. I don't care to further discuss it ad infinitum.

coins/ethereum/contracts/Router.sol
kayabaNerve: ... huh. I wonder if it's a technicality since all arguments are padded to the size of the words. Would a FixedArray work?

coins/ethereum/contracts/Router.sol
kayabaNerve: We can remove this function since nonce is available via nonce, since the variable is public. That was my comment.

noot: done

coins/ethereum/contracts/Router.sol
noot: ah I left it as `encode`  not `encodePacked` because the `RTransaction[] calldata` type isn't supported by `encodePacked`. I think dynamic arrays aren't allowed

coins/ethereum/contracts/Router.sol
noot: updated

coins/ethereum/contracts/ReentrancyGuard.sol
noot: how do you feel about using forge to version this? it's all rust :D would rather not add any js

noot: would also allow us not to copy pasta the contract

coins/ethereum/contracts/Router.sol
noot: updated, idk if you had a preference for 27 or 28, I just put 27

coins/ethereum/contracts/Router.sol
noot: updated

coins/ethereum/contracts/Router.sol
noot: updated

coins/ethereum/contracts/Router.sol
noot: `FixedArray` is for a vec of Tokens, could do `FixedBytes`, that'd probably work. shall I try it?

coins/ethereum/contracts/ReentrancyGuard.sol
kayabaNerve: Perfectly fine with me, if it works

PR #114, opened 2022-09-01
Cluster Orchestration with Docker Compose

TheArchitect108:
Not quite ready to merge. Opening PR for initial feedback --

- [x] Build Serai Image
- [x] Build Bitcoin Image
- [x] Build Monero Image
- [x] Build Ethereum Image
- [x] Serai Entrypoint
- [x] Bitcoin Entrypoint
- [x] Monero Entrypoint
- [x] Ethereum Entrypoint
- [x] Profiles
- [ ] Substrate Frontend Template
- [ ] Process Manager
- [ ] Processors

As discussed with @kayabaNerve, the current implementation sets a single node (Alice) as authority until we are able to finish consensus.


.gitattributes
```diff
@@ -0,0 +1,3 @@
+# Auto detect text files and perform LF normalization
+* text=auto
+* text eol=lf
```
kayabaNerve: Missing NL @ EOF

deploy/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+We utilize compose profiles to easily orchestrate various pieces of the infrastructure.
+
+**Example:** `docker compose --profile cluster-coins-sm up`
+
+All commands are assumed to be ran from the deploy folder, not the serai root folder.
+
+### Profiles:
+* base - single node, named base
+* coins - node clients for coins only (XMR, BTC, ETH)
```
kayabaNerve: Sort BTC, ETH, XMR please. I'd also prefer this work incrementally, so coins, base, cluster, cluster-coins, yet you let me know if you prefer the various coins be buried as if they're of little importance.

deploy/README.md
```diff
@@ -0,0 +1,50 @@
...
+* cluster-coins-lg - cluster-lg with coins
+* monero - full node monero only
+* bitcoin - full node bitcoin only
+* ethereum - full node ethereum only
+
+You can supply one or more profiles to the docker compose command to orchestrate the desired components.
+
+## Orchestration Approach
+### Builds
+The Serai node is the only piece of our infrastructure that we compile locally and for it first build, it can take 10 minutes or more to complete the image. Images for external coins use download binaries, verifying the signatures and checksums of the build. Overall the standard image build process looks like:
```
kayabaNerve: *its
*downloaded

deploy/coins/bitcoin/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+# Get Binary
+RUN wget https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz \
+	&& wget https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/SHA256SUMS \
+	&& wget https://bitcoincore.org/bin/bitcoin-core-${BITCOIN_VERSION}/SHA256SUMS.asc
+
+# Verify Binary
+# For the PGP keys of builders and developers, refer to 
+# https://github.com/bitcoin/bitcoin/tree/master/contrib/builder-keys 
+ENV KEYS 152812300785C96444D3334D17565732E08E5E41 0AD83877C1F0CD1EE9BD660AD7CC770B81FD22A8 590B7292695AFFA5B672CBB2E13FC145CD3F4304 28F5900B1BB5D1A4B6B6D1A9ED357015286A333D 637DB1E23370F84AFF88CCE03152347D07DA627C CFB16E21C950F67FA95E558F2EEB9F5CC09526C1 F4FC70F07310028424EFC20A8E4256593F177720 D1DBF2C4B96F2DEBF4C16654410108112E7EA81F 287AE4CA1187C68C08B49CB2D11BD4F33F1DB499 F9A8737BF4FF5C89C903DF31DD78544CF91B1514 9DEAE0DC7063249FB05474681E4AED62986CD25D E463A93F5F3117EEDE6C7316BD02942421F4889F 9D3CC86A72F8494342EA5FD10A41BDC3F4FAFF1C 4DAF18FE948E7A965B30F9457E296D555E7F63A7 28E72909F1717FE9607754F8A7BEB2621678D37D 74E2DEF5D77260B98BC19438099BAD163C70FBFA
```
kayabaNerve: What's the actual criteria for these? It looks like you grabbed ~20 from a list of 57. Is this then set up to only require a single signer of 20 to be considered valid?

deploy/coins/bitcoin/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+&& mv lib/* /lib \
+&& mv share/* /share
+
+COPY ./scripts /scripts
+
+EXPOSE 8332 8333 18332 18333 18443 18444
+VOLUME ["/home/bitcoin/.bitcoin"]
+
+# Run
+CMD ["bitcoind"]
```
kayabaNerve, line 55: Docker ignorance on my end, sorry, but how does this get arguments passed?

deploy/coins/bitcoin/scripts/entry-dev.sh
```diff
@@ -0,0 +1,29 @@
...
+bitcoind -regtest -txindex -fallbackfee=0.000001 -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS -rpcallowip=0.0.0.0/0 -rpcbind=127.0.0.1 -rpcbind=$(hostname) &
+
+# give time to bitcoind to start
+while true
+do
+	bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS generatetoaddress 100 $MINER && break
+	sleep 5
+done
+
+bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS createwallet "serai" false false $RPC_PASS false false true &&
```
kayabaNerve: Can we name this miner? Just to be clear it's not Serai's wallet which it's independently managing.

deploy/coins/ethereum/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+	&& apk --no-cache add ca-certificates gnupg bash su-exec
+
+# Get Binary
+RUN wget https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-${GETH_VERSION}-d901d853.tar.gz\
+	&& wget https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-${GETH_VERSION}-d901d853.tar.gz.asc
+
+# Verify Binary
+# refer to https://geth.ethereum.org/downloads/#openpgp_signatures
+# for the PGP keys of builders and developers
+ENV KEYS F9585DE6 C2FF8BBF 9BA28146 7B9E2481 D2A67EAC E058A81C 05A5DDF0 1CCB7DD2
```
kayabaNerve: 9BA28146 - Linux build
E058A81C - Felix Lange
05A5DDF0 - Martin Holst Swende
1CCB7DD2 - Péter Szilágyi

NACK to including random build servers as well, unless they are co-signing for some reason.

Also, what's the security of 4 byte fingerprints? What happens if the Ubuntu key-server has two keys with this fingerprint? Will it error due to ambiguity?

deploy/coins/ethereum/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+RUN gpg --keyserver hkp://keyserver.ubuntu.com:80 --recv-keys ${KEYS} \
+	&& gpg --verify geth-linux-amd64-${GETH_VERSION}-d901d853.tar.gz.asc geth-linux-amd64-${GETH_VERSION}-d901d853.tar.gz
```
kayabaNerve: Can we enforce its signed by the build server and at least 1 developer? Using 2/3 may be better yet I'm not sure how actively the devs sign.

deploy/coins/ethereum/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+
+ENV GETH_VERSION=1.10.23
+
+WORKDIR /home/ethereum
+
+RUN apk update \
+	&& apk --no-cache add ca-certificates gnupg bash su-exec
+
+# Get Binary
+RUN wget https://gethstore.blob.core.windows.net/builds/geth-linux-amd64-${GETH_VERSION}-d901d853.tar.gz\
```
kayabaNerve: Version should be version + commit tag, not just version.

deploy/coins/bitcoin/scripts/entry-dev.sh
```diff
@@ -0,0 +1,29 @@
...
+bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS walletpassphrase $RPC_PASS 60 &&
+bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS importprivkey $PRIV_KEY
+
+# mine a new block every BLOCK_TIME
+while true
+do
+	bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS generatetoaddress 1 $MINER
+	sleep $BLOCK_TIME
+done
```
kayabaNerve: Missing NL

deploy/coins/ethereum/scripts/entry-dev.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+geth --dev --dev.period 5 --verbosity 2 --networkid 15 --datadir "data" -mine --miner.threads 1 -http --http.addr 0.0.0.0 --http.port 8545 --allow-insecure-unlock --http.api "eth,net,web3,miner,personal,txpool,debug" --http.corsdomain "*" -nodiscover --http.vhosts="*"
```
kayabaNerve: Missing NL.

What's the provenance on 15? I see it on a StackExchange Q but no further details. It doesn't appear to conflict with anything relevant but hardhat and foundry use 31337 for development purposes, so I'd rather use that which is more acknowledged.

Also, this seems fine for a dev environment, yet in general seems to be extremely accessible regarding bind + cors + etc. I know its a dev env, behind Docker, yet still would appreciate if it can be tightened.

deploy/coins/monero/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+WORKDIR /home/monero
+
+RUN apk update \
+	&& apk --no-cache add ca-certificates gnupg bash su-exec
+
+# Get Binary
+RUN wget https://downloads.getmonero.org/cli/monero-linux-x64-v${MONERO_VERSION}.tar.bz2
+
+# Verify Binary -- fingerprints from https://github.com/monero-project/monero-site/issues/811
+ENV KEYS F0AF4D462A0BDF92 7455C5E3C0CDCEB9
```
kayabaNerve: These keys aren't present there. Source?

deploy/coins/monero/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+
+# Cleanup
+RUN tar -xvjf monero-linux-x64-v${MONERO_VERSION}.tar.bz2 --strip-components=1
+
+# Prepare Image
+FROM ubuntu:latest as image
+
+WORKDIR /home/monero
+COPY --from=builder /home/monero/* .
+RUN mv * /bin/
```
kayabaNerve, line 36: Do we care to only move relevant binaries? Monero ships like 10 of them and AFAIK we need 1 for prod, and +1 for test.

deploy/coins/monero/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+COPY ./scripts /scripts
+
+EXPOSE 18080 18081
+VOLUME /home/monero/.bitmonero
+
+CMD ["monerod"]
+
```
kayabaNerve: Too many new lines 0_o

deploy/coins/monero/scripts/entry-dev.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# Setup Environment
+RPC_USER="${RPC_USER:=serai}"
+RPC_PASS="${RPC_PASS:=seraidex}"
+MINER="${MINER:=xmraddr}"
+BLOCK_TIME=${BLOCK_TIME:=5}
+
+
+# Run Monero
+monerod --regtest --rpc-login ${RPC_USER}:${RPC_PASS} \
```
kayabaNerve, line 10: `./monerod --regtest --offline --fixed-difficulty=1 --detach`

is sufficient if the tests control block production. If we want tests which have a block every few seconds, I'd rather the test spawn an async test to do that on its end. That way it can still run tests where it does control block timing.

Also, your miner command will spam block addition as fast as it can process. NACK for that.

There's also no need for RPC access controls, except perhaps the foreign IP ones, UNLESS the foreign IP makes Monero mandates a login. My preferred method would then be to override the login, unless its an external (as in, external to Docker) access control measure. Then my question is on the security of a static login and why this is externally accessible.

deploy/coins/monero/scripts/entry-dev.sh
```diff
@@ -0,0 +1,22 @@
...
+--rpc-access-control-origins * --rpc-bind-ip=0.0.0.0 --offline \
+--fixed-difficulty=1 --non-interactive --start-mining ${MINER} \
+--mining-threads 1 --bg-mining-enable --detach
+
+# give time to monerod to start
+while true; do
+    sleep 5
+done
```
kayabaNerve, line 18: No break condition/incomplete script, I'm sure you know, just flagging it.

deploy/docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+  serai-dave:
+  serai-eve:
+  serai-ferdie:
+
+    # ------------------------------------------------------------
+    # Services
+    # ------------------------------------------------------------
+    # Profiles:
+    # base - single node, unidentified
+    # coins - node clients for coins only (XMR, BTC, ETH)
```
kayabaNerve: Same comments on order from the start of this review.

deploy/docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+    # cluter-lg - Alice, Bob, Charlie, Dave, Eve, Ferdie
+    # cluster-coins-sm - cluster-sm with coins
+    # cluster-coins-lg - cluster-lg with coins
+    # monero - full node monero only
+    # bitcoin - full node bitcoin only
+    # ethereum - full node ethereum only
+    # Example: docker compose --profile bitcoin up
+
+services:
+  _serai:
```
kayabaNerve, line 29: What's this for? I don't recognize the _ as syntax and it's not used as a symbol by any other service.

deploy/serai/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN solc-select install 0.8.16
+RUN solc-select use 0.8.16
+
+# Build it
+RUN cargo build --release
+
+# Mount for Cache
+RUN --mount=type=cache,target=/usr/local/cargo/registry \
+    --mount=type=cache,target=/serai/target
+
```
kayabaNerve, line 29: Double newlines?

deploy/serai/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# Copy necessary files to run node
+COPY --from=builder /serai/target/release/* /bin/
+COPY --from=builder /serai/AGPL-3.0 .
+
+# Run node
+
+EXPOSE 30333 9615 9933 9944
+CMD ["serai-node"]
```
kayabaNerve: Missing NL.

deploy/serai/scripts/entry-dev.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+if [[ -z $VALIDATOR ]]; then
+    serai-node --chain $CHAIN --name $NAME
+else
+    serai-node --chain $CHAIN --name $NAME --validator
+fi
```
kayabaNerve: Missing NL.

kayabaNerve (changes requested):
LGTM overall :D

deploy/coins/bitcoin/Dockerfile
TheArchitect108: The only criteria is that they are on the list provided by core. Can add more or less but these are common on other dockerfiles as well.

deploy/coins/bitcoin/Dockerfile
TheArchitect108: Anything you append to the docker run command I.E `docker run -it bitcoin arg` will be added to this entrypoint when using the default image. We override this with compose by calling the shell script.

deploy/README.md
TheArchitect108: updated, like this?

deploy/docker-compose.yml
TheArchitect108: denotes this is a template service, the others inherit the properties of this.

deploy/coins/bitcoin/Dockerfile
kayabaNerve: 1) Do we only require a single signer?
2) Is the list by core core, or is it a list by core? If its effectively a keyserver than we shouldn't carry this bad practice. It also doesn't explain why the other 37 were excluded.

deploy/docker-compose.yml
```diff
@@ -0,0 +1,174 @@
...
+volumes:
+  serai-base:
+  serai-alice:
+  serai-bob:
+  serai-charlie:
+  serai-dave:
+  serai-eve:
+  serai-ferdie:
+
+    # ------------------------------------------------------------
```
kayabaNerve: Random ident?

deploy/coins/bitcoin/Dockerfile
TheArchitect108: We need to have the key of the signer imported when we verify. These are the most relevant signers for bitcoin. We can adjust this to grab and iterate through https://raw.githubusercontent.com/bitcoin/bitcoin/master/contrib/builder-keys/keys.txt, but it's not necessary and would slow down build times. The practice here is correct, each developer publishes their keys independently to various key servers then the footprints are published across various sources.

deploy/docker-compose.yml
TheArchitect108: This is correct for yml. Only at the top level can a comment be left without indents.

TheArchitect108:
@kayabaNerve do we want to merge this, then spawn another PR to include the PM and processors when we're ready for that extension?

Anything additional we need for our local monero / eth dev nodes? -- cc @noot 

The bitcoin environment makes the coins that are mined available for use with the the cli. I think geth handles this automatically but this still needs to be done for monero if behavior is wanted.

deploy/coins/bitcoin/Dockerfile
kayabaNerve: That doesn't answer either of my questions. I fully understand we need these keys. My questions are:

1. Do we only require a single signer?
2. What were the criteria for these 57 people? With Monero, generally anyone who contributes can be on the list. Their presence is not an endorsement of them. As humans, we need to review to decide who we trust.
3. What were the criteria for the 20 you selected? If we're picking some over others, I'd like to know why. If it's a list of Bitcoin contributors, without strict requirements on inclusion, we should be picking a subset yet we need to discuss that subset.

deploy/docker-compose.yml
kayabaNerve: Thanks for clarifying.

deploy/coins/bitcoin/scripts/entry-dev.sh
```diff
@@ -0,0 +1,29 @@
...
+PRIV_KEY="${PRIV_KEY:=cV9X6E3J9jq7R1XR8uPED2JqFxqcd6KrC8XWPy1GchZj7MA7G9Wx}"
+BLOCK_TIME=${BLOCK_TIME:=5}
+
+bitcoind -regtest -txindex -fallbackfee=0.000001 -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS -rpcallowip=0.0.0.0/0 -rpcbind=127.0.0.1 -rpcbind=$(hostname) &
+
+# give time to bitcoind to start
+while true
+do
+	bitcoin-cli -regtest -rpcuser=$RPC_USER -rpcpassword=$RPC_PASS generatetoaddress 100 $MINER && break
+	sleep 5
```
kayabaNerve, line 17: When I pull up this raw file, it claims it's a one space ident? Is that correct?

kayabaNerve:
Requirements to merge:
- Finish discussing all PGP key inclusions as I consider that a security issue (not in that they're a security issue right now, but that the topic is a security issue and requires the proper focus).

What I'd appreciate being done in this PR:
- Not mining a block regularly. That needs to be an additional script/flag (or able to be disabled with one). The Monero/processor tests expect control over the block timing/creation. Though yes, from a local devnet perspective, they should be auto-mining to a usable wallet.
- Removing the infinite loop in `deploy/coins/monero/scripts/entry-dev.sh` and completing the task from the prior bullet point. I'd also need to have @noot comment on geth devnet behavior/setup.
- Misc tweaks and tightening, yet none are so severe I consider them blockers.

So after the PGP discussion, if this has to be merged, we can.

kayabaNerve:
clippy idiocy external to this, fixed in https://github.com/serai-dex/serai/commit/31b64b30828390510dfa63626152016d7b16d42c.

deploy/coins/bitcoin/Dockerfile
TheArchitect108: 1. Yes, we need to have the key for the single signer imported when we verify afaik bitcoin does not multisign builds at this time.
2 & 3. No particular criteria, these keys were verified to be on the list of core builders and found to be common on multiple other docker files. Bitcoin recommends iterating / importing the list in entirety. If they feel the full list is secure, then our subset only adds to security in this way.

TheArchitect108:
> Requirements to merge:
> 
> * Finish discussing all PGP key inclusions as I consider that a security issue (not in that they're a security issue right now, but that the topic is a security issue and requires the proper focus).
> 
> What I'd appreciate being done in this PR:
> 
> * Not mining a block regularly. That needs to be an additional script/flag (or able to be disabled with one). The Monero/processor tests expect control over the block timing/creation. Though yes, from a local devnet perspective, they should be auto-mining to a usable wallet.
> * Removing the infinite loop in `deploy/coins/monero/scripts/entry-dev.sh` and completing the task from the prior bullet point. I'd also need to have @noot comment on geth devnet behavior/setup.
> * Misc tweaks and tightening, yet none are so severe I consider them blockers.
> 
> So after the PGP discussion, if this has to be merged, we can.

I recommend we create additional bash scripts for tests versus trying to add flags / conditionals here with logic. These scripts are meant for local development and we should adjust behavior as such. Easy to add a script for test, any other specific behavior you want for monero?

deploy/coins/bitcoin/Dockerfile
TheArchitect108: After digging deeper, they have a more than merrier approach with signing releases. https://bitcoincore.org/bin/bitcoin-core-23.0/SHA256SUMS.asc contains 17 keys, so to pass all our checks we need at least those. Signers seemingly can vary release to release so it may be better include all 57 to prevent failures. For 23.0 these are fine though.

deploy/coins/bitcoin/Dockerfile
```diff
@@ -3,34 +3,51 @@ FROM alpine:latest as builder
...
-	# removing problematic sigs from SHA256SUMS.asc (using sed)
-	&& sed -i -e '156,162d' SHA256SUMS.asc \ 
-	&& gpg --verify SHA256SUMS.asc SHA256SUMS \
-	&& grep bitcoin-${BITCOIN_VERSION}-x86_64-linux-gnu.tar.gz SHA256SUMS | sha256sum -c
+	&& gpg --keyserver hkp://keys.openpgp.org:80 --recv-keys ${KEYS2}
+
+# Capture error output and count good signature occurrence
+# if above threshold continue
+# else exit code 2
+RUN { err="$( { gpg --verify --status-fd 1 --verify SHA256SUMS.asc SHA256SUMS; } 2>&1 >&3 3>&- )"; } 3>&1; num="$(echo $err | grep -o 'Good signature' | wc -l)" \
```
kayabaNerve: This is vulnerable to keys literally named "Good signature". It is also vulnerable to signature replay to bypass the threshold with just a single party.

`awk '/-----BEGIN PGP SIGNATURE-----/{n++}{print >"sig" n ".txt" }' SHA256SUMS.asc` will split the signatures and let us work on them as individuals since gpg seems unequipped to deal with a list (however incompetent that is). We'd need to count how many of those validate WITH distinct emails (or by deleting a key which passes so it doesn't successfully pass again).

kayabaNerve (changes requested):

onion-route:
Why don't we just check the hashes of the downloaded files?
That makes verification a lot simpler and we don't have to trust obscure GPG keys.
Every release we would then simply have to update the download-URL, verify the files are legit (this involves checking signatures) and then update the hash to be checked against.
So in the Docker scripts we would never have to deal with GPG keys in any way and therefor make things a lot more secure.

kayabaNerve:
While I do have a lot of comments remaining on security tightening, and preferences, this seems functional and is a step forward enabling us to locally run the full suite as needed which is great. Thanks @TheArchitect108.

*These comments were discussed privately, which is part of why I'm making this decision. I'd appreciate if you responded to the above comment just to be transparent there ;)

TheArchitect108:
> Why don't we just check the hashes of the downloaded files? That makes verification a lot simpler and we don't have to trust obscure GPG keys. Every release we would then simply have to update the download-URL, verify the files are legit (this involves checking signatures) and then update the hash to be checked against. So in the Docker scripts we would never have to deal with GPG keys in any way and therefor make things a lot more secure.

I'm not sure that with a manual approach we gain additional security here, automated signature/checksum verification of binaries is common practice. Am a big advocate for keeping automation where we can, the small manual todos add up in a codebase and relaying/preserving the knowledge adds complexity. With automation in this instance only one line needs to be changed and no additional manual steps are necessary for the reviewers, simplifying things a small amount.

onion-route:
I would argue that higher security would be worth the small additional overhead. For one thing, the risk of release binaries being tampered with is greatly reduced, because they are protected by the hash. For another, it would be harder to make mistakes when updating the set of trusted PGP keys.

However, I can see that the security aspect may not be of highest priority during development.

TheArchitect108:
> I would argue that higher security would be worth the small additional overhead. For one thing, the risk of release binaries being tampered with is greatly reduced, because they are protected by the hash. For another, it would be harder to make mistakes when updating the set of trusted PGP keys.
> 
> However, I can see that the security aspect may not be of highest priority during development.

Generally, the issuer of the binary should be trusted versus shifting the trust to us, a third party. I see how changes to the checksum or link on a regular basis could create an easier avenue for a social vulnerability than updates only to the human readable version numbers. Many entities follow this verification convention for binaries, including packaging systems. No sense in reinventing the wheel here imo, there are layers to this process that prevent compromise.

Issue #47, opened 2022-07-17
Serai Consensus


kayabaNerve:
The required effects of consensus are firmly detailed in https://github.com/serai-dex/serai/blob/develop/docs/protocol/Consensus.md. While it's not perfect, namely already describing itself as a modified Aura, it does describe the needed UX.

Substrate offers [Inherent transactions](https://docs.substrate.io/main-docs/fundamentals/transaction-types/#inherent-transactions), which, combined with per-block finalization, should hit all bases. The question becomes:

1) Can we properly validate inherent transactions? It seems like we'll be able to suitably populate a DB.
2) Can we bypass inherent transaction finalization if a block has been confirmed by the BFT process? It looks like nodes will currently always reject a block if it has invalid inherents.
3) Do we work with Aura, BABE + Grandpa, a fork of them, or another algorithm entirely? The template suggested it was using Aura + Grandpa, not just Aura, while Grandpa isn't immediately feasible due to the lack of per-block finalization. Aura doesn't use a VRF to decide the next producer, which enables DoSs, however.

kayabaNerve:
Relation to https://github.com/serai-dex/serai/issues/6, which tracks a different aspect.

kayabaNerve:
So, at its core:

- Block production needs to bypass inherent checks if the block is already finalized.
- Block finality needs to come per-block.

GRANDPA could potentially be forked for the latter, yet we need to merge it with Aura/BABE which has to be forked to bypass checks. I'd be potentially more interested in implementing Tendermint, which is dead simple, instead, with the caveat that Tendermint has the same problem as Aura regarding producer selection.

Immediately, the work is solely to move from the current accept-any to something, even if it's a no-fault-tolerant PoA setup.

TheArchitect108:
@kayabaNerve let's outline something temporary for protonet that will get us moving towards local cluster setups then open a new testnet focused issue to focus on finalizing our ultimate approach.

kayabaNerve:
Immediate is just the accept-any with a single node config. When we do have a multi-node setup, we can do no fault tolerant PoA where every node does a round robin, as I said. Then for mainnet, Tendermint with a randomized selection instead of a linear one or BABE + GRANDPA.

kayabaNerve:
[...]

Issue #137, opened 2022-10-22
Resolving serai_consensus into a proper sc_tendermint


kayabaNerve:
As of now, there's Tendermint Machine, an implementation of the academic protocol left isolated for not only universal use, yet rigorous testing without complications.

This is being bound into Substrate via serai_consensus, which also takes the liberty of defining the entire block production pipeline with specific types. This issue is meant to discussion resolution of various non-Substrate patterns present in serai_consensus, where if all were resolved, this crate could even be offered under Substrate itself.

The benefit to such would be the wider eyes we'd gain our work, potentially including support/review from other parties.

### Usage of a custom Substrate fork

This is tracked in #135.

### Unification of BlockImport, JustificationImport, and Verifier

This was done as the straightest way forward since Tendermint details block production and finalization. There remains benefit to more modularity as with it comes flexibility. This may already happen before the tendermint branch is merged.

### Lack of use of the traditional block pipeline

Instead of receiving blocks with seals, we use the Tendermint process to receive blocks and the temporal Tendermint proposal message (along with the current clock) acts as the seal. While we could use the existing process, it'd require moving the propose message handling out of the Tendermint machine, and break that desired bound. While it may be fine to maintain this custom pipeline, doing so breaks interoperability within Substrate, meaning we can likely never create a fully-Substrate-esque crate.

### Custom handling of inherent transactions

We bypass inherent transaction checks when a block is finalized. A Parity developer explicitly told me inherent transactions were not intended for the flow used in Serai, and this flow which enables their usage as such may not be deemed Substrate-esque. It could be argued as a valid optimization though and harmless to Parity's intention for inherent transactions.

### Strict handling of the import pipeline

We ban post-digests and auxiliary data. While Substrate itself will error if auxiliary data remains at the end, we don't allow it when the Tendermint importer is called. This was intended to protect against unforeseen cases/malleability and ensure security, yet may break interoperability (though I'm unsure what such interoperability would be).

### Children on-chain with a lower priority than their parents

In order to fully validate blocks, we must import them, yet imported blocks (under PoW/BABE/Aura) are intended to built upon. Under Tendermint, only the finalized block should be. The workaround was to import with a lower preference than the chain tip, then trigger moving to them by finalizing when the time comes. This can't be resolved with the `SelectChain` API, mainly used by nodes (I don't see it used within Substrate), yet highlights incompatibilities with the Substrate design and Tendermint.

### Forced usage of Ristretto and Ristretto alone.

[...]

PR #145, opened 2022-11-02
Initial Tendermint implementation

kayabaNerve:
Doesn't handle slashing nor staking.

With the new docker composition, the 6 nodes (3 validators, 3 nodes) are able to connect, produce blocks, and achieve finalizations over extended periods of time. I'm currently running such a config overnight to do initial tests on its stability.

Achieves the necessary handling of inherent transactions, enabling closing https://github.com/serai-dex/serai/issues/47 for a series of smaller issues.


kayabaNerve:
Created #146 #147 to enable closing #47.

kayabaNerve:
This halted while building block #294, AKA after roughly half an hour. It looks to have done so with heavily dropping peer counts.

```
serai-dev-serai-ferdie-1   | 2022-11-02 08:03:17 ✨ Imported #293 (0x0a48…d965)    
serai-dev-serai-eve-1      | 2022-11-02 08:03:17 💤 Idle (4 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 791.7kiB/s ⬆ 0.3kiB/s    
serai-dev-serai-charlie-1  | 2022-11-02 08:03:17 💤 Idle (5 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 487.8kiB/s ⬆ 1.5MiB/s    
serai-dev-serai-alice-1    | 2022-11-02 08:03:20 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 122.4kiB/s ⬆ 305.1kiB/s    
serai-dev-serai-dave-1     | 2022-11-02 08:03:21 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 90.3kiB/s ⬆ 0.3kiB/s    
serai-dev-serai-ferdie-1   | 2022-11-02 08:03:21 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 0.7kiB/s ⬆ 0.2kiB/s    
serai-dev-serai-bob-1      | 2022-11-02 08:03:21 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 0.8kiB/s ⬆ 1.2kiB/s    
serai-dev-serai-eve-1      | 2022-11-02 08:03:22 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 49 B/s ⬆ 49 B/s    
serai-dev-serai-charlie-1  | 2022-11-02 08:03:22 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 30 B/s ⬆ 28.8kiB/s    
serai-dev-serai-alice-1    | 2022-11-02 08:03:25 💤 Idle (1 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 2.4kiB/s ⬆ 2.5kiB/s    
serai-dev-serai-dave-1     | 2022-11-02 08:03:26 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.6kiB/s ⬆ 1.7kiB/s    
serai-dev-serai-ferdie-1   | 2022-11-02 08:03:26 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.8kiB/s ⬆ 1.8kiB/s    
serai-dev-serai-bob-1      | 2022-11-02 08:03:26 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 3.0kiB/s ⬆ 3.0kiB/s    
serai-dev-serai-eve-1      | 2022-11-02 08:03:27 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.9kiB/s ⬆ 1.9kiB/s    
serai-dev-serai-charlie-1  | 2022-11-02 08:03:27 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 3.4kiB/s ⬆ 3.4kiB/s    
serai-dev-serai-alice-1    | 2022-11-02 08:03:30 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 2.5kiB/s ⬆ 2.4kiB/s    
serai-dev-serai-dave-1     | 2022-11-02 08:03:31 💤 Idle (2 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.6kiB/s ⬆ 1.7kiB/s    
serai-dev-serai-ferdie-1   | 2022-11-02 08:03:31 💤 Idle (4 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.4kiB/s ⬆ 1.4kiB/s    
serai-dev-serai-bob-1      | 2022-11-02 08:03:31 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 2.8kiB/s ⬆ 2.7kiB/s    
serai-dev-serai-eve-1      | 2022-11-02 08:03:32 💤 Idle (3 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.0kiB/s ⬆ 1.0kiB/s    
serai-dev-serai-charlie-1  | 2022-11-02 08:03:32 💤 Idle (1 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 3.8kiB/s ⬆ 3.8kiB/s    
serai-dev-serai-alice-1    | 2022-11-02 08:03:35 💤 Idle (0 peers), best: #293 (0x0a48…d965), finalized #293 (0x0a48…d965), ⬇ 1.3kiB/s ⬆ 1.3kiB/s   
```

I believe this may recover naturally, yet this is still a failure. I don't believe this should re-arrange its peers at all and assume its doing so based on some *idea* a peer is malicious (though libp2p may just be doing libp2p stuff). I'll try to conduct further review.

EDIT: Despite healthy peer counts now, this won't recover as the precommit timeout is only scheduled after receiving enough precommits. If the net was offline, and old messages weren't replayed when coming back online, that would cause this failure. More robust P2P code would fix this, and an analysis of why peer counts dropped would prevent it.

I still consider this branch a candidate for merge.

kayabaNerve:
New code was able to create a stable net for 24h, and presumably indefinitely. While it was locally tested, offering ideal network conditions, the fault tolerance was 0 and my computer did exhibit significant strain, creating some level of erratic timing. With more nodes, the fault tolerance would go up.

kayabaNerve:
CI will fail for a bit since the addition of the step channel. That hasn't been integrated into Substrate yet as I'm currently working on refactoring the machine for ease of development (without introducing invariants) and security.

kayabaNerve:
This PR leaves 9 TODOs in the `tendermint` folder, of varying significance.

1) Support a genesis time. Right now, it just sets the nearest 5s moment to boot as the 'genesis'. Docker has enough of a discrepancy when starting nodes they booted with different genesis times (when using the current time). This seems to work reliably and has yet to fail. This should be moved to an argument from Docker. Tracked in #155.

2) This may not be able to handle syncing when also configured as a validator. If the node syncs a block via Substrate, the Tendermint machine needs to be updated to work off of the block which was synced externally. In testing, I do not believe the node ever synced a justification it was working on producing before it created one itself, yet this works to recover from unexpected networking situations and to handle the initial sync process. The issue is when initially syncing, we may sync blocks at an extremely high rate. sc-tendermint will create a block proposal for each block stepped past, and while there shouldn't be anything in the mempool making this take time, it *may* take time causing a backlog. Once that backlog exceeds a certain point (256 blocks), the state for the block in question will be pruned, and creating a proposal will fail :/ Tracked in #157.

3) Slashing is marked `todo!()`. This requires integrating staking (#146).

4) A misc note on BlockOrigin: `BlockOrigin::ConsensusBroadcast, // TODO: Use BlockOrigin::Own when it's our block`

5) This test TODO which I don't believe will occur, yet needs a test written to confirm.
```rs
  // TODO: Is there a DoS where you send a block without justifications, causing it to error,
  // yet adding it to the blacklist in the process preventing further syncing?
```

6) A TODO to call `lock_import_and_run` on our end, which makes us less fragile/prevents a race condition (which I think Substrate already handles? yet we should handle it on our end to be sure). Tracked in #158.

7) Integrate pallet-staking. This means proper calculation of weights by sc-tendermint. Tracked in #146.

8) Implement a better round robin. The current one is a simple, unweighted round robin, without context on the amount of rounds previous blocks used. Tracked in #159.

9) An additional misc note on the machine on improving the code hygiene.

None of these must be handled now to move forward with development. All which are meaningful have tracking issues. Since this branch is relied on by multiple others, I would like it to be merged into `develop` now. The above will then be handled at protonet/testnet/whenever.

substrate/node/src/service.rs
```diff
@@ -65,36 +144,55 @@ pub fn new_partial(config: &Configuration) -> Result<PartialComponents, ServiceE
...
+      keystore_container,
+      select_chain: _,
+      other: mut telemetry,
+      transaction_pool,
+    },
+  ) = new_partial(&config)?;
+
+  let is_authority = config.role.is_authority();
+  let tendermint_protocol = sc_tendermint::protocol_name(
+    client.block_hash(0).unwrap().unwrap(),
```
TheArchitect108: Let's explicitly map this above -- `let genesis_hash = client.block_hash(0).ok().flatten().expect("Genesis block exists; qed");`

substrate/node/src/service.rs
```diff
@@ -147,14 +243,32 @@ pub fn new_full(config: Configuration) -> Result<TaskManager, ServiceError> {
...
-  if role.is_authority() {
-    serai_consensus::authority(
-      &task_manager,
-      client,
-      network,
-      transaction_pool,
-      select_chain,
-      prometheus_registry.as_ref(),
+  if is_authority {
+    task_manager.spawn_essential_handle().spawn(
```
TheArchitect108, line 251: This would look prettier if we moved this back to serai_consensus. Any reason to import TM directly over a consensus interface?

substrate/tendermint/client/src/authority/gossip.rs
```diff
@@ -0,0 +1,67 @@
...
+      return ValidationResult::Discard;
+    }
+
+    // Verify the signature here so we don't carry invalid messages in our gossip layer
+    // This will cause double verification of the signature, yet that's a minimal cost
+    if !msg.verify_signature(&self.signature_scheme) {
+      return ValidationResult::Discard;
+    }
+
+    ValidationResult::ProcessAndKeep(Self::topic(msg.number().0))
```
TheArchitect108: `msg.verify_signature(&self.signature_scheme) && !msg.number().0 < *self.number.read().unwrap()` for `ValidationResult::ProcessAndKeep(Self::topic(msg.number().0))`? Would rather the default for this call tree be ValidationResult::Discard having ProcessAndKeep guarded explicitly by the if statement.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    mut self,
+    protocol: ProtocolName,
+    keys: Arc<dyn CryptoStore>,
+    providers: T::CIDP,
+    spawner: impl SpawnEssentialNamed,
+    env: T::Environment,
+    network: T::Network,
+    registry: Option<&Registry>,
+  ) {
+    let (best_hash, last) = self.get_last();
```
TheArchitect108: any cons to splitting this to an explicit get_best and get_last?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+}
+
+/// Tendermint Authority. Participates in the block proposal and voting process.
+pub struct TendermintAuthority<T: TendermintValidator> {
+  genesis: Option<u64>,
+  import: TendermintImport<T>,
+  active: Option<ActiveAuthority<T>>,
+}
+
+async fn get_proposal<T: TendermintValidator>(
```
TheArchitect108, line 86: Curious why this is get_proposal versus propose_block?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+  stub: bool,
+) -> T::Block {
+  let proposer =
+    env.lock().await.init(header).await.expect("Failed to create a proposer for the new block");
+
+  proposer
+    .propose(
+      import.inherent_data(*header.parent_hash()).await,
+      Digest::default(),
+      if stub {
```
TheArchitect108, line 99: Why stub? Behavior is unclear from name and comment. Duration::Zero seems low for anything.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    protocol: ProtocolName,
+    keys: Arc<dyn CryptoStore>,
+    providers: T::CIDP,
+    spawner: impl SpawnEssentialNamed,
+    env: T::Environment,
+    network: T::Network,
+    registry: Option<&Registry>,
+  ) {
+    let (best_hash, last) = self.get_last();
+    let new_number = last.0 .0 + 1;
```
TheArchitect108: Can we call this new_block or new_block_number?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+      network.clone(),
+      protocol,
+      Arc::new(TendermintGossip::new(number.clone(), self.import.validators.clone())),
+      registry,
+    );
+
+    // This should only have a single value, yet a bounded channel with a capacity of 1 would cause
+    // a firm bound. It's not worth having a backlog crash the node since we aren't constrained
+    let (new_number_tx, mut new_number_rx) = mpsc::unbounded();
+    let (gossip_tx, mut gossip_rx) = mpsc::unbounded();
```
TheArchitect108: Can we break the tx/rx convention here and name them send/receive or gossip_in, gossip_out?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+            // Both wait for a write lock on this number and don't release it until after updating
+            // it accordingly
+            {
+              let mut number = number.write().unwrap();
+              if this_number < *number {
+                continue;
+              }
+              let new_number = this_number + 1;
+              *number = new_number;
+              recv = gossip.messages_for(TendermintGossip::<T>::topic(new_number));
```
TheArchitect108: new_gossip_channel instead of new_number?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+              // TODO: Set true when initial syncing
+              get_proposal(&env, &import, &notif.header, false).await
+            )).await.unwrap();
+          } else {
+            break;
+          }
+        },
+
+        // Machine reached a new height
+        new_number = new_number_rx.next() => {
```
TheArchitect108: block_gossip_channel?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+          } else {
+            break;
+          }
+        },
+
+        // Received a message
+        msg = recv.next() => {
+          if let Some(msg) = msg {
+            messages.send(
+              match SignedMessage::decode(&mut msg.message.as_ref()) {
```
TheArchitect108, line 317: Self note: review validation of libp2p / signed message decoding since we call it directly from an external message.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    let mut gossip = GossipEngine::new(
+      network.clone(),
+      protocol,
+      Arc::new(TendermintGossip::new(number.clone(), self.import.validators.clone())),
+      registry,
+    );
+
+    // This should only have a single value, yet a bounded channel with a capacity of 1 would cause
+    // a firm bound. It's not worth having a backlog crash the node since we aren't constrained
+    let (new_number_tx, mut new_number_rx) = mpsc::unbounded();
```
TheArchitect108: From @VRX -- Given that the scope of these channels are limited, could we use bounded channels? What are the advantages of unbounded in this context?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+  type Block = T::Block;
+
+  const BLOCK_PROCESSING_TIME: u32 = T::BLOCK_PROCESSING_TIME_IN_SECONDS;
+  const LATENCY_TIME: u32 = T::LATENCY_TIME_IN_SECONDS;
+
+  fn signer(&self) -> TendermintSigner<T> {
+    self.active.as_ref().unwrap().signer.clone()
+  }
+
+  fn signature_scheme(&self) -> TendermintValidators<T> {
```
TheArchitect108, line 355: signature_party?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    self.import.validators.clone()
+  }
+
+  fn weights(&self) -> TendermintValidators<T> {
```
TheArchitect108, line 359: temporary until API? Mark todo.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+  // When Tendermint completes, the block is finalized, setting it as the tip regardless of work.
+  async fn validate(&mut self, block: &T::Block) -> Result<(), BlockError> {
+    let hash = block.hash();
+    let (header, body) = block.clone().deconstruct();
+    let parent = *header.parent_hash();
+    let number = *header.number();
+
+    // Can happen when we sync a block while also acting as a validator
+    if number <= self.import.client.info().best_number {
+      Err(BlockError::Temporal)?;
```
TheArchitect108, line 402: from @vrx00 should we include more specific errors for these cases? This would help with logging.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    let mut queue_write = self.import.queue.write().await;
+    *self.import.importing_block.write().unwrap() = Some(hash);
+
+    queue_write.as_mut().unwrap().import_blocks(
+      BlockOrigin::ConsensusBroadcast, // TODO: Use BlockOrigin::Own when it's our block
```
TheArchitect108, line 409: This seems lowhanging, why not add before merge?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+      vec![IncomingBlock {
+        hash,
+        header: Some(header),
+        body: Some(body),
+        indexed_body: None,
+        justifications: None,
+        origin: None,
```
TheArchitect108: Why none here?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    // Sanity checks that a child block can have less work than its parent
+    {
+      let info = self.import.client.info();
+      assert_eq!(info.best_hash, parent);
+      assert_eq!(info.finalized_hash, parent);
+      assert_eq!(info.best_number, number - 1u8.into());
+      assert_eq!(info.finalized_number, number - 1u8.into());
+    }
+
+    Ok(())
```
TheArchitect108, line 435: I don't like return success by default things, any way we can better guard this?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+  // would be incredibly unsafe thanks to the unchecked inherents, it's defined as a tip with less
+  // work, despite being a child of some parent. This means it won't be moved to nor operated on by
+  // the node.
+  //
+  // When Tendermint completes, the block is finalized, setting it as the tip regardless of work.
+  async fn validate(&mut self, block: &T::Block) -> Result<(), BlockError> {
+    let hash = block.hash();
+    let (header, body) = block.clone().deconstruct();
+    let parent = *header.parent_hash();
+    let number = *header.number();
```
TheArchitect108, line 397: proposed_number

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    Ok(())
+  }
+
+  async fn add_block(
+    &mut self,
+    block: T::Block,
+    commit: Commit<TendermintValidators<T>>,
+  ) -> T::Block {
+    let hash = block.hash();
+    let justification = (CONSENSUS_ID, commit.encode());
```
TheArchitect108: Can you elaborate on CONSENSUS_ID? Not seeing anywhere this changes and it seems to be a hardcoded value.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    debug_assert!(self.import.verify_justification(hash, &justification).is_ok());
+
+    let raw_number = *block.header().number();
+    let this_number: u64 = match raw_number.try_into() {
```
TheArchitect108: the frequent usage of number here makes this section of code difficult to track, can we refactor using unified conventions? proposed_block_number, current_block_number, casted_block_number?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+    let new_number = this_number + 1;
+
+    {
+      let active = self.active.as_mut().unwrap();
+      // Acquire the write lock
+      let mut number = active.new_number.write().unwrap();
+
+      // This block's number may not be the block we're working on if Substrate synced it before
+      // the machine synced the necessary precommits
+      if this_number == *number {
```
TheArchitect108: if current_number == assigned_number?

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+          warn!(
+            target: "tendermint",
+            "Attempted to send a new number to the gossip handler except it's closed. {}",
+            "Is the node shutting down?"
+          );
+        }
+      }
+
+      // Clear any blocks for the previous height we were willing to recheck
+      *self.import.recheck.write().unwrap() = HashSet::new();
```
TheArchitect108: recheck_queue?

substrate/tendermint/client/src/block_import.rs
```diff
@@ -0,0 +1,186 @@
...
+
+impl<T: TendermintValidator> TendermintImport<T> {
+  fn check_already_in_chain(&self, hash: <T::Block as Block>::Hash) -> bool {
+    let id = BlockId::Hash(hash);
+    // If it's in chain, with justifications, return it's already on chain
+    // If it's in chain, without justifications, continue the block import process to import its
+    // justifications
+    // This can be triggered if the validators add a block, without justifications, yet the p2p
+    // process then broadcasts it with its justifications
+    if (self.client.status(id).unwrap() == BlockStatus::InChain) &&
```
TheArchitect108: can we wrap the whole conditional too?

substrate/tendermint/client/src/block_import.rs
```diff
@@ -0,0 +1,186 @@
...
+      self.client.justifications(hash).unwrap().is_some()
+    {
+      return true;
+    }
+    false
```
TheArchitect108: return false; bc line 27. I understand this is a rust thing, but if we are returning other things in the function we should explicitly return.

substrate/tendermint/client/src/block_import.rs
```diff
@@ -0,0 +1,186 @@
...
+
+use sc_consensus::{BlockCheckParams, BlockImportParams, ImportResult, BlockImport, Verifier};
+
+use sc_client_api::{Backend, BlockBackend};
+
+use crate::{TendermintValidator, tendermint::TendermintImport};
+
+impl<T: TendermintValidator> TendermintImport<T> {
+  fn check_already_in_chain(&self, hash: <T::Block as Block>::Hash) -> bool {
+    let id = BlockId::Hash(hash);
```
TheArchitect108, line 18: block_id -- we may have multiple ids in a given function so should be explicit.

substrate/tendermint/client/src/block_import.rs
```diff
@@ -0,0 +1,186 @@
...
+where
+  Arc<T::Client>: BlockImport<T::Block, Transaction = T::BackendTransaction>,
+  <Arc<T::Client> as BlockImport<T::Block>>::Error: Into<Error>,
+{
+  type Error = Error;
+  type Transaction = T::BackendTransaction;
+
+  // TODO: Is there a DoS where you send a block without justifications, causing it to error,
+  // yet adding it to the blacklist in the process preventing further syncing?
+  async fn check_block(
```
TheArchitect108, line 40: Where is this called/used?

substrate/tendermint/client/src/block_import.rs
```diff
@@ -0,0 +1,186 @@
...
+      // condition since Substrate will eventually prune the block's state, potentially before
+      // indexing finishes when syncing)
+      //
+      // The alternative to this would be editing Substrate directly, which would be a lot less
+      // fragile, manually triggering the notifications (which may be possible with code intended
+      // for testing), writing our own notification system, or implementing lock_import_and_run
+      // on our end, letting us directly set the notifications, so we're not beholden to when
+      // Substrate decides to call notify_finalized
+      //
+      // TODO: Call lock_import_and_run on our end, which already may be needed for safety reasons
```
TheArchitect108: Not familiar enough to comment on the decision to remap all of these to NetworkBroadcast. Could you elaborate on the "safety reasons"?

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+  pub(crate) fn verify_order(
+    &self,
+    parent: <T::Block as Block>::Hash,
+    number: <<T::Block as Block>::Header as Header>::Number,
+  ) -> Result<(), Error> {
+    let info = self.client.info();
+    if (info.finalized_hash != parent) || ((info.finalized_number + 1u16.into()) != number) {
+      Err(Error::Other("non-sequential import".into()))?;
+    }
+    Ok(())
```
TheArchitect108, line 149: default return error, return guarded Ok

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+    if justification.0 != CONSENSUS_ID {
+      Err(Error::InvalidJustification)?;
+    }
+
+    let commit: Commit<TendermintValidators<T>> =
+      Commit::decode(&mut justification.1.as_ref()).map_err(|_| Error::InvalidJustification)?;
+    if !TendermintAuthority::new(None, self.clone()).verify_commit(hash, &commit) {
+      Err(Error::InvalidJustification)?;
+    }
+    Ok(())
```
TheArchitect108, line 182: Error default, guard Ok

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+  }
+
+  // Verifies the justifications aren't malformed, not that the block is justified
+  // Errors if justifications is neither empty nor a sinlge Tendermint justification
```
TheArchitect108: single

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+        let mut iter = justifications.iter();
+        let next = iter.next();
+        if next.is_none() || iter.next().is_some() {
+          Err(Error::InvalidJustification)?;
+        }
+        self.verify_justification(block.header.hash(), next.unwrap())?;
+        block.finalized = true;
+      }
+    }
+    Ok(())
```
TheArchitect108, line 203: guard Ok

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+    block: &mut BlockImportParams<T::Block, BT>,
+  ) -> Result<(), Error> {
+    if block.finalized {
+      if block.fork_choice != Some(ForkChoiceStrategy::Custom(false)) {
+        // Since we alw1ays set the fork choice, this means something else marked the block as
+        // finalized, which shouldn't be possible. Ensuring nothing else is setting blocks as
+        // finalized helps ensure our security
+        panic!("block was finalized despite not setting the fork choice");
+      }
+      return Ok(());
```
TheArchitect108, line 217: guard ok

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+    Ok(())
+  }
+
+  pub(crate) async fn check<BT>(
+    &mut self,
+    block: &mut BlockImportParams<T::Block, BT>,
+  ) -> Result<(), Error> {
+    if block.finalized {
+      if block.fork_choice != Some(ForkChoiceStrategy::Custom(false)) {
```
TheArchitect108, line 211: self.enabled && block.fork_choice

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+        // finalized helps ensure our security
+        panic!("block was finalized despite not setting the fork choice");
+      }
+      return Ok(());
+    }
+
+    // Set the block as a worse choice
+    block.fork_choice = Some(ForkChoiceStrategy::Custom(false));
+
+    self.verify_order(*block.header.parent_hash(), *block.header.number())?;
```
TheArchitect108, line 223: Shouldn't we do something if these error?

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+    self.verify_justifications(block)?;
```
TheArchitect108, line 224: ditto ^

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+    // They *should* be unused, so requiring their emptiness prevents malleability and ensures
+    // nothing slips through
+    if !block.post_digests.is_empty() {
+      Err(Error::Other("post-digests included".into()))?;
+    }
+    if !block.auxiliary.is_empty() {
+      Err(Error::Other("auxiliary included".into()))?;
+    }
+
+    Ok(())
```
TheArchitect108, line 245: Guard Ok, error default

substrate/tendermint/client/src/validators.rs
```diff
@@ -0,0 +1,195 @@
...
+
+  total_weight: u64,
+  weights: Vec<u64>,
+
+  lookup: Vec<Public>,
+}
+
+impl TendermintValidatorsStruct {
+  fn from_module<T: TendermintClient>(client: &Arc<T::Client>) -> Self {
+    let last = client.info().finalized_hash;
```
TheArchitect108, line 35: last_block_hash?

substrate/tendermint/client/src/validators.rs
```diff
@@ -0,0 +1,195 @@
...
+  client: Arc<T::Client>,
+  _refresh: Arc<RwLock<TendermintValidatorsStruct>>,
+}
+
+impl<T: TendermintClient> Refresh<T> {
+  // If the session has changed, re-create the struct with the data on it
+  fn refresh(&self) {
+    let session = self._refresh.read().unwrap().session;
+    if session !=
+      self
```
TheArchitect108: can we assign this to reduce the line splits?

substrate/tendermint/client/src/validators.rs
```diff
@@ -0,0 +1,195 @@
...
+  fn verify_aggregate(&self, validators: &[u16], msg: &[u8], sigs: &Vec<Signature>) -> bool {
+    if validators.len() != sigs.len() {
+      return false;
+    }
+    for (v, sig) in validators.iter().zip(sigs.iter()) {
+      if !self.verify(*v, msg, sig) {
+        return false;
+      }
+    }
+    true
```
TheArchitect108, line 168: guard true default false

substrate/tendermint/client/src/validators.rs
```diff
@@ -0,0 +1,195 @@
...
+
+  fn total_weight(&self) -> u64 {
+    self.0.read().unwrap().total_weight
+  }
+
+  fn weight(&self, id: u16) -> u64 {
+    self.0.read().unwrap().weights[usize::try_from(id).unwrap()]
+  }
+
+  // TODO
```
TheArchitect108: What is the todo here? any way we can make these blocks easier to read/understand?

substrate/tendermint/machine/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "tendermint-machine"
+version = "0.1.0"
+description = "An implementation of the Tendermint state machine in Rust"
+license = "MIT"
+repository = "https://github.com/serai-dex/serai/tree/develop/substrate/tendermint/machine"
+authors = ["Luke Parker <lukeparker5132@gmail.com>"]
+edition = "2021"
```
TheArchitect108, line 8: 2022

substrate/tendermint/machine/README.md
```diff
@@ -0,0 +1,70 @@
...
+     any consumers must be aware of it. No machine will vote nor precommit to a
+     block it considers invalid, so for a network with an honest majority, this
+     is a non-issue.
+
+### Paper
+
+The [paper](https://arxiv.org/abs/1807.04938) describes the algorithm with
+pseudocode on page 6. This pseudocode is written as a series of conditions for
+advancement. This is extremely archaic, as its a fraction of the actually
+required code. This is due to its hand-waving away of data tracking, lack of
```
TheArchitect108: Do we need this feedback in our readme? maybe a better channel for it.

substrate/tendermint/machine/src/block.rs
```diff
@@ -0,0 +1,127 @@
...
+  // data, if we are the proposer.
+  pub(crate) fn new_round(
+    &mut self,
+    round: RoundNumber,
+    proposer: N::ValidatorId,
+    time: Option<CanonicalInstant>,
+  ) -> Option<DataFor<N>> {
+    debug_assert_eq!(round.0 == 0, time.is_some());
+
+    // If skipping rounds, populate end_time
```
TheArchitect108: Can we document cases skipping might happen? Need more clarity about the need and function of populate_end_time.

substrate/tendermint/machine/src/block.rs
```diff
@@ -0,0 +1,127 @@
...
+    self.end_time.insert(round, self.round().end_time());
+
+    // 14-21
+    if Some(proposer) == self.validator_id {
+      let (round, block) = if let Some((round, block)) = &self.valid {
+        (Some(*round), block.clone())
+      } else {
+        (None, self.proposal.clone())
+      };
+      Some(Data::Proposal(round, block))
```
TheArchitect108, line 114: BroadcastProposal(p, round[p], proposal, validRound[p]) -- why are we only broadcasting round and block?

substrate/tendermint/machine/src/block.rs
```diff
@@ -0,0 +1,127 @@
...
+    } else {
+      self.round_mut().set_timeout(Step::Propose);
```
TheArchitect108, line 116: there are several checks missed before we are supposed to propose on timeout. 

```
OnTimeoutPropose(p) ==
  /\ step[p] = "PROPOSE"
  /\ p /= Proposer[round[p]]
  /\ BroadcastPrevote(p, round[p], NilValue)
  /\ step' = [step EXCEPT ![p] = "PREVOTE"]
  /\ UNCHANGED <<round, lockedValue, lockedRound, validValue,
                validRound, decision, evidence, msgsPropose, msgsPrecommit>>
  /\ action' = "OnTimeoutPropose"
```

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+      }
+      round.0 -= 1;
+    }
+    debug_assert_eq!(self.block.end_time[&round].canonical(), commit.end_time);
+
+    self.reset(round, proposal).await;
+  }
+
+  async fn slash(&mut self, validator: N::ValidatorId) {
+    if !self.block.slashes.contains(&validator) {
```
TheArchitect108, line 232: Why would we not slash if they have already been slashed? Is this to prevent double slashes? How does it get cleared?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+    last: (BlockNumber, u64),
+    proposal: N::Block,
+  ) -> TendermintHandle<N> {
+    let (msg_send, msg_recv) = mpsc::unbounded();
+    let (step_send, step_recv) = mpsc::unbounded();
+    TendermintHandle {
+      step: step_send,
+      messages: msg_send,
+      machine: {
+        let last_time = sys_time(last.1);
```
TheArchitect108: it would be better to have two variables than to map a tuple here. Tuples are hard to read when accessed.

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+        // If the last block hasn't ended yet, sleep until it has
+        sleep(last_time.duration_since(SystemTime::now()).unwrap_or(Duration::ZERO)).await;
```
TheArchitect108: Why provide duration zero, shouldn't we create an exception or panic if systemtime isn't available?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+      let mut queue_future =
+        if self.queue.is_empty() { Fuse::terminated() } else { future::ready(()).fuse() };
+
+      if let Some((broadcast, msg)) = futures::select_biased! {
+        // Handle a new height occuring externally (an external sync loop)
+        // Has the highest priority as it makes all other futures here irrelevant
+        msg = self.step_recv.next() => {
+          if let Some((block_number, commit, proposal)) = msg {
+            // Commit is for a block we've already moved past
+            if block_number != self.block.number {
```
TheArchitect108, line 305: Should we also check temporally?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+        match res {
+          Ok(None) => (),
+          Ok(Some(block)) => {
+            let mut validators = vec![];
+            let mut sigs = vec![];
+            // Get all precommits for this round
+            for (validator, msgs) in &self.block.log.log[&msg.round] {
+              if let Some(Data::Precommit(Some((id, sig)))) = msgs.get(&Step::Precommit) {
+                // If this precommit was for this block, include it
+                if id == &block.id() {
```
TheArchitect108, line 373: block_id

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+              validators,
+              signature: N::SignatureScheme::aggregate(&sigs),
+            };
+            debug_assert!(self.network.verify_commit(block.id(), &commit));
+
+            let proposal = self.network.add_block(block, commit).await;
+            self.reset(msg.round, proposal).await;
+          }
+          Err(TendermintError::Malicious(validator)) => {
+            self.slash(validator).await;
```
TheArchitect108: Do we need something more severe for malice? Shouldn't this trigger a ban vote?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+  // Returns Ok(true) if this was a Precommit which had its signature validated
+  // Returns Ok(false) if it wasn't a Precommit or the signature wasn't validated yet
+  // Returns Err if the signature was invalid
+  fn verify_precommit_signature(
+    &self,
+    sender: N::ValidatorId,
+    round: RoundNumber,
+    data: &DataFor<N>,
+  ) -> Result<bool, TendermintError<N::ValidatorId>> {
+    Ok(if let Data::Precommit(Some((id, sig))) = data {
```
TheArchitect108: This is almost impossible to read, let's split this up.

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+    } else {
+      false
+    })
+  }
+
+  async fn message(
+    &mut self,
+    msg: MessageFor<N>,
+  ) -> Result<Option<N::Block>, TendermintError<N::ValidatorId>> {
+    if msg.number != self.block.number {
```
TheArchitect108: msg.block_number

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+      Err(TendermintError::Temporal)?;
+    }
+
+    // If this is a precommit, verify its signature
+    self.verify_precommit_signature(msg.sender, msg.round, &msg.data)?;
+
+    // Only let the proposer propose
+    if matches!(msg.data, Data::Proposal(..)) &&
```
TheArchitect108, line 439: why the need for a matches macro here?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+    // Run the finalizer to see if it applies
+    // 49-52
+    if matches!(msg.data, Data::Proposal(..)) || matches!(msg.data, Data::Precommit(_)) {
+      let proposer = self.weights.proposer(self.block.number, msg.round);
+
+      // Get the proposal
+      if let Some(Data::Proposal(_, block)) = self.block.log.get(msg.round, proposer, Step::Propose)
+      {
+        // Check if it has gotten a sufficient amount of precommits
+        // Use a junk signature since message equality disregards the signature
```
TheArchitect108, line 461: We could check the signature in has_consensus, yes?

substrate/tendermint/machine/src/lib.rs
```diff
@@ -0,0 +1,620 @@
...
+        self.block.log.message_instances(self.block.round().number, Data::Prevote(None));
+      // 34-35
+      if participation >= self.weights.threshold() {
+        self.block.round_mut().set_timeout(Step::Prevote);
+      }
+
+      // 44-46
+      if weight >= self.weights.threshold() {
+        self.broadcast(Data::Precommit(None));
+        return Ok(None);
```
TheArchitect108, line 530: Should this step the precommit too?

substrate/tendermint/client/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "sc-tendermint"
+version = "0.1.0"
+description = "Tendermint client for Substrate"
+license = "AGPL-3.0-only"
+repository = "https://github.com/serai-dex/serai/tree/develop/substrate/tendermint/client"
+authors = ["Luke Parker <lukeparker5132@gmail.com>"]
+edition = "2021"
```
TheArchitect108, line 8: 2022

substrate/tendermint/machine/src/message_log.rs
```diff
@@ -0,0 +1,100 @@
...
+      if let Some(prev) = self.precommitted.get(&msg.sender) {
+        if hash != prev {
+          Err(TendermintError::Malicious(msg.sender))?;
+        }
+      }
+      self.precommitted.insert(msg.sender, *hash);
+    }
+
+    msgs.insert(step, msg.data);
+    Ok(true)
```
TheArchitect108, line 52: guard Ok

substrate/tendermint/pallet/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "pallet-tendermint"
+version = "0.1.0"
+description = "Tendermint pallet for Substrate"
+license = "AGPL-3.0-only"
+repository = "https://github.com/serai-dex/serai/tree/develop/substrate/tendermint/pallet"
+authors = ["Luke Parker <lukeparker5132@gmail.com>"]
+edition = "2021"
```
TheArchitect108, line 8: 2022

substrate/tendermint/pallet/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+      pub type AuthorityPair = Pair;
+    }
+    pub type AuthoritySignature = Signature;
+    pub type AuthorityId = Public;
+  }
+
+  impl<T: Config, V> OneSessionHandler<V> for Pallet<T> {
+    type Key = crypto::Public;
+
+    fn on_genesis_session<'a, I: 'a>(_validators: I)
```
TheArchitect108, line 47: The empty functions could benefit from TODOs.

substrate/tendermint/primitives/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "sp-tendermint"
+version = "0.1.0"
+description = "Tendermint primitives for Substrate"
+license = "AGPL-3.0-only"
+repository = "https://github.com/serai-dex/serai/tree/develop/substrate/tendermint/primitives"
+authors = ["Luke Parker <lukeparker5132@gmail.com>"]
+edition = "2021"
```
TheArchitect108, line 8: 2022

TheArchitect108 (commented):
Most comments are nitpicks / minor changes. We have not tested this over any networked environments but locally it seems sound. Logic seems sound when comparing to the paper and other specs, although I am not a TM/Consensus expert.

substrate/tendermint/primitives/Cargo.toml
kayabaNerve: There is no Rust 2022 edition. https://doc.rust-lang.org/edition-guide/editions/index.html

substrate/node/src/service.rs
kayabaNerve: sp/pallet/sc is a pattern from Subtrate, which has relevancy to #137. While we could restore serai_consensus as ~100 lines managing this for us, serai_consensus started as a garbage bin of code I didn't understand/like while attempting to rip out/minimize consensus in order for the eventual rewrite. It's not present in other codebases and I don't care to restore a minimal form of it.

As a side note, the authority function args are ugly and likely should have some args moved into TendermintAuthority::new (which possibly shouldn't return TendermintAuthority yet the async handle authority currently does?).

substrate/tendermint/client/src/authority/gossip.rs
kayabaNerve: ... this is guarded explicitly by the if statement. It's just not nested into it. Not only is that a non-standard pattern for this ecosystem, and one frequently advised against in general, your immediately suggested code isn't possible.

```rs
if msg.verify_signature(&self.signature_scheme) && !msg.number().0 < *self.number.read().unwrap() {
    ValidationResult::ProcessAndKeep(Self::topic(msg.number().0))
}
```
isn't valid as a return as it's incomplete. It'd need to be
```rs
if msg.verify_signature(&self.signature_scheme) && (!msg.number().0 < *self.number.read().unwrap()) {
    ValidationResult::ProcessAndKeep(Self::topic(msg.number().0))
} else {
  ValidationResult::Discard
}
```

This adds more lines, while not easily allowing expanding the conditional. Instead of just inserting a new if, we have to further extend that long conditional/do multiple layers of nesting inside of it. You also suggested combining two separate if statements, which alone isn't an issue, EXCEPT if those conditionals SHOULD have different return values, they can't be nested as such. While here, they both trigger discard, that's not always the case, and why the early return pattern remains the most flexible.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: It solely gets a block to propose. It then passes it off to Tendermint to *potentially* propose.

substrate/tendermint/machine/src/block.rs
kayabaNerve: We aren't.  This is expanded into a Message with the sender/current height/round, etc. This is solely the last two arguments of the above, as the first two are common to all Messages so it was moved out of the enum. The `round` here is the valid round this block was proposed from.

substrate/tendermint/machine/src/block.rs
kayabaNerve: This solely schedules it. `OnTimeoutPropose` is scheduled on line 21 of the paper, and is done if you're not the proposer when starting a round, as done here.` OnTimeoutPropose`,  57-60, is under the `timeout_future` handler. That handler checks the step matches the timeout and then broadcasts the prevote accordingly. Broadcasting the prevote also sets the step to prevote.

My above description aligns properly, albeit skipping over the checks the height and round haven't changed. Any round change will clear all timeouts, offering that property.

I'm not sure what your notation there is.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: It's cleared on a new block, as it's contained under new block. It's intended to prevent general spam issues, since this `slash` presumably will cause the validator to make a TX voting to slash at this height. If a malicious validator spams malicious actions, other validators will flood the net.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: This allows a genesis time in the future.

If the current time is 10, and the block is scheduled for 100, this will do 100 - 10 = 90, sleep for 90.

If the current time is 10, and the block occurred at 0, this will do 0 - 10, and hit the exception. ZERO causes a NOP as it shouldn't sleep in that case.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: Temporally? As in, the age of the message?

The block number is the age of the message, mapped to a usable value without sufficient accuracy, without question other than if the sender is lying. If they are lying, it'll be ignored as past/future, or handled, and if malicious, slashed.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: Running two nodes at once is sufficient to cause a slash. Banning is a bit harsh, albeit possible. I'll note that tendermint-machine should solely slash, and we as Serai should discuss if slash == ban.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: Broadcasting Precommit does step to Precommit.

```rs
  fn broadcast(&mut self, data: DataFor<N>) {
    if let Some(msg) = self.block.message(data) {
      self.queue.push_back(msg);
    }
  }
```

```rs
  // Transform Data into an actual Message, using the contextual data from this block
  pub(crate) fn message(&mut self, data: DataFor<N>) -> Option<MessageFor<N>> {
    debug_assert_eq!(
      self.round().step,
      match data.step() {
        Step::Propose | Step::Prevote => Step::Propose,
        Step::Precommit => Step::Prevote,
      },
    );
    // 27, 33, 41, 46, 60, 64
    self.round_mut().step = data.step();
```

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: ... anything for me to do or can I resolve this? This just performs SCALE decoding of a byte string. Tendermint machine fully validates the signature and then operates on the data, validating it. I don't see it as a problem in any way.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: The comment on this line describes why they're not bounded. They SHOULD only have one value when operating as a live validator. Given this is asynchronous, a backlog can form when blocks are synced faster than this is polled though. #157 touches on this.

unbounded also does not have a meaningful performance penalty. It just risks OOM errors if it grows to big, as a Vec would.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: Considering it's more than just the party, yet also signature aggregation and verification, I do believe it's a scheme. It is a contextual scheme however...

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: BlockError isn't under sc_tendermint. It's under tendermint-machine. While tendermint-machine can provide a series of errors, and a function returning if an error is temporal enough, I'd rather we just add a `log::trace` statement here if your concern is logging.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: Because I have to double check the notification and logging behavior when using `Own`, and I also have to manage a new cache variable if I take the easy way out (just checking hash equivalence, which may leave malleability on the table leading to issues if we mislabel it), or access more information from the Tendermint machine to do this properly (checking we actually proposed). It's also relatively meaningless.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: indexed_body: We don't do indexing.
justifications: It's not justified yet.
origin: It's not the aforementioned BlockOrigin yet actually a LibP2P PeerId. I believe it's intended to be a minor optimization on not sending it back, yet this has no issue at runtime. Possibly worth a TODO? Except then we have to hook into LibP2P AND tendermint-machine to track this...

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: It's not returning success by default. It explicitly re-raises the result of the above ImportFuture. This could just be `ImportFuture::new(hash, queue_write.as_mut().unwrap()).await` if it weren't for the sanity checks.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: ... it's literally the number of the header we're currently dealing with. While yes, it's the number of a proposed block, it's also the block we're currently adding to Substrate which doesn't have a concept of proposals. Solely additions. I do not feel a need to be so verbose with our variable names in this case.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: It's expected by Substrate. It enables multiple finalization algorithms to each apply finalizations. I believe this is intended for usage with GRANDPA and Beefy, where the latter creates efficiently verifiable BLS justifications less frequently.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: It's not a queue. We only recheck them if they come up again. For it to be a queue, it'd presumably be ordered and we'd be rechecking them every so often.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: As in `if ((`? No, Rust warns about un-needed parentheses. It's polite enough to allow wrapping individual conditions for those who don't trust the compilers order of operations though, even though they're technically unnecessary. I personally believe in wrapping every condition/OoO ambiguity, yet don't mind the if statement itself being open.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: That'll raise a warning in clippy for being pointless. If we then do further things, this won't compile for the missing semicolon. If someone adds a semicolon without thinking about it, it'll raise a warning about unused value.

This is how Rust code is written and it will break linters while going against convention if we try to change that.

kayabaNerve: ... why didn't I just write the condition itself as the tail value? We don't need an if statement to return true if true and false if false. That's a change I'd like to make since I don't see this function changing in any notable way. Good with you?

substrate/tendermint/client/src/block_import.rs
kayabaNerve: ... except this is the only ID in this 12 line function. While some enhanced verbosity would be beneficial, I also see it beneficial in areas of confusion. I disagree any given function which may theoretically be evolved over years into something radically different which causes confusion is itself confusing.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: It's part of the BlockImport trait the ImportQueue (BasicQueue) manages.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: I'm mainly concerned about a race condition where we perform checks on a block, await, perform checks on another block, await, then call import on A and import on B. I *believe* Substrate already handles this and it hasn't come up yet in near-0 latency environments, but I wouldn't deploy this to mainnet without certainty.

substrate/tendermint/client/src/tendermint.rs
kayabaNerve: There's no `self.enabled` available. These conditionals also can't be merged. If the block is already finalized, it doesn't re-run checks. It's a caching measure given we call this multiple times on a given block as it's imported through the system. The inner `if` is simply a sanity check which looks for behavior we exhibit when we finalize, to ensure nothing else does, as the comment inside of it explains.

substrate/tendermint/client/src/tendermint.rs
kayabaNerve: There's a question mark at the end of this line. If it errors, the result is raised up accordingly. If we didn't have the question mark, Rust would warn us for dropping a Result without handling it. The arguments, which are function calls, also can't error.

substrate/tendermint/machine/src/block.rs
kayabaNerve: The paper itself describes the case when jumping rounds. I don't feel a need to accordingly.

Documenting populate_end_time makes more sense. A few comments explaining our end_time cache and so on?

substrate/node/src/service.rs
kayabaNerve: I completely disagree with carrying "; qed" everywhere. Panicking if the condition doesn't hold declares the statement should be true. The string explaining why it's true is the explanation. Further stating "qed" is saying, this is true and here's why, and this itself is an explanation justifying it's truthiness. It's redundant to the point advocating "qed" inclusion could lead to advocating for "qed; qed" inclusion. Neither contribute anything and both are equally redundant.

No objection to the variable creation though. Can do.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: I'd be more interested in removing get_last, which only has one caller right now (here), and accordingly getting rid of the tuple. Good with you?

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: The only number this tracks is block numbers. It also isn't a block, solely the number of the next block which is currently being worked on. I feel the former is inaccurate and the latter isn't a helpful contribution when there are no other numbers in the scope.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: new_number is written to `number`, an Arc of the current block being worked on. new_gossip_channel wouldn't be a better name accordingly, IMO.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: This isn't a gossip channel. This is a mpsc channel we use to receive notifications a new number was achieved in this loop.

substrate/tendermint/client/src/validators.rs
kayabaNerve: `last` always refers to the last block in this function (though it arguably conflicts with the last time present elsewhere?). We never obtain the block, so I don't feel a need to differentiate between "block" and "block_hash". While `last_block_hash` would be correct, and `last_block` wouldn't be, I don't feel the need to suffix anything given the lack of immediate competition. I'll review this tomorrow for proximity to last time though.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: Lack of other IDs in this vicinity. We *technically* have validator ID, yet this file solely refers to `signer`. We don't consider the block number nor the round number IDs.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: This just returns Ok so long as nothing in the function errors. The alternative is

```
let res = if { ...
Ok(res)
````

I'm not a fan of the pointless assignment, especially since this isn't causing line breaks, yet I'd do it to squash this.

File: substrate/tendermint/machine/src/lib.rs (hunk 54 above)
Comment (kayabaNerve) [reply substrate/tendermint/machine/src/lib.rs]: `msg.block` acceptable? It parallels `msg.round` which refers to a `RoundNumber`.

File: substrate/tendermint/machine/src/lib.rs (hunk 55 above)
Comment (kayabaNerve) [reply substrate/tendermint/machine/src/lib.rs:439]: Because this check is only executed if this is a Proposal. We could use a match statement, yet that's bad practice when you only care about a single branch.

File: substrate/tendermint/machine/src/lib.rs (hunk 56 above)
Comment (kayabaNerve) [reply substrate/tendermint/machine/src/lib.rs:461]: To be clear, Precommit has an additional signature. This signature isn't used for message authenticity yet rather creating efficient commits. Accordingly, it needs to be authentic, yet isn't relevant to the discussion.

In order to quickly confirm if Precommits were functionally equivalent, Precommit equality was defined as follows:
```rs
      (Data::Precommit(None), Data::Precommit(None)) => true,
      (Data::Precommit(Some((i, _))), Data::Precommit(Some((i2, _)))) => i == i2,
```
The signature is dropped and only the ID is compared (which I called `i` for some reason).

This code section is asking for all precommits matching the ID of the proposed block. In order to do this, we pass in a Data::Precommit to check equality against. Equality doesn't care about the signature, yet to instantiate the struct, we must provide one. That's why it uses a junk signature. We could also decode 0 IIRC.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: To further explain why this is buried here, this function used to be part of Machine::broadcast. When creating a `Block` object, to better isolate data and ensure its temporal nature when intended to be temporal, functions solely using members of Block were moved into it. Some functions, which had sections only using members of Block, were also made independent functions. This is one of those.

It says if you create a message, you step forward to it, as the paper requires in every single case. By consolidating this here, we never have to double check all 6 of these locations has this line correctly implemented. It's a reduction in surface even if yes, it's a bit buried now. I do believe the line numbers are helpful (and sufficient), since those can be searched.

substrate/tendermint/client/src/tendermint.rs
```diff
@@ -0,0 +1,228 @@
...
+
+  // Ensure this is part of a sequential import
+  pub(crate) fn verify_order(
+    &self,
+    parent: <T::Block as Block>::Hash,
+    number: <<T::Block as Block>::Header as Header>::Number,
+  ) -> Result<(), Error> {
+    let info = self.client.info();
+    if (info.finalized_hash != parent) || ((info.finalized_number + 1u16.into()) != number) {
+      Err(Error::Other("non-sequential import".into()))?;
```
vrx00, line 147: it would be helpful to print out (we can use log) the parent and number

substrate/tendermint/client/src/authority/mod.rs
TheArchitect108: Yes, adding logging would be great, goes with VRX comment above about tracking the erroring hash for wrong sequential imports.

substrate/tendermint/client/src/authority/mod.rs
TheArchitect108: 👍 now this is here for future participants

substrate/tendermint/client/src/authority/mod.rs
TheArchitect108: Yea was targeting origin here, thought we might want to provide the origin, especially as we move towards validating from storage?

substrate/tendermint/client/src/authority/mod.rs
TheArchitect108: Queues don't have to be ordered or reprocessed. recheck_list?

substrate/tendermint/client/src/block_import.rs
TheArchitect108: Let's wrap the second condition then please, no need to wrap both.

substrate/tendermint/client/src/block_import.rs
TheArchitect108: I vote verbosity in this context, up to you...

substrate/tendermint/client/src/validators.rs
TheArchitect108: +1 for verbosity.

substrate/tendermint/machine/src/block.rs
TheArchitect108: yeah it would be great to document or comment our understanding/approach to timeouts too 👍

substrate/tendermint/machine/src/block.rs
TheArchitect108: Looking at the spec here -- https://github.com/tendermint/spec/blob/master/spec/light-client/accountability/TendermintAcc_004_draft.tla. May be worth adding the extra checks, they seem simple enough?

substrate/tendermint/machine/src/lib.rs
TheArchitect108: There is potential for multiple slashes in the same block, maybe I am mixing/confusing concepts here and it will not matter, but we do not want to spawn a future bug with this line.

substrate/tendermint/client/src/authority/mod.rs
TheArchitect108: +1verbosity

substrate/tendermint/machine/src/block.rs
TheArchitect108: Can we comment / document the message abstraction here so a reviewer/auditor can quickly understand?

substrate/tendermint/machine/src/lib.rs
TheArchitect108: I mean checking that it's an ordered increase in block number. Only 1 above last, no more no less.

substrate/tendermint/machine/src/lib.rs
TheArchitect108: +verbose

substrate/tendermint/machine/src/lib.rs
TheArchitect108: How will we distinguish the type of slash?

substrate/tendermint/machine/src/lib.rs
TheArchitect108: ++verbose round and block

substrate/tendermint/machine/src/lib.rs
TheArchitect108: Makes it easier to read...

substrate/tendermint/machine/src/lib.rs
TheArchitect108: This is a good explanation, thanks.

substrate/tendermint/machine/src/lib.rs
TheArchitect108: Your call on if we want to get more verbose with `i`. Trust your guidance on the signature piece, thanks for adding wisdom here.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: https://github.com/serai-dex/serai/issues/158 for the relevant issue.

substrate/tendermint/machine/src/block.rs
kayabaNerve: That's not the spec. That's solely the Tendermint company's spec for an instantiation of Tendermint, and not even the up-to-date one at that. I did research that the modern Tendermint deployed under cosmos aligns with the paper, and they have an issue noting a break which we handle properly.

That code block, AFAICT, says on timeout propose, if still step Propose, broadcast prevote None. Then set step to prevote. We do exactly that under the timeout_future. There's no distinction in checks/actions.

The code you commented on is to schedule the timeout. That's not what your cited thing does.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: Yes, but by doing so enables a DoS since someone triggering a slash causes every other validator to vote on that property, working as Nx multiplier. Slashes also solely penalize the malicious actor. They do not effect algorithmic functionality.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: ... That's what this line does. It checks the message's block number is the number of the block currently being worked on. The block currently being worked on is the last block's + 1.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: It's actually not endorsed by Rust and I believe triggers clippy for a single line match.

substrate/tendermint/client/src/block_import.rs
kayabaNerve: The second condition is a single term with no spaces/operators, and parentheses delineate whitespace/operators. Adding parentheses to it would also trigger Rust/clippy.

substrate/tendermint/client/src/tendermint.rs
kayabaNerve: Also adding debugs to all slash conditions.

substrate/tendermint/client/src/authority/mod.rs
```diff
@@ -0,0 +1,453 @@
...
+              // This will fail if syncing occurs radically faster than machine stepping takes
+              // TODO: Set true when initial syncing
+              get_proposal(&env, &import, &notif.header, false).await
+            )).await.unwrap();
+          } else {
+            break;
+          }
+        },
+
+        // Machine reached a new height
```
kayabaNerve: This uses the term "height" which has been agreed not to use.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: This isn't temporary. weights is supposed to return an object which returns weights. This does that. The actual weight information isn't currently provided, yet that's elsewhere, and yes, marked TODO.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: Will add TODO.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: Documented in 9b51eafc5fb038348a3e67baa365ba7930252b0a.

substrate/tendermint/machine/src/lib.rs
kayabaNerve: I'm adding log::debug statements on slash. Slashes can be designated by the height at which they occur thanks to their single occurrence per block. While we could have slash specify the reason from an enum, I don't feel a need to complicate the API as such. It doesn't help with logging if the machine itself has logging.

While we could make the argument for differing slash penalties, all slashes boil down to incompetence. You're either offline, misconfigured, or malicious and incompetent. Accordingly, I don't believe they should have differing slash penalties and we shouldn't complicate the API as such.

substrate/tendermint/client/src/authority/mod.rs
kayabaNerve: Resolving since this is low impact and we have no better name clearly available. It may theoretically be worth merging with weights in the future but...

TheArchitect108:
LGTM 🔥 🔥 🔥 🥳

PR #160, opened 2022-11-16
created shared volume between containers

vrx00:
Genesis can be read from any serai container. It is a local file of a serai container which can be read from local "/genesis" file. Genesis will be renewed whenever a cluster is restarted.


kayabaNerve:
Locally merged into tendermint :)

Issue #166, opened 2022-11-23
Monero transaction library fails intermittently


kayabaNerve:
This has now happened three times, all within the last week or so. The first two happened at the same time when doing a lot of CI work, so I had hoped GitHub had simply killed monerod for resource use. The issue with that idea is that GitHub kills tasks, not processes in tasks. It also happened again now, when running 2 CI runs at once, yet I do believe there's an intermittent error here. I'm not sure what it is.

https://github.com/serai-dex/serai/actions/runs/3467617451/jobs/5792656820

```
running 4 tests
test multisig_send_single_input ... FAILED
test multisig_send_multiple_inputs ... ok
test send_multiple_inputs ... ok
test send_single_input ... ok

failures:

---- multisig_send_single_input stdout ----
thread 'multisig_send_single_input' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidTransaction([136, 197, 242, 44, 231, 80, 252, 242, 23, 116, 110, 33, 18, 209, 131, 116, 88, 125, 52, 161, 46, 157, 86, 61, 158, 2, 251, 107, 171, 201, 45, 6])', coins/monero/tests/send.rs:170:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'multisig_send_single_input' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(1), ...)', coins/monero/tests/send.rs:186:1


failures:
    multisig_send_single_input
```

https://github.com/serai-dex/serai/actions/runs/3467617445/jobs/5792656679

```
running 2 tests
test send_single_input ... FAILED
test send_multiple_inputs ... ok

failures:

error: test failed, to rerun pass `-p monero-serai --test send`
---- send_single_input stdout ----
thread 'send_single_input' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidTransaction([7, 109, 70, 203, 226, 122, 224, 109, 213, 70, 45, 209, 43, 106, 51, 77, 245, 208, 147, 45, 138, 87, 251, 40, 147, 241, 157, 196, 54, 75, 9, 236])', coins/monero/tests/send.rs:170:57
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'send_single_input' panicked at 'called `Result::unwrap()` on an `Err` value: JoinError::Panic(Id(1), ...)', coins/monero/tests/send.rs:175:1


failures:
    send_single_input
```

[...]

PR #172, opened 2022-11-29
fix send tests

akildemir:


TheArchitect108:
Thanks for landing this! Requesting @vrx00 for initial review then pass to @kayabaNerve.

kayabaNerve:
Handling.

.gitignore
```diff
@@ -1 +1,2 @@
 target
+.vscode
```
kayabaNerve, line 2: Please ensure all files have newlines after them

coins/monero/src/rpc.rs
```diff
@@ -477,4 +477,23 @@ impl Rpc {
 
     Ok(())
   }
+
+  pub async fn mine_regtest_blocks(
```
kayabaNerve, line 481: Can we name this generate_blocks to match the RPC, please?

coins/monero/src/rpc.rs
```diff
@@ -477,4 +477,23 @@ impl Rpc {
...
+    &self,
+    address: &str,
+    block_count: u32,
```
kayabaNerve, line 484: usize

coins/monero/src/rpc.rs
```diff
@@ -477,4 +477,23 @@ impl Rpc {
...
+  ) -> Result<EmptyResponse, RpcError> {
```
kayabaNerve, line 485: Result<(), RpcError>

coins/monero/tests/send.rs
```diff
@@ -28,11 +28,9 @@ use monero_serai::{
     address::Network, ViewPair, Scanner, SpendableOutput, SignableTransaction,
     SignableTransactionBuilder,
   },
+  rpc::{Rpc},
```
kayabaNerve, line 31: Pointless {}

coins/monero/tests/send.rs
```diff
@@ -7,19 +7,19 @@ use lazy_static::lazy_static;
...
+use curve25519_dalek::{constants::ED25519_BASEPOINT_TABLE, scalar::Scalar};
+
 #[cfg(feature = "multisig")]
 use blake2::{digest::Update, Digest, Blake2b512};
-
-use curve25519_dalek::constants::ED25519_BASEPOINT_TABLE;
-
 #[cfg(feature = "multisig")]
-use dalek_ff_group::Scalar;
+use dalek_ff_group;
```
kayabaNerve, line 15: This is pointless. You can just use dalek_ff_group::Scalar below without the use. If you're using members of dalek_ff_group regularly, I'd recommend `use dalek_ff_group as dfg;`, yet this here can just be removed.

coins/monero/tests/send.rs
```diff
@@ -7,19 +7,19 @@ use lazy_static::lazy_static;
 use zeroize::Zeroizing;
 use rand_core::OsRng;
 
+use curve25519_dalek::{constants::ED25519_BASEPOINT_TABLE, scalar::Scalar};
+
 #[cfg(feature = "multisig")]
 use blake2::{digest::Update, Digest, Blake2b512};
-
-use curve25519_dalek::constants::ED25519_BASEPOINT_TABLE;
-
```
kayabaNerve, line 14: Please keep this new line between hash functions and ECC impls.

coins/monero/tests/send.rs
```diff
@@ -55,168 +53,305 @@ macro_rules! async_sequential {
...
+}
 
-  if multisig {
-    #[cfg(not(feature = "multisig"))]
-    panic!("Running a multisig test without the multisig feature");
-    #[cfg(feature = "multisig")]
-    {
-      view = Scalar::from_hash(Blake2b512::new().chain("Monero Serai Transaction Test")).0;
-      spend_pub = keys[&1].group_key().0;
+async fn mine_until_unlocked(rpc: &Rpc, addr: &str, tx_hash: [u8; 32]) {
```
kayabaNerve, line 70: Please add:
```
// TODO: Support transactions already on-chain
// TODO: Don't have a side effect of mining blocks more blocks than needed under race conditions
```

coins/monero/tests/send.rs
```diff
@@ -55,168 +53,305 @@ macro_rules! async_sequential {
...
+    // publish the tx
+    rpc.publish_transaction(&tx).await.unwrap();
   }
 }
 
 #[cfg(feature = "multisig")]
 async_sequential! {
   async fn multisig_send_single_input() {
-    send_core(0, true).await;
+    let rpc = Rpc::new("http://127.0.0.1:18081".to_string()).unwrap();
```
kayabaNerve, line 200: This large block of additions and the block below it are largely duplicated code. While the point of this PR is to remove the monolithic function, duplicating it everywhere isn't a clean solution.

We may *need* to have these tests written around TransactionBuilder with a custom test runner to ensure cleanliness. To recap our previous discussions:
- Each test would take in a `&mut Builder` and perform operations
- The test runner would create a Builder, assign it a coinbase output as an input, and pass a builder to modify
- The test runner would sign and confirm it can be broadcasted
- The test runner would then call another function to have whatever desired checks performed (such as, the transaction can then be scanned)

This minimal runner design doesn't work in the long-term however as it's only designed for a single input. We either need:
1) A runner which can take the above design yet run multiple transactions. This would likely have to be a macro.
2) A very suitable set of helper functions combined with some statics.

I'll put forth a commit offering the former as it should be sufficiently brief and within my skill set.

kayabaNerve:
Superseded by https://github.com/serai-dex/serai/commit/8f352353ba2ff8430e7a03a31cd18cce0eb32ef8.

@akildemir Please port any changes you want to keep from this PR (the fix for #166 + any misc cleanups) to a new PR. Thanks :)

PR #181, opened 2022-12-05
cache solc

TheArchitect108:


deploy/serai/Dockerfile
```diff
@@ -18,12 +18,13 @@ WORKDIR /serai
 RUN rustup update
 
 # Install Solc @ 0.8.16
-RUN pip3 install solc-select==0.2.1
-RUN solc-select install 0.8.16
-RUN solc-select use 0.8.16
+RUN --mount=type=cache,target=/root/.cache/ pip3 install solc-select==0.2.1
```
kayabaNerve: Should this be .cache/pip? I'm not against caching everything, it's there for a reason, just wanted to double check.

deploy/serai/Dockerfile
```diff
@@ -18,12 +18,13 @@ WORKDIR /serai
...
+RUN --mount=type=cache,target=/root/.cache/ solc-select install 0.8.16
```
kayabaNerve: solc installs to ~/.local/bin

There's also ~/.solc-select

kayabaNerve (changes requested):

PR #184, opened 2022-12-06
fix for #166 & cleanup

akildemir:


coins/monero/src/rpc.rs
```diff
@@ -477,4 +477,21 @@ impl Rpc {
...
+        Some(json!({
+          "method": "generateblocks",
+          "params": {
+            "wallet_address": address,
+            "amount_of_blocks": block_count
+          },
+        })),
+      )
+      .await
+      .unwrap();
```
kayabaNerve: ? instead of unwrap

coins/monero/tests/runner.rs
```diff
@@ -31,44 +29,27 @@ pub fn random_address() -> (Scalar, ViewPair, MoneroAddress) {
...
-    view: &random_scalar(&mut OsRng) * &ED25519_BASEPOINT_TABLE,
-  }
-  .to_string();
-
-  // Mine 40 blocks to ensure decoy availability
-  for _ in 0 .. 4 {
-    mine_blocks(&rpc, &addr).await;
+// TODO: Support transactions already on-chain
+// TODO: Don't have a side effect of mining blocks more blocks than needed under race conditions
+// TODO: mine as much as needed instead of default 10 blocks
```
kayabaNerve, line 34: These two are the same TODO

coins/monero/tests/runner.rs
```diff
@@ -162,17 +143,16 @@ macro_rules! test {
...
             let mut scanner =
               Scanner::from_view(view.clone(), Network::Mainnet, Some(HashSet::new()));
             let addr = scanner.address();
 
-            let start = rpc.get_height().await.unwrap();
-            for _ in 0 .. 7 {
-              mine_blocks(&rpc, &addr.to_string()).await;
-            }
+            // mine 40(decoy availability) + 60(miner tx unlock) blocks
+            rpc.generate_blocks(&addr.to_string(), 100).await.unwrap();
```
kayabaNerve: This always mines 40 blocks for decoy ability instead of only doing at the start like the rpc function did. Do we have a reason to get rid of that helper?

coins/monero/tests/runner.rs
```diff
@@ -237,8 +217,13 @@ macro_rules! test {
 
             let signed = sign(tx).await;
             rpc.publish_transaction(&signed).await.unwrap();
-            mine_blocks(&rpc, &random_address().2.to_string()).await;
-            ($first_checks)(rpc.clone(), signed.hash(), view.clone(), state).await
+            mine_until_unlocked(&rpc, &random_address().2.to_string(), signed.hash()).await;
+            ($first_checks)(
+              rpc.clone(),
+              signed,
+              Scanner::from_view(view.clone(), Network::Mainnet, Some(HashSet::new())),
```
kayabaNerve: Should we have a `let scanner = Scanner::from_view()`  to prevent this expansion?

coins/monero/tests/runner.rs
```diff
@@ -237,8 +217,13 @@ macro_rules! test {
 
             let signed = sign(tx).await;
             rpc.publish_transaction(&signed).await.unwrap();
-            mine_blocks(&rpc, &random_address().2.to_string()).await;
-            ($first_checks)(rpc.clone(), signed.hash(), view.clone(), state).await
+            mine_until_unlocked(&rpc, &random_address().2.to_string(), signed.hash()).await;
+            ($first_checks)(
+              rpc.clone(),
+              signed,
```
kayabaNerve: Please grab this from the RPC instead of passing it from here. That way, we also test deserialization and get_transaction.

coins/monero/tests/runner.rs
```diff
@@ -253,11 +238,16 @@ macro_rules! test {
...
             #[allow(unused_assignments)]
             {
               carried_state =
-                Box::new(($checks)(rpc.clone(), signed.hash(), view.clone(), state).await);
+                Box::new(($checks)(
+                  rpc.clone(),
+                  signed,
+                  Scanner::from_view(view.clone(), Network::Mainnet, Some(HashSet::new())),
+                  state
+                ).await);
```
kayabaNerve: Prior two comments on the above block apply here as well.

coins/monero/tests/send.rs
```diff
@@ -12,9 +13,8 @@ test!(
       builder.add_payment(addr, 5);
       (builder.build().unwrap(), ())
     },
-    |rpc: Rpc, hash, view, _| async move {
-      let mut scanner = Scanner::from_view(view, Network::Mainnet, Some(HashSet::new()));
-      let tx = rpc.get_transaction(hash).await.unwrap();
+    |rpc: Rpc, signed: Transaction, mut scanner: Scanner, _| async move {
+      let tx = rpc.get_transaction(signed.hash()).await.unwrap();
```
kayabaNerve: Just have the argument be named `tx` and do this call in the runner.

kayabaNerve (commented):
Largely looks good :)

coins/monero/tests/runner.rs
kayabaNerve: Invalid comment on my end end.

PR #187, opened 2022-12-08
Initial validator sets pallet

kayabaNerve:
Only works for a static validator set yet does have multisig voting/tracking.


kayabaNerve:
This needs tests and a further issue for session management (comprehensive to here, staking, and tendermint) made before merging can be considered. I'm requesting review now so we can also discuss what else to merge this, primarily as a multisig tracker, for protonet.

docs/protocol/Constants.md
```diff
@@ -48,7 +37,6 @@ Coins exist over a network and have a distinct integer ID.
 | Coin     | Network  | ID |
 |----------|----------|----|
 | Bitcoin  | Bitcoin  | 0  |
-| Ethereum | Ethereum | 1  |
-| USDC     | Ethereum | 2  |
-| DAI      | Ethereum | 3  |
-| Monero   | Monero   | 4  |
+| Ether    | Ethereum | 1  |
+| DAI      | Ethereum | 2  |
```
TheArchitect108, line 41: It may be work taking the ID to a string so that we can distinguish different chains within the ID as well, or we could also sign a separate chain ID.

TheArchitect108: An even better thing would be to generate a random GUID and form consensus around this GUID per coin/token.

docs/protocol/Staking.md
```diff
@@ -0,0 +1,25 @@
+# Staking
+
+Serai's staking pallet offers a DPoS system. All stake which enters the system
+is delegated somewhere. Delegatees can then allocate their stake to different
```
TheArchitect108: delegates**

docs/protocol/Staking.md
```diff
@@ -0,0 +1,25 @@
...
+validator sets, justifying their inclusion and providing financial security.
+
+Delegators may re-delegate stake whenever, so long as that stake isn't actively
+allocated. Delegators may also unstake whenever, so long as the prior condition
```
TheArchitect108: we should define as "active" and "inactive" stakes?

docs/protocol/Staking.md
```diff
@@ -0,0 +1,25 @@
...
+is still met.
+
+### Stake (message)
+
+  - `delegate` (Address): Address to delegate the newly added stake to.
+  - `amount`   (Amount):  Amount to stake and delegate.
+
+### Re-delegate (message)
```
TheArchitect108: "Move"

docs/protocol/Staking.md
```diff
@@ -0,0 +1,25 @@
...
+
+  - `from`   (Address): Address to transfer delegated stake from.
+  - `to`     (Address): Address to transfer delegated stake to.
+  - `amount` (Amount):  Amount to re-delegate.
+
+### Unstake (message)
```
TheArchitect108, line 16: We also need bond/unbond if we are making a distinction between staking for PoS and Bond for the set.

docs/protocol/Validator Sets.md
```diff
@@ -2,29 +2,80 @@
...
+accessible via `ValidatorSetInstance`.
 
-### Multisig Management
-
-Every validator set is expected to form a t-of-n multisig, where n is the amount
-of key shares in the validator set and t is `n / 3 * 2 + 1`, per curve required
-by its coins. This multisig is secure to hold funds up to 67% of the validator
-set's bond value. If funds exceed that threshold, there's more value in the
-multisig than in the supermajority of bond that must be put forth to control it.
+At launch, there will solely be Validator Set 1, managing Bitcoin, Ether, DAI,
```
TheArchitect108: set0

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  #[cfg(feature = "std")]
+  impl<T: Config> Default for GenesisConfig<T> {
+    fn default() -> Self {
+      GenesisConfig { bond: Amount(0), coins: Coin(0), participants: vec![] }
+    }
+  }
+
+  // Max of 16 coins per validator set
+  // At launch, we'll have BTC, ETH, DAI, and XMR
```
TheArchitect108, line 40: Maybe needs discussion, I do not think we should include tokens in the set count. Once we reach a multi-set environment we can probably remove tokens from bonding reqs as they can be guarded by contracts managed by these multiple sets.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+  // In the future, these will be split into separate validator sets, so we're already not
+  // planning expansion beyond just a few coins per validator set
+  // The only case which really makes sense for multiple coins in a validator set is:
+  // 1) The coins are small, easy to run, and make no sense to be in their own set
+  // In this case, it's still hard to ask validators to run 16 different nodes
+  // 2) The coins are all on the same network yet there's no DEX on-chain
+  // In these cases, it'd be hard to find and justify 16 different coins from that single chain
+  // This could probably be just 8, yet 16 is a hedge for the unforseen
+  // If necessary, this can be increased with a fork
+  type MaxCoinsPerSet = ConstU32<16>;
```
TheArchitect108, line 50: Why 16 and why u32 if 16?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  // Support keys up to 96 bytes (BLS12-381 G2)
+  const MAX_KEY_LEN: u32 = 96;
+  type MaxKeyLen = ConstU32<MAX_KEY_LEN>;
+
+  #[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+  pub struct ValidatorSet<T: Config> {
+    bond: Amount,
+    coins: BoundedVec<Coin, MaxCoinsPerSet>,
```
TheArchitect108, line 59: Coin could contain tokens?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  type Key = BoundedVec<u8, MaxKeyLen>;
+
+  /// The key for a given validator set instance coin.
+  #[pallet::storage]
+  #[pallet::getter(fn key)]
+  pub type Keys<T: Config> =
+    StorageMap<_, Twox64Concat, (ValidatorSetInstance, Coin), Key, OptionQuery>;
+
+  /// If an account has voted for a specific key or not. Prevents them from voting multiple times.
```
TheArchitect108, line 84: is there redundancy for this in the FROST lib?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+  #[pallet::storage]
+  #[pallet::getter(fn voted)]
+  pub type Voted<T: Config> = StorageMap<_, Blake2_128Concat, (T::AccountId, Key), (), OptionQuery>;
+
+  /// How many times a key has been voted for. Once consensus is reached, the keys will be adopted.
+  #[pallet::storage]
+  #[pallet::getter(fn vote_count)]
+  pub type VoteCount<T: Config> =
+    StorageMap<_, Blake2_128Concat, (ValidatorSetInstance, Coin, Key), u16, ValueQuery>;
```
TheArchitect108, line 93: With this being u16, if someone overflowed it, what would happen? Maybe voted prevents such, but just being sure.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  #[pallet::genesis_build]
+  impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
+    fn build(&self) {
+      // Start at 1 to leave 0 available for a future design where Serai itself has a distinct
```
TheArchitect108: Not sure if we should consider DPoS as a set. I think of it more logically as the foundation, and sets are built on top.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+      for participant in self.participants.clone() {
+        participants.push((participant, self.bond));
+      }
+
+      ValidatorSets::<T>::set(
+        ValidatorSetInstance(Session(0), ValidatorSetIndex(1)),
+        Some(ValidatorSet {
+          bond: self.bond,
+          coins: BoundedVec::try_from(coins).unwrap(),
+          participants: BoundedVec::try_from(participants).unwrap(),
```
TheArchitect108, line 113: This adds immutability to participants?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+    /// Validator Set already generated keys.
+    AlreadyGeneratedKeys,
+    /// Vvalidator has already voted for these keys.
+    AlreadyVoted,
+  }
+
+  #[pallet::call]
+  impl<T: Config> Pallet<T> {
+    #[pallet::weight(0)] // TODO
+    pub fn vote(
```
TheArchitect108, line 153: would it be logical to separate keygen vote and churn vote? I understand we are targeting single set, single generation for now, but this function seems incompatible with churning.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+      key: Key,
+    ) -> DispatchResult {
+      let signer = ensure_signed(origin)?;
+      // TODO: Do we need to check key is within bounds?
+
+      // TODO: Get session
+      let session: Session = Session(0);
+
+      // Confirm a key hasn't been set for this set instance
+      let instance = ValidatorSetInstance(session, index);
```
TheArchitect108, line 168: todo && not SetRotation?

substrate/validator-sets/primitives/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+
+/// The type used to identify a specific session of validators.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct Session(pub u32);
+
+/// The type used to identify a validator set.
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct ValidatorSetIndex(pub u16);
```
TheArchitect108, line 16: index vs guid?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -0,0 +1,206 @@
...
+
+      Self::deposit_event(Event::Vote { voter: signer, instance, coin, key: key.clone(), votes });
+
+      // If we've reached consensus, set the key
+      if usize::try_from(votes).unwrap() == set.participants.len() {
+        Keys::<T>::set((instance, coin), Some(key.clone()));
+        Self::deposit_event(Event::KeyGen { instance, coin, key });
+      }
+
+      Ok(())
```
TheArchitect108, line 200: would like to sanity check things in pallets before Ok -- understand this pallet will undergo major refactor so not necessary for this PR.

docs/protocol/Constants.md
silverpill: Perhaps CAIP-19 specification can be used: https://github.com/ChainAgnostic/CAIPs/blob/master/CAIPs/caip-19.md. It describes how to identify assets in multi-chain applications.

kayabaNerve: Absolute NACK around GUIDs. Alternatively to an integer, a string could be used. My objections are the inherent inefficiencies and the fact it's a protocol acknowledgement of what the coins are, rather than just another coin.

These inefficiencies aren't just minute. Bitcoin instructions are almost at the OP_RET size limit as-is. If integrators still are forced to a number scheme, I don't care to try to offer a nicety sporadically. I'd rather just have our lib define the constants as expected. And then for a native Substrate TX, I think this would bloat them 10%? As it adds ~12 bytes to each TX.

Alternatively, we can have a numeric ID for reference *AND* an on-chain string label, which is called upon request, not causing any inefficiencies. This on-chain label would be explicitly formatted so clients could get asset metadata without issue. CAIP-19 seems generally competent (URI scheme, chain ID, class, ID). My gripes are reuse of `:` (instead of a further `/`) and usage of SLIP-44 instead of a magic like `self` for the base.

So I do think the best move is an additional metadata section. My immediate question is should that be here, a new pallet entirely, and do we use CAIP-19 as-is, or a more LibP2P/Substrate esque `eip155:1/native`, `eip155:1/erc20:0x6b175474e89094c44da98b954eedeac495271d0f`...

docs/protocol/Staking.md
kayabaNerve: I didn't forget that word existed. Will fix.

docs/protocol/Staking.md
kayabaNerve: While active/inactive works, I prefer "allocate" to "activate" since you *allocate* X to Y and A to B. Saying you *activated* 5 SRI to set 1 and 10 SRI to set 2 is... weird.

docs/protocol/Staking.md
kayabaNerve: The delegate allocates stake for use as bond. That's just not currently implemented and not under the Staking pallet documented here.

docs/protocol/Staking.md
kayabaNerve: Transfer?

substrate/validator-sets/primitives/src/lib.rs
kayabaNerve: We absolutely are not doing GUIDs.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: If this is made a todo!, the static set can't vote on keys. Mind clarifying?

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: What "churn" vote?

vote isn't incompatible with multiple sets since its bound to a set instance. When a new instance confirms their key, it'll trigger a churn from the old multisig to the new one.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: This is indexed by instance. It's intended to be immutable. Nothing here forces it to be though.

docs/protocol/Validator Sets.md
kayabaNerve: Set 1, as documented in the Rust code. 0 is reserved for Serai itself, if we ever go that path.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: We have to include tokens here. It's a list of the coins managed and each token has a distinct token impl on Serai, classifying it as a coin. There's no way around it.

As for this "max", it is a hard limit, yet the reason we need a hard limit is for gas estimation purposes. If we set it to 32, then it'll be estimated as 32 elements long when it's likely 4. The max I'd see is 8 and I went with 16 just to be cautious.

Finally, no, we can't guard tokens by contracts AFAIK. Mind elaborating there? Even if we did a light client, it wouldn't void bond requirements.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: This isn't considering "DPoS" a set. It's considering the Serai Tendermint process, if we break Serai Tendermint away from the Serai Bitcoin validators, a set. It's an idea we've discussed before. If we do make Serai itself have a distinct validator set, rather than using the sum of all others, I don't want it to be set no. 9 of all things, when it's trivial to reserve 0 for what likely is a scalability requirement.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: This is about preventing double voting. If a validator could vote for this key twice, it'd bump its count twice, leading to 'consensus' with enough repeated votes.

FROST does not have state outside of its own and can't have redundancy for... whatever the case you're thinking of is.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: I either have to call them all coins or call them all tokens. I went with coins since most are.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: It has to be a ConstU32 as there's no const usize, and u32 takes the place of usize. It's 16 because I never see a single set managing more than 8 and 16 was a sane buffer.

Even if we expand with 4 more blockchains before multiple sets, we'd be at 8. If we ever list XHV (not an endorsement, they just don't have a on-chain DEX and would rely on a group such as ourselves), they technically have 11 yet only 3 or 4 worth a damn.

We can also trivially increase this in the future (with an upgrade, without migration logic). Accordingly, we can set it to 4 for now, and if we come back for more, raise it then. I just wanted to be a a *bit* more forward thinking.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Mind suggesting a few sanity checks?

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: It should be compiled with bounds checks and accordingly impossible to overflow. If it isn't, and the amount of votes required exceeded u16, it'd cycle and never reach consensus. If the amount of votes was less than u16, we had more participants than votes required/double votes and... beyond the far greater implications there, this overflow couldn't be reached as after hitting consensus it rejects future votes. If you could still bypass that, it'd write the keys multiple times/enable writing different keys.

docs/protocol/Constants.md
silverpill: Note that Monero currently doesn't have a dedicated chain ID in CAIP system.
I've recently created an issue for that: https://github.com/ChainAgnostic/namespaces/issues/41 (feedback is appreciated!).

docs/protocol/Staking.md
TheArchitect108: Prefer "staked" and "bonded" myself. Seems simple to teach others the mechanism with those words.

docs/protocol/Staking.md
TheArchitect108: make sense to me 👍

docs/protocol/Staking.md
TheArchitect108: Sure, prototype thing 👍

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: Understood, didn't know if we needed a distinction between the genesis keygen of a set versus churning keygens.

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: Didn't mean a macro, meant making this via comment with context for next steps here. I thought we would need to check if we were generating or rotating keys here but understand now each session will create new instances for each set.

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: When set members outgrow dPoS counts we will be in a situation where we can form super-set consensus to manage token contract ownership versus a single set. I think this security can result in reduced or eliminated assets for tokens that can be guarded. As you mentioned, the other sets would need observation on the relevant chains to be a party to this contract but in theory, this could work with light client driven watchers in other sets as you already have one BFT set forming consensus around full nodes.

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: I am saying we should clearly distinguish between coins and tokens. Coins are blockchain driven in this definition, requiring validators to add a chain to a set. Tokens are contract driven from an already supported chain.

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: I'm not sure why Serai would need to be considered a set. We will always need one logical group that manages consensus around SRI, we can see this distinct and separate from the general sets that will form. Better to separate the concept in my opinion than try to reserve 0 in a combined concept.

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: Seems low-hanging and a good pattern to enforce immutability past this point yes? Once the set is compiled here it will not need to change unless a new session forms, yes?

docs/protocol/Validator Sets.md
TheArchitect108: Resolving this for convo below about sets vs consensus group.

docs/protocol/Constants.md
TheArchitect108: I prefer to use string to integers from a security perspective. As far as CAIP goes, the effort is indeed appreciated but I generally feel we're a bit early in the cross-chain pursuit to define fully backwards compatible standards that will account for the needs of various environments. I believe at one point we discussed defining our own steganography towards efficiency of data included on other chains. I'm a bit of an optimization maximalist and hope we can even be compatible if chains more restrictive of metadata arise.

Maybe we should keep this a simplistic structure for protonet, keep CAIP in mind for testnet+, and focus this discussion on implementing a distinction between tokens and native assets?

substrate/validator-sets/pallet/src/lib.rs
TheArchitect108: no worries for this PR, let's resolve the functional aspects and have a bigger discussion on this for testnet design.

docs/protocol/Staking.md
kayabaNerve: I'll look at replacing allocate with bond :+1:

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: So currently, the idea is the chain itself is validated by a super-set. In the future discussions, the chain may be validated by a dedicated set.

In order for tokens to not be under a multisig, we'd need a light client. Light clients would effectively be a multisig of whoever produces consensus commits for the Serai blockchain. That maintains a bond requirement on either the super-set or the dedicated set. Else, the dedicated set can forge consensus messages and...

Right now, they can drain coins externally and SRI in pools, yet we dismiss that SRI having further value (since the coins in the pools were drained) AND their bond outweighs the externally stolen coins. Under a light client scheme, we lose bond despite having the same risk architecture.

I'd also note light clients as inefficient, which is why I don't prefer them, and largely incompatible with our current scheme. We absolutely need every coin, including tokens, here under our current architecture since consensus/in-instructions will look to this. While yes, in the future, we can list by light client, that'd be outside of a validator set (as you posit) and accordingly removed from here. I don't care to write special code here for something which won't end up here anyways.

Sorry if I'm missing something. LMK if the numbers are still questionable/there is a practical change you'd still prefer here.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Eh. Will Serai care? It may, depending on exactly how we handle gas, or it may not. Bifurcating, if they have the same on-network effects (a Serai token, independent in instructions, a need to reach consensus by that set), doesn't seem valuable.

I do know the processor/coordinator will care. That's just a detail there.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Right now, a super-set of all validators is used for Serai consensus. We've briefly discussed, due to limitations in Tendermint's scalability, solely using a dedicated set for Serai. We could use a pallet other than validator-sets? Except it'd still require the same bond mechanisms here. That's why I assumed we'd just re-use this pallet.

I'm not totally against using a dedicated pallet. That'd enable a few niceties...

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: We can't force it to be immutable. All we can do is never write to it again. We can, however, introduce niceties for that, such as a struct we'd have to explicitly bypass. I'd be willing to add that in BUT we may not want such explicitly intended immutability. I'm unsure how removal due to slashes works under Parity's session pallet/how we'll end up handling it.

PR #189, opened 2022-12-08
support add/read multiple arbitrary tx  data

akildemir:
- support to add multiple data fields into tx
- support to read multiple data fields from tx
- add tests

### DIFF
--- a/coins/monero/src/wallet/scan.rs
+++ b/coins/monero/src/wallet/scan.rs
@@ -83,7 +83,7 @@ impl Metadata {
     res.extend(self.subaddress.1.to_le_bytes());
     res.extend(self.payment_id);
 
-    res.extend([u8::try_from(self.arbitrary_data.len()).unwrap()]);
+    res.extend(u32::try_from(self.arbitrary_data.len()).unwrap().to_le_bytes());
     for part in &self.arbitrary_data {
       res.extend([u8::try_from(part.len()).unwrap()]);
       res.extend(part);
@@ -97,7 +97,7 @@ impl Metadata {
       payment_id: read_bytes(r)?,
       arbitrary_data: {
         let mut data = vec![];
-        for _ in 0 .. read_byte(r)? {
+        for _ in 0 .. read_u32(r)? {
           let len = read_byte(r)?;
           data.push(read_raw_vec(read_byte, usize::from(len), r)?);
         }
@@ -128,8 +128,8 @@ impl ReceivedOutput {
     self.data.commitment.clone()
   }
 
-  pub fn arbitrary_data(&self) -> Vec<Vec<u8>> {
-    self.metadata.arbitrary_data.clone()
+  pub fn arbitrary_data(&self) -> &[Vec<u8>] {
+    &self.metadata.arbitrary_data
   }
 
   pub fn serialize(&self) -> Vec<u8> {
--- a/coins/monero/src/wallet/scan.rs
+++ b/coins/monero/src/wallet/scan.rs
@@ -82,8 +82,10 @@ impl Metadata {
     res.extend(self.subaddress.0.to_le_bytes());
     res.extend(self.subaddress.1.to_le_bytes());
     res.extend(self.payment_id);
+
+    res.extend([u8::try_from(self.arbitrary_data.len()).unwrap()]);
     for part in &self.arbitrary_data {
-      res.extend([1, u8::try_from(part.len()).unwrap()]);
+      res.extend([u8::try_from(part.len()).unwrap()]);
       res.extend(part);
     }
     res
@@ -95,8 +97,7 @@ impl Metadata {
       payment_id: read_bytes(r)?,
       arbitrary_data: {
         let mut data = vec![];
-        let exist = read_byte(r)?;
-        while exist == 1 {
+        for _ in 0 .. read_byte(r)? {
           let len = read_byte(r)?;
           data.push(read_raw_vec(read_byte, usize::from(len), r)?);
         }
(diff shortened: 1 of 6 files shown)


akildemir:
- I expect comments about consume/borrow
- wasn't sure about modifying processor/monero `prepare_send()`  to take the arbitrary data as a parameter in this pr.

coins/monero/src/wallet/scan.rs
```diff
@@ -82,11 +82,11 @@ impl Metadata {
     res.extend(self.subaddress.0.to_le_bytes());
     res.extend(self.subaddress.1.to_le_bytes());
     res.extend(self.payment_id);
-    if let Some(data) = self.arbitrary_data.as_ref() {
-      res.extend([1, u8::try_from(data.len()).unwrap()]);
-      res.extend(data);
-    } else {
-      res.extend([0]);
+
+    res.extend([u8::try_from(self.arbitrary_data.len()).unwrap()]);
```
kayabaNerve: This check isn't correct. We can have >255 segments of data. It's just that each segment must be less than 256 in length. Please use a u32. u16 has some theoretical issues I'd rather not leave open.

coins/monero/src/wallet/scan.rs
```diff
@@ -96,12 +96,12 @@ impl Metadata {
       subaddress: (read_u32(r)?, read_u32(r)?),
       payment_id: read_bytes(r)?,
       arbitrary_data: {
-        if read_byte(r)? == 1 {
+        let mut data = vec![];
+        for _ in 0 .. read_byte(r)? {
```
kayabaNerve: This will also have to be edited to be a u32.

coins/monero/src/wallet/scan.rs
```diff
@@ -128,6 +128,10 @@ impl ReceivedOutput {
     self.data.commitment.clone()
   }
 
+  pub fn arbitrary_data(&self) -> Vec<Vec<u8>> {
```
kayabaNerve: I'd return &[Vec<u8>] here. That lets the user iterate over it without a clone, or if they need an owned instance, clone it themselves.

coins/monero/src/wallet/send/builder.rs
```diff
@@ -15,14 +15,14 @@ struct SignableTransactionBuilderInternal {
...
-  data: Option<Vec<u8>>,
+  data: Vec<Vec<u8>>,
 }
 
 impl SignableTransactionBuilderInternal {
   // Takes in the change address so users don't miss that they have to manually set one
   // If they don't, all leftover funds will become part of the fee
   fn new(protocol: Protocol, fee: Fee, change_address: Option<MoneroAddress>) -> Self {
-    Self { protocol, fee, inputs: vec![], payments: vec![], change_address, data: None }
+    Self { protocol, fee, inputs: vec![], payments: vec![], change_address, data: Vec::new() }
```
kayabaNerve: vec![], not that it makes a difference ;)

coins/monero/src/wallet/send/builder.rs
```diff
@@ -100,9 +100,12 @@ impl SignableTransactionBuilder {
     self.shallow_copy()
   }
 
-  pub fn set_data(&mut self, data: Vec<u8>) -> Self {
-    self.0.write().unwrap().set_data(data);
-    self.shallow_copy()
+  pub fn add_data(&mut self, data: &Vec<u8>) -> Result<Self, TransactionError> {
+    if data.len() > 255 {
```
kayabaNerve: We should probably make a constant out of this at this point...

coins/monero/tests/add_data.rs
```diff
@@ -0,0 +1,80 @@
...
+    |_, mut builder: Builder, addr| async move {
+      // make a data that is less than 255 bytes
+      let arbitrary_data = Vec::from("this is an arbitrary data less than 255 bytes");
+
+      // make sure we can add to tx
+      let result = builder.add_data(&arbitrary_data);
+      assert!(result.is_ok());
+
+      builder.add_payment(addr, 5);
+      (builder.build().unwrap(), (arbitrary_data,))
```
kayabaNerve, line 17: The state doesn't need to be a tuple.

coins/monero/tests/add_data.rs
```diff
@@ -0,0 +1,80 @@
...
+      // make a data that is bigger than 255 bytes
+      let mut arbitrary_data = vec![];
+      for _ in 0 .. 256 {
+        arbitrary_data.push(b'a');
+      }
+
+      // make sure we get an error if we try to add it to tx
+      let mut result = builder.add_payment(addr, 5).add_data(&arbitrary_data);
+      assert!(result.is_err());
+      matches!(result.err(), Some(TransactionError::TooMuchData));
```
kayabaNerve: You can combine these two lines with an assert_eq!

kayabaNerve (changes requested):

coins/monero/tests/add_data.rs
akildemir: yes. Just a preference. So that passing structure is the same if we need to pass multiple variables!

coins/monero/tests/add_data.rs
akildemir: if you mean like ` assert_eq!(result.err(), Some(TransactionError::TooMuchData))` looks like assert_eq doesn't work on Option type.

kayabaNerve: `assert_eq!(result, Err(TransactionError::TooMuchData));`

PR #193, opened 2022-12-15
Monero lib imrovements

akildemir:
- add new `integrated_address` and  `featured_address` functions to scanner
- add ability to encode address directly from `ViewPair`
- make some comments more elaborate
- add new tests for scanning outputs


coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+    let mut scanner = Scanner::from_view(
+      ViewPair::new(spend_pub, Zeroizing::new(Scalar::from_bits(ADDR_VIEW))),
+      Network::Mainnet,
+      None,
+    );
+
+    // generate all types of addresses
+    let payment_ids =
+      vec![[88, 37, 149, 111, 171, 108, 120, 181], [125, 69, 155, 152, 140, 160, 157, 186]];
+    let mut addresses = vec![];
```
kayabaNerve: You can write this entire section using `vec![]`.
```
vec![
  // comment
  AddressInfo {},
  ...
]
```
It'd remove the pushes and offer consistent formatting.

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+    let spend_pub = &Scalar::from_bits(ADDR_SPEND) * &ED25519_BASEPOINT_TABLE;
+    let mut scanner = Scanner::from_view(
+      ViewPair::new(spend_pub, Zeroizing::new(Scalar::from_bits(ADDR_VIEW))),
+      Network::Mainnet,
+      None,
+    );
+
+    // generate all types of addresses
+    let payment_ids =
+      vec![[88, 37, 149, 111, 171, 108, 120, 181], [125, 69, 155, 152, 140, 160, 157, 186]];
```
kayabaNerve: What are these payment IDs? If they're random, we should randomly generate them using `fill_bytes`.

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+
+    // submit and unlock tx
+    rpc.publish_transaction(&tx).await.unwrap();
+    runner::mine_until_unlocked(&rpc, &miner_addr.to_string(), tx.hash()).await;
+
+    // get the tx and confirm receipt
+    tx = rpc.get_transaction(tx.hash()).await.unwrap();
+    let output = addr_scanner.scan_transaction(&tx).not_locked().swap_remove(0);
+    assert_eq!(output.commitment().amount, amount);
+    if addr.payment_id.is_some() {
```
kayabaNerve: This doesn't confirm if the subaddress also lines up.

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+use monero_serai::{
+  wallet::{
+    address::{Network, MoneroAddress},
+    ViewPair, SignableTransaction, Scanner,
+  },
+};
+
+mod runner;
+
+const ADDR_SPEND: [u8; 32] =
```
kayabaNerve: Why not randomly generate these?

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+      .await
+      .unwrap()
+      .swap_remove(0)
+      .ignore_timelock()
+      .swap_remove(0);
+  }
+}
+
+async_sequential!(
+  async fn scan_all_standard_addresses() {
```
kayabaNerve: standard -> non_guaranteed?

kayabaNerve: Should we use terms other then standard/non_guaranteed and just have more of these tests?

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,160 @@
...
+  hex!("bf02a79d9e30ea76872e565722517924151793c535832e4353e54dc0be698001");
+const ADDR_VIEW: [u8; 32] =
+  hex!("c300b798c000f3c5da4ebc5a288418fe9060d623c9830b7abc3d9e7a3152eb08");
+
+pub struct AddressInfo {
+  address: String,
+  payment_id: Option<[u8; 8]>,
+}
+
+pub async fn scan_incomings(addr_scanner: &mut Scanner, addresses: &Vec<AddressInfo>) {
```
kayabaNerve: While I'm not against this custom runner, shouldn't we define this as a macro around the existing macro, and just run it on one address type at a time?

coins/monero/tests/runner.rs
```diff
@@ -1,5 +1,6 @@
 use core::ops::Deref;
 use std::sync::Mutex;
+use std::collections::HashSet;
```
kayabaNerve, line 2: Please use a single import here. `use std::{sync::Mutex, collections::HashSet};`

coins/monero/src/wallet/mod.rs
```diff
@@ -218,4 +242,32 @@ impl Scanner {
       self.pair.view.deref() * spend,
     )
   }
+
+  /// Returns a featured address.
+  ///
+  /// if you created the scanner with `Some` value for `burning_bug` be aware that
```
kayabaNerve: "A Scanner will fail to scan this address unless it is appropriate to the Scanner. If the Scanner was defined as not handling the burning bug, by passing `None` for `burning_bug` upon construction, this must be a guaranteed address in order to be scanned by it."

coins/monero/src/wallet/mod.rs
```diff
@@ -218,4 +242,32 @@ impl Scanner {
...
+  /// even though you can pass `true` for the `guaranteed` parameter and generate
+  /// a guaranteed address, scanner wont be able decode incoming outputs for this address.
+  /// If you want to support guaranteed addresses please create a scanner with `None` value to
+  /// `burning_bug`. Reverse of the case also holds.
+  pub fn featured_address(
+    &mut self,
+    index: Option<(u32, u32)>,
```
kayabaNerve: index -> subaddress? It's a clearer argument name, even though it would conflict with a variable in this function (which would then have to be renamed).

coins/monero/src/wallet/mod.rs
```diff
@@ -180,23 +194,33 @@ impl Scanner {
...
+        AddressType::Featured(false, None, true)
+      } else {
+        AddressType::Standard
+      },
+    );
+    self.pair.address(meta)
   }
 
-  /// Return the specified subaddress for this view pair.
+  /// Returns the integrated address for a given payment id.
```
kayabaNerve: "ID"

coins/monero/src/wallet/mod.rs
```diff
@@ -167,8 +176,13 @@ impl Scanner {
   /// burning_bug is a HashSet of used keys, intended to prevent key reuse which would burn funds.
   /// When an output is successfully scanned, the output key MUST be saved to disk.
   /// When a new scanner is created, ALL saved output keys must be passed in to be secure.
-  /// If None is passed, a modified shared key derivation is used which is immune to the burning
-  /// bug (specifically the Guaranteed feature from Featured Addresses).
+  /// If None is passed, a modified shared key derivation is used to generate new addresses
+  /// through this scanner which is immune to the burning bug (specifically
+  /// the Guaranteed feature from Featured Addresses).
```
kayabaNerve: The original section had clearer English. If you want to highlight the address implications, how about adding
"This will effect generated/supported addresses."?

coins/monero/src/wallet/mod.rs
```diff
@@ -167,8 +176,13 @@ impl Scanner {
...
+  ///
+  /// If you want to use guaranteed feature for both generating addresses and
```
kayabaNerve: I don't feel a note is needed here, but if we are to have one:

"In order to generate traditional Monero addresses, you must manage a cache of used keys, and pass in that cache here."

coins/monero/src/wallet/mod.rs
```diff
@@ -121,12 +123,19 @@ impl ViewPair {
...
+  pub fn address(&self, meta: AddressMeta<MoneroAddressBytes>) -> MoneroAddress {
+    MoneroAddress::new(meta, self.spend, self.view.deref() * &ED25519_BASEPOINT_TABLE)
+  }
 }
 
 /// Transaction scanner.
-/// This scanner is capable of generating subaddresses, additionally scanning for them once they've
-/// been explicitly generated. If the burning bug is attempted, any secondary outputs will be
-/// ignored.
+/// - This scanner is capable of generating subaddresses, additionally scanning for
```
kayabaNerve: These shouldn't be bullet points. The existing new lines should be fine.

coins/monero/src/wallet/mod.rs
```diff
@@ -121,12 +123,19 @@ impl ViewPair {
...
+/// them once they've been explicitly generated by using `subaddress()` member function.
```
kayabaNerve: "explicitly generated via `Scanner::subbaddress`."

coins/monero/src/wallet/mod.rs
```diff
@@ -121,12 +123,19 @@ impl ViewPair {
...
+/// It wont be able to track the outputs for subaddresses that aren't generated this way.
```
kayabaNerve: This is redundant. If you feel it still helps for clarity:
"The Scanner won't be able to track outputs for subaddresses which haven't been generated through it."

coins/monero/src/wallet/mod.rs
```diff
@@ -121,12 +123,19 @@ impl ViewPair {
...
+/// - If the burning bug is attempted, any secondary outputs will be ignored.
+/// or standard address type if none passed.
```
kayabaNerve: This line isn't valid. The prior statement was if the burning bug is attempted, the secondary outputs will be ignored. That requires using a standard, burning bug vulnerable address. If None is passed for the burning_bug, immune addresses will be used and the burning bug can't be attempted. If Some is passed, it can be attempted, yet the secondaries will be ignored. That's just not the "or" conditions.

coins/monero/src/wallet/mod.rs
```diff
@@ -121,12 +123,19 @@ impl ViewPair {
       .concat(),
     ))
   }
+
+  /// returns the MoneroAddress type according to specifications in the meta
```
kayabaNerve: "Returns an address with the provided metadata."

It removes the term "specifications" and is more direct accordingly.

coins/monero/src/wallet/address.rs
```diff
@@ -182,6 +182,14 @@ impl<B: AddressBytes> ToString for Address<B> {
...
 
 impl<B: AddressBytes> Address<B> {
+  /// Generates an Address type according to the specification provided in the meta.
+  ///
+  /// WARNING: Specification on the meta are not "commands" and are
+  /// just "specifications" for the address. This function relies on the
+  /// caller to make sure all the passed parameters makes sense and assumes
+  /// specifications provided in the meta are correct and applicable for the passed in keys.
+  /// For example, passing `AddressType::Subaddress` type in the meta wont generate
+  /// a correct subaddress if the keys aren't already for a valid subaddress.
```
kayabaNerve: "Creates an Address with the provided properties."

I don't feel a need for the warning. This isn't intended to be called from private keys, a view pair, or a scanner. It's literally creating an address object without processing. I feel there's a natural understanding it won't do anything further, especially when it cannot (subaddresses require private view key knowledge).

Users are also sheparded to the Scanner for address creation.

kayabaNerve (changes requested):

coins/monero/src/wallet/mod.rs
```diff
@@ -121,6 +123,11 @@ impl ViewPair {
       .concat(),
     ))
   }
+
+  /// Returns an address with the provided metadata.
+  pub fn address(&self, meta: AddressMeta<MoneroAddressBytes>) -> MoneroAddress {
+    MoneroAddress::new(meta, self.spend, self.view.deref() * &ED25519_BASEPOINT_TABLE)
```
kayabaNerve: This isn't correct when the metadata specifies a subaddress.

Since AddressMeta doesn't include the subaddress index, there's no way to have a safe function with this API. It'd have to take payment ID/subaddress/guaranteed as individual arguments and then build a derivation/meta from that.

coins/monero/src/wallet/mod.rs
```diff
@@ -218,4 +235,32 @@ impl Scanner {
...
+    payment_id: Option<[u8; 8]>,
+    guaranteed: bool,
+  ) -> MoneroAddress {
+    let is_subaddress = subaddress.is_some() && (subaddress.unwrap() != (0, 0));
+
+    let mut spend = self.pair.spend;
+    let mut view: EdwardsPoint = self.pair.view.deref() * &ED25519_BASEPOINT_TABLE;
+    if is_subaddress {
+      spend =
+        self.pair.spend + (&self.pair.subaddress(subaddress.unwrap()) * &ED25519_BASEPOINT_TABLE);
```
kayabaNerve: Let's not have two instances of this cryptography/(0, 0) handling. We should have a `fn subaddress_derivation` function, handling calculation + registration, that both this and subaddress call.

coins/monero/tests/scan.rs
```diff
@@ -0,0 +1,213 @@
+use rand::RngCore;
+
+use monero_serai::{transaction::Transaction};
```
kayabaNerve: Unneeded {}.

kayabaNerve (changes requested):

kayabaNerve:
CLI squashed in https://github.com/serai-dex/serai/commit/a646ec5aaaf816887874c8311bf8bd9c02d2a33f :+1:

PR #196, opened 2022-12-30
DKG Blame

kayabaNerve:
Implements FROST DKG blame and a new encryption scheme, which while overkill for this, should be without side effects and usable for a variety of protocols.


kayabaNerve:
Still needs testing.

PR #211, opened 2023-01-09
Move to dtolnay/toolchain

kayabaNerve:
actions-rs/toolchain is unmaintained (https://github.com/actions-rs/toolchain/issues/216).

actions-rs will stop working in a few months thanks to the deprecation of `set-output` and NodeJS 12 (https://github.com/actions-rs/toolchain/issues/219 and https://github.com/actions-rs/toolchain/issues/221, respectively).

This moves to https://github.com/dtolnay/rust-toolchain, which is maintained, and deletes cached-rust per commentary in https://github.com/dtolnay/rust-toolchain/issues/26. We may later want to add https://github.com/Swatinem/rust-cache/ in.


kayabaNerve:
Tests take notably longer. Will merge this and create issue for cache.

pinkforest:
pstt fyi I forked https://github.com/rustsec/audit-check - PR's welcome

There was a discussion that we might host some simple version that doesn't require NodeJS

kayabaNerve:
Sorry, yet I'm not entirely sure where/what you forked.

As for a version which doesn't require NodeJS, that sounds great just as we're discussing minimizing bloat. I don't think it's worthwhile enough for us to move though. If it also handled toolchain caching intelligently, that'd be pleasant though (since we use stable, nightly, additional targets, various components, update nightly monthly...). Would love to not hammer Rust's servers while also only downloading relevant pieces from the cache (while simultaneously not getting stuck on an old version of nightly from the cache).

Feel free to clarify, as I wouldn't necessarily mind poking around :) I'd also love if an org handled it, not an individual...

pinkforest:
We forked https://github.com/actions-rs/audit-check/ into https://github.com/rustsec/audit-check

I guess we need to address this/these deprecation/s there now too in our fork

kayabaNerve:
Ah, got it, thanks. We actually don't currently use audit-check. We use `deny`, which also handles advisories, yet also yankings/licenses, so that wasn't relevant to us.

Forking the existing action, and merging the open PRs on it, would be seamless to everyone else. RustSec also should be a trusted organization. I'd love to see y'all step up to handle it.

Issue #223, opened 2023-01-12
monero-serai has been fully yanked


kayabaNerve:
@akildemir, while testing the library, identified we did not properly calculate the payment ID encryption mask. I have yanked all such problematic crates (all of them) accordingly, as sending to a payment ID address will lead to a loss of funds for the sender (as it'll not be credited to the intended address, despite the recipient still receiving the funds in question).

monero-serai is eligible for a new alpha release one a fix is published. Unfortunately, publishing a new release would also require publishing the latest versions for all of our crypto crates, which I'd rather not do until after the audit (see #199).

Since I don't believe we have any active users, it should be fine to wait? If anyone is blocked by this and needs a working payment ID structure, and can't simply use a git revision, please let me know. Until then, this serves as documentation for the yanking.

kayabaNerve:
The bug in question is fixed on develop. Using a rev of 3b920ad471df150e2d871a3963f4677b3d1d05ed should be safe for continued usage of the alpha release.

kayabaNerve:
We additionally did not have the additional keys in TX extra properly. A fix for that will be released soon. I'll note the revision here when it's available.

kayabaNerve:
monero-serai 0.1.3-alpha has now been published.

PR #224, opened 2023-01-13
Monerolib Improvements

akildemir:
- Fix Monero Wallet RPC payment id compatibilty
- add tests for  Monero wallet RPC compatibility


kayabaNerve:
Failing deny fixed in https://github.com/serai-dex/serai/commit/d37aff515dfdc869714d5091c587f53d271ce173.

This branch should fix the reason for #223.

coins/monero/src/wallet/mod.rs
```diff
@@ -106,6 +108,14 @@ impl ViewPair {
     ViewPair { spend, view }
   }
 
+  pub fn spend(&self) -> EdwardsPoint {
+    self.spend
+  }
+
+  pub fn view(&self) -> Zeroizing<Scalar> {
```
kayabaNerve: I'm unsure we should expose the private key again.  This should return its public key via `&EDWARDS_BASEPOINT_TABLE * self.view.deref()`

coins/monero/Cargo.toml
```diff
@@ -55,6 +55,7 @@ monero-generators = { path = "generators", version = "0.1" }
 
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
+monero-rpc = "0.3.2"
```
kayabaNerve: This locks monero-rpc to only 0.3.2. Using "0.3" would allow automatically updating to the most recent bug fix.

.github/actions/monero-wallet-rpc/action.yml
```diff
@@ -0,0 +1,44 @@
...
+    default: v0.18.1.2
+
+runs:
+  using: "composite"
+  steps:
+    - name: Monero Wallet RPC Cache
+      id: cache-monerowalletrpc
+      uses: actions/cache@v3
+      with:
+        path: monerowalletrpc
```
kayabaNerve: This is the path of the file to cache. It should be monero-wallet-rpc accordingly.

.github/actions/monero-wallet-rpc/action.yml
```diff
@@ -0,0 +1,44 @@
...
+  version:
+    description: "Version to download and run"
+    required: false
+    default: v0.18.1.2
+
+runs:
+  using: "composite"
+  steps:
+    - name: Monero Wallet RPC Cache
+      id: cache-monerowalletrpc
```
kayabaNerve: I'd rather keep the tabs, here and for the key, if you don't mind.

.github/actions/monero-wallet-rpc/action.yml
```diff
@@ -0,0 +1,44 @@
...
+
+        FILE=monero-$RUNNER_OS-$RUNNER_ARCH-${{ inputs.version }}.tar.bz2
+        wget https://downloads.getmonero.org/cli/$FILE
+        tar -xvf $FILE
+
+        mv monero-x86_64-linux-gnu-${{ inputs.version }}/monero-wallet-rpc  monero-wallet-rpc 
+
+    - name: Monero Wallet RPC
+      shell: bash
+      run: ./monero-wallet-rpc --disable-rpc-login --rpc-bind-port 6061 --allow-mismatched-daemon-version  --wallet-dir ./
```
kayabaNerve: Two spaces before `--wallet-dir`

.github/actions/monero-wallet-rpc/action.yml
```diff
@@ -0,0 +1,44 @@
...
+        RUNNER_ARCH=${RUNNER_ARCH,,}
+
+        RUNNER_OS=linux
+        RUNNER_ARCH=x64
+
+        FILE=monero-$RUNNER_OS-$RUNNER_ARCH-${{ inputs.version }}.tar.bz2
+        wget https://downloads.getmonero.org/cli/$FILE
+        tar -xvf $FILE
+
+        mv monero-x86_64-linux-gnu-${{ inputs.version }}/monero-wallet-rpc  monero-wallet-rpc
```
kayabaNerve: Two spaces before `monero-wallet-rpc`

.github/actions/monero/action.yml
```diff
@@ -5,7 +5,7 @@ inputs:
   version:
     description: "Version to download and run"
     required: false
-    default: v0.18.0.0
+    default: v0.18.1.2
```
kayabaNerve, line 8: Please also update the matrix in the monero-tests workflow.

.github/actions/test-dependencies/action.yml
```diff
@@ -29,3 +29,8 @@ runs:
       uses: ./.github/actions/monero
       with:
         version: ${{ inputs.monero-version }}
+
+    - name: Run a Monero Wallet-RPC Client
```
kayabaNerve: "Client" isn't correct. Just "Wallet" is fine.

.github/actions/test-dependencies/action.yml
```diff
@@ -29,3 +29,8 @@ runs:
...
+      uses: ./.github/actions/monero-wallet-rpc
+      with:
```
kayabaNerve: Remove this with. We can just always use monero-wallet-rpc 0.18.2. The reason we test older daemons is to test our successful handling of old TX types. That isn't relevant here.

coins/monero/tests/wallet-rpc-compatibility.rs
```diff
@@ -0,0 +1,88 @@
+use std::{
+  collections::{HashSet, HashMap},
+  str::FromStr,
+};
+
+use rand_core::OsRng;
+use rand::RngCore;
```
kayabaNerve: rand_core::{RngCore, OsRng};

coins/monero/tests/wallet-rpc-compatibility.rs
```diff
@@ -0,0 +1,88 @@
...
+};
+
+use monero_serai::{
+  wallet::address::{Network, AddressSpec, SubaddressIndex},
+  wallet::Scanner,
+};
+
+mod runner;
+
+async fn test_from_wallet_rpc_to(spec: AddressSpec) {
```
kayabaNerve: Hanging `_to`. Either `test_from_wallet_rpc` or `test_from_wallet_rpc_to_self`.

coins/monero/tests/wallet-rpc-compatibility.rs
```diff
@@ -0,0 +1,88 @@
...
+  let wallet_rpc =
+    monero_rpc::RpcClientBuilder::new().build("http://127.0.0.1:6061").unwrap().wallet();
+  let daemon_rpc = runner::rpc().await;
+
+  // initialize wallet rpc
+  let address_resp = wallet_rpc.get_address(0, None).await;
+  let wallet_rpc_addr = if address_resp.is_ok() {
+    address_resp.unwrap().address
+  } else {
+    wallet_rpc.create_wallet("test_wallet".to_string(), None, "English".to_string()).await.unwrap();
```
kayabaNerve, line 30: Should we always create a fresh wallet? Either is fine in this case. Just a commentary on providing fresh contexts/reusable code.

coins/monero/tests/wallet-rpc-compatibility.rs
```diff
@@ -0,0 +1,88 @@
...
+
+  // create the scanner
+  let mut scanner = Scanner::from_view(view_pair, Some(HashSet::new()));
+  if let AddressSpec::Subaddress(index) = spec {
+    scanner.register_subaddress(index);
+  }
+
+  // retrieve it and confirm
+  let tx = daemon_rpc.get_transaction(tx_hash).await.unwrap();
+  let output = scanner.scan_transaction(&tx).ignore_timelock().swap_remove(0);
```
kayabaNerve: not_locked, not ignore_timelock

coins/monero/tests/wallet-rpc-compatibility.rs
```diff
@@ -0,0 +1,88 @@
...
+
+  match spec {
+    AddressSpec::Subaddress(index) => assert_eq!(output.metadata.subaddress, Some(index)),
+    AddressSpec::Integrated(payment_id) => assert_eq!(output.metadata.payment_id, payment_id),
```
kayabaNerve: Also check the subaddress is None.

kayabaNerve (changes requested):
Overall looks good :) Please rebase onto latest develop (fixing deny) and correct the minor comments.

coins/monero/src/wallet/mod.rs
akildemir: The whole reason triggered putting these 2 getter functions was that private view key needed to be accesses in some places(when interacting with wallet-rpc) although not at the moment.  I would think those functions are there to give us access to view pair private parts  when needed. Caller can always derive the public key itself when/if needed but other way around isn't possible with only exposing the public key.

coins/monero/tests/wallet-rpc-compatibility.rs
akildemir: Thought about it but always creating a fresh wallet per test will need extra boilerplate code to manage the wallet-rpc. Things like switch wallet, refresh etc.. Also will need to mine 70 blocks per tx instead of 10. Either is fine as you said thought going with single one is simpler/faster.

PR #231, opened 2023-01-16
deleted compromised key from dockerfile

vrx00:
Key, which belongs to "Luke Dashjr <luke@dashjr.org>", can not be validated by opengpg. It occurs "no user ID" error  for the key. It was not revoked on ubuntu and can be validated by "hkp://keyserver.ubuntu.com:80" server. Therefore, I removed the key.


TheArchitect108:
Can you go ahead a bump https://github.com/RustCrypto/signatures/issues/615 with this so that the deny remedies?

kayabaNerve:
@TheArchitect108 I just pushed a `cargo update`, along with the new Substrate update.

kayabaNerve:
As for the key, I brought this up with @TheArchitect108 when it happened. At the time, we were unsure the validity of the report, and had acknowledged BTC had yet to do anything.

In order for us to merge this, I would like to know the following:
- Is there a revocation certificate, or was it just removed from some server? If so, which?
- Can we still use this key if we also ack the Ubuntu server?
- Will not having this key imported break verifying BTC signatures, since unrecognized keys cause the entire sig file to error?

If the missing key doesn't break the sig verification, I can merge this. I just want the context documented. I'd also like a separate issue for discussing how we handle this overall to ensure its tracked.

kayabaNerve:
Sorry, just read the issue which ack'd the compromise. Happy that was logged. I don't see the exact error case, despite the included log, yet will merge this on the assumption BTC works after. Please confirm it works, despite the missing key, and comment if there's an official revocation certificate available.

vrx00:
I downloaded his key from https://keys.openpgp.org/search?q=E463A93F5F3117EEDE6C7316BD02942421F4889F and verified that it was revoked on 2023-01-05. The log is
```
$ gpg -v --import-options show-only --import --with-fingerprint luke-jr.asc
gpg: armor header: Comment: E463 A93F 5F31 17EE DE6C  7316 BD02 9424 21F4 889F
pub   rsa8192 2012-03-23 [SCEA] [revoked: 2023-01-05]
      E463 A93F 5F31 17EE DE6C  7316 BD02 9424 21F4 889F
sub   rsa8192 2012-03-23 [E] [revoked: 2023-01-05]
```

PR #233, opened 2023-01-17
Initial In Instructions pallet and Serai client lib

kayabaNerve:
Exceeds its initial scope of the In Instructions pallet, and the stated additional scope of a Serai client lib, by additionally doing cleanup and infra work.

Does not complete the II pallet, which has to call other pallets, yet this code is currently required elsewhere, hence this PR existing now.

Defines a horrific InInstructionsProvider, which polls a JSON RPC server, to be replaced by Kafka.

Uses the `serai-client` lib to define e2e node tests.

Modularizes the processor tests, due to the history of this branch not currently reflected by this PR, with the caveat the processor currently uses `src/tests` when it should just use `tests/`. That'll be corrected after this is merged to develop.


substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+      for (coin, update) in updates.iter().enumerate() {
+        if let Some(update) = update {
+          let coin = Coin::from(u32::try_from(coin).unwrap());
+          BlockNumbers::<T>::insert(coin, update.block_number);
+
+          for batch in &update.batches {
+            // TODO: EXECUTE
+            Self::deposit_event(Event::Batch { coin, id: batch.id });
+            NextBatch::<T>::insert(coin, batch.id + 1);
```
TheArchitect108: Maybe we should use unique IDs instead of incremental?

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+        };
+
+        if updates.len() != expected.len() {
+          Err(InherentError::InvalidUpdateQuantity(
+            updates.len().try_into().unwrap(),
+            expected.len().try_into().unwrap(),
+          ))?;
+        }
+
+        for (coin, both) in updates.iter().zip(&expected).enumerate() {
```
TheArchitect108: Let's comment each case here, important for us to ensure we have all cases covered.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+            (Some(update), Some(expected)) => {
+              if update.block_number.0 > expected.block_number.0 {
+                Err(InherentError::UnrecognizedBlockNumber(
+                  coin,
+                  update.block_number,
+                  expected.block_number,
+                ))?;
+              }
+
+              let prev = BlockNumbers::<T>::get(coin);
```
TheArchitect108: Why/how does this get the previous number? Wouldn't this get the current?

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+              if update.block_number.0 <= prev.0 {
+                Err(InherentError::InvalidBlockNumber(coin, update.block_number, prev))?;
+              }
+
+              if update.batches.len() > expected.batches.len() {
```
TheArchitect108: Based on the way these are populated, I'm failing to see how expected would ever be different than update. Mind elaborating?

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+
+  #[pallet::storage]
+  pub(crate) type Once<T: Config> = StorageValue<_, bool, ValueQuery>;
+  #[pallet::storage]
+  #[pallet::getter(fn block_number)]
+  pub(crate) type BlockNumbers<T: Config> =
+    StorageMap<_, Blake2_256, Coin, BlockNumber, ValueQuery>;
+  #[pallet::storage]
+  #[pallet::getter(fn executed_batches)]
+  pub(crate) type NextBatch<T: Config> = StorageMap<_, Blake2_256, Coin, u64, ValueQuery>;
```
TheArchitect108: Let's comment our storage maps please.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+              }
+
+              if update.batches.len() > expected.batches.len() {
+                Err(InherentError::UnrecognizedBatches(
+                  coin,
+                  (update.batches.len() - expected.batches.len()).try_into().unwrap(),
+                ))?;
+              }
+
+              let mut next_batch = NextBatch::<T>::get(coin);
```
TheArchitect108: Are we effectively processing two batches with this?

substrate/in-instructions/primitives/src/incoming.rs
```diff
@@ -0,0 +1,52 @@
...
+
+use sp_core::{ConstU32, bounded::BoundedVec};
+
+use serai_primitives::NativeAddress;
+
+use crate::{MAX_DATA_LEN, ExternalAddress};
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Application {
```
TheArchitect108, line 15: This seems unnecessary at this stage, not sure that we would define applications under an enum anyways?

substrate/in-instructions/primitives/src/incoming.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Target {
+  Application(ApplicationCall),
+  Address(NativeAddress),
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct InInstruction {
```
TheArchitect108: I think it would simplify things a lot if we have InternalInstruction and ExternalInstructions. InInstruction is confusing/ambiguous.

substrate/in-instructions/primitives/src/incoming.rs
```diff
@@ -0,0 +1,52 @@
...
+  pub origin: ExternalAddress,
+  pub target: Target,
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct ExternalInInstruction {
```
TheArchitect108: I feel like we're missing fields here. Only origin and target?

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+// Should be enough for a Uniswap v3 call
+pub const MAX_DATA_LEN: u32 = 512;
+
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct ExternalAddress(BoundedVec<u8, ConstU32<{ MAX_ADDRESS_LEN }>>);
+impl ExternalAddress {
+  #[cfg(feature = "std")]
+  pub fn new(address: Vec<u8>) -> Result<ExternalAddress, &'static str> {
+    Ok(ExternalAddress(address.try_into().map_err(|_| "address length exceeds {MAX_ADDRESS_LEN}")?))
```
TheArchitect108, line 25: let's move to a validation function call for external addresses? We probably need to add a lot of checks.

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  #[cfg(feature = "std")]
+  pub fn consume(self) -> Vec<u8> {
+    self.0.into_inner()
+  }
+}
+
+// Not "in" as "in" is a keyword
+mod incoming;
```
TheArchitect108, line 39: txin

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+pub use incoming::*;
+
+// Not "out" to match in
+mod outgoing;
```
TheArchitect108, line 43: txout?

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+pub use outgoing::*;
+
+mod shorthand;
```
TheArchitect108, line 46: txencoded?

substrate/in-instructions/primitives/src/shorthand.rs
```diff
@@ -0,0 +1,42 @@
...
+use sp_core::{ConstU32, bounded::BoundedVec};
+
+use serai_primitives::{NativeAddress, Coin, Amount};
+
+use crate::{MAX_DATA_LEN, ExternalAddress, ExternalInInstruction, OutInstruction};
+
+#[derive(Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub enum Shorthand {
+  Raw(BoundedVec<u8, ConstU32<{ MAX_DATA_LEN }>>),
```
TheArchitect108, line 16: Why would we ever map Raw?

substrate/in-instructions/primitives/src/shorthand.rs
```diff
@@ -0,0 +1,42 @@
...
+  Swap {
+    origin: Option<ExternalAddress>,
+    coin: Coin,
+    minimum: Amount,
+    out: OutInstruction,
+  },
+  AddLiquidity {
+    origin: Option<ExternalAddress>,
+    minimum: Amount,
+    gas: Amount,
```
TheArchitect108, line 26: why gas for add liq?

substrate/in-instructions/primitives/src/shorthand.rs
```diff
@@ -0,0 +1,42 @@
...
+  Raw(BoundedVec<u8, ConstU32<{ MAX_DATA_LEN }>>),
+  Swap {
+    origin: Option<ExternalAddress>,
+    coin: Coin,
+    minimum: Amount,
+    out: OutInstruction,
+  },
+  AddLiquidity {
+    origin: Option<ExternalAddress>,
+    minimum: Amount,
```
TheArchitect108, line 25: no minimum needed for add liq.

substrate/in-instructions/primitives/src/shorthand.rs
```diff
@@ -0,0 +1,42 @@
...
+    gas: Amount,
+    address: NativeAddress,
```
TheArchitect108: SeraiAddress?

substrate/in-instructions/primitives/src/shorthand.rs
```diff
@@ -0,0 +1,42 @@
...
+  },
+}
+
+impl TryFrom<Shorthand> for ExternalInInstruction {
+  type Error = &'static str;
+  fn try_from(shorthand: Shorthand) -> Result<ExternalInInstruction, &'static str> {
+    Ok(match shorthand {
+      Shorthand::Raw(raw) => {
+        ExternalInInstruction::decode(&mut raw.as_ref()).map_err(|_| "invalid raw instruction")?
```
TheArchitect108: ExternalInstructionIn

substrate/node/src/chain_spec.rs
```diff
@@ -17,22 +14,22 @@ fn insecure_pair_from_name(name: &'static str) -> Pair {
   Pair::from_string(&format!("//{name}"), None).unwrap()
 }
 
-fn account_id_from_name(name: &'static str) -> AccountId {
+fn address_from_name(name: &'static str) -> NativeAddress {
```
TheArchitect108: why did we switch from accountId to "NativeAddress"?

substrate/runtime/src/lib.rs
```diff
@@ -158,10 +173,10 @@ impl pallet_balances::Config for Runtime {
   type DustRemoval = ();
   type ExistentialDeposit = ConstU64<500>;
   type AccountStore = System;
-  type WeightInfo = pallet_balances::weights::SubstrateWeight<Runtime>;
```
TheArchitect108, line 161: why the move away from pallet_ convention?

substrate/serai/client/src/in_instructions.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+    Ok(res)
+  }
+
+  pub async fn get_coin_block_number(
+    &self,
+    coin: Coin,
+    block: [u8; 32],
+  ) -> Result<BlockNumber, SeraiError> {
+    Ok(self.storage(PALLET, "BlockNumbers", Some(coin), block).await?.unwrap_or(BlockNumber(0)))
```
TheArchitect108, line 47: To be clear, we are publishing the block number for each coin to substrate now?

substrate/serai/client/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+use primitives::{Signature, NativeAddress};
+
+use serai_runtime::{system::Config, Runtime};
+
+pub mod in_instructions;
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub(crate) struct SeraiConfig;
+impl SubxtConfig for SeraiConfig {
+  type BlockNumber = <Runtime as Config>::BlockNumber;
```
TheArchitect108, line 18: Why not just use the block struct? for these and below.

substrate/serai/primitives/src/lib.rs
```diff
@@ -1,7 +1,31 @@
...
+
+pub type PublicKey = Public;
+pub type NativeAddress = PublicKey;
+pub type Signature = RistrettoSignature;
+
+/// The type used to identify block numbers.
+// Doesn't re-export TendermintMachine due to traits.
+#[derive(Clone, Copy, PartialEq, Eq, Default, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+pub struct BlockNumber(pub u32);
```
TheArchitect108, line 31: I really feel like we should be mapping whole blocks where applicable, not just relying on the block number.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: A batch is intended to be one block. I'd have no object to the ID being the block hash (represented as `[u8; 32]`). It helps resolve the issue where BlockNumbers will tell you if you fall behind, yet not if you're eclipsed. Will change :+1:

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+    fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+      data
+        .get_data::<Updates>(&INHERENT_IDENTIFIER)
+        .unwrap()
+        .map(|updates| Call::execute { updates })
+    }
+
+    fn check_inherent(call: &Self::Call, data: &InherentData) -> Result<(), Self::Error> {
+      // Assumes that only not yet handled batches are provided as inherent data
+      if let Some(expected) = data.get_data::<Updates>(&INHERENT_IDENTIFIER).unwrap() {
```
kayabaNerve: This is critically flawed. If we don't provide data, it assumes valid.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -0,0 +1,236 @@
...
+        // Match to be exhaustive
+        let updates = match call {
+          Call::execute { ref updates } => updates,
+          _ => Err(InherentError::InvalidCall)?,
+        };
+
+        if updates.len() != expected.len() {
```
kayabaNerve: Should have a comment this is checking the amount of coins having updates provided matches the amount of coins we expect to have updates for.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: I'm unsure which cases you refer to. Mind detailing?

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: It gets whatever was last written. Since we now have a new batch, there now must be a new block. Accordingly, the last written number is old.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: The former is what the block producer wrote in. The latter is what we would've written in.

A malicious block producer, or one with a different view of the network, may have written in something different.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: This gets the expected ID of the next batch, verifying it matches the batch ID claimed in the Batch itself.

substrate/in-instructions/primitives/src/incoming.rs
kayabaNerve: It is unnecessary, yet if we ever write anything distinct, this allows us to do so in a non-breaking way for the cost of a single byte.

Without this, we'd need to do a breaking change in format if we ever offer any additional services. While I'm not here to harp for adding back a SC VM right now, or say we're shipping lending, I just figured this was the path offering the least resistance.

I believe using an enum absolutely makes sense though.

substrate/in-instructions/primitives/src/incoming.rs
kayabaNerve: It's because the entire call is crammed into Target.

This does also need an amount field in general, I just didn't get to that part of the pallet. Please revisit after I make the other discussed changes (EII dropped for RII).

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: We can't validate EAs here right now. Validation, which is on purposely defined as non-existent at this step, requires context of the coin we're running against and requires a provided BTC/ETH/XMR address format. This is agnostic to that by just being the raw bytes.

The processor should perform validation when converting from Vec<u8> to the actual type IMO.

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: I don't necessarily hate it? It's just awkward since we don't handle transactions in. We handle outputs/coins/messages in. That's why this was vague to "incoming" (or just "in", if that wasn't a key word).

We can edit to txin to be closer to the BTC standard, yet that raises its own discussions on the BTC standard being garbage.

Do you have any reason to dislike "incoming" so much or yet another option available?

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: 1) I'd rather not prefix everything with "tx"
2) No. All of these can be encoded on their own right. We just always use Shorthand for external networks representations.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: If you wanted to mint sriBTC without swapping or adding it.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: Minimum LP tokens. It prevents adding at expected ratio X:Y, having it frontrun to 2X:Y, yet getting credited for X:0.5y which your amount now is.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: I'm fine with that. Native just pleasantly matches External.

substrate/node/src/chain_spec.rs
kayabaNerve: We needed a primitive type for native addresses in serai-primitives. I added one which was equivalent to AccountId. AccountId didn't work as a name in the context of s-p. I then updated the runtime to re-export the primitives' type, instead of defining its own, to achieve a singular source of truth. While I could've had the runtime do `type AccountId = NativeAddress;`, I felt also have a singular name would be best. Given we're a cross-chain DEX, there is no "AccountId". Just the ID for an account on a chain. That's why specifying "Native" (or "Serai"), even within the Substrate code, is appreciated.

substrate/runtime/src/lib.rs
kayabaNerve: The line you replied to doesn't.

In general, I've used `_pallet` for three reasons.

1) `pallet` is under X. X is not under `pallet`. X has a pallet. It's much more legible to see project_part than part_project IMO.
2) It matches our folder layout, which uses the above commentary to define as `X/{primitives, pallet, client}`, unlike Substrate, minimizing travel distance/hunting length between the parts of a given project (no need to open a series of other folders/scroll all the way through 100 other pallets just to get to the pallet for the client you're working on).
3) It immediately says it's a Serai pallet, not a Substrate pallet. Tendermint bucks this convention as Tendermint was meant to be usable as a Substrate pallet, per https://github.com/serai-dex/serai/issues/137.

substrate/serai/client/src/in_instructions.rs
kayabaNerve: We need to in order to coordinate Monero properly.

If Serai doesn't have the inputs available, the first block at which they are available can be used without question. If Serai does, then the question is "is it block X, my current block, or block X -1, what everyone else has due to latency?" Placing the height on chain solves that.

It also provides a heartbeat, usable to detect desyncs, which helps with #6.

substrate/serai/client/src/lib.rs
kayabaNerve: The `BlockNumber(u32)` struct? Because Substrate requires a lot of traits on these types. I tried implementing it for `Amount` and it was ~300 lines of bs which broke several of the goals of the `Amount` type (float support, some math ops you likely shouldn't be using with `Amounts` is an uber-traditional system...).

substrate/serai/primitives/src/lib.rs
kayabaNerve: in-instructions can't use block hashes since it can't verify they're in a series. We can relax that check since its own view should be sorted? I'll look at that.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: It's the gas turned over to the Serai account receiving the LP tokens. It lets you add XMR:SRI liq from Monero, and when the time comes to remove it, not need to do a Monero TX to acquire SRI for gas. This is documented in the docs.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: Assuming this was for the match below this, this is now fixed. Please resolve if so

substrate/serai/client/src/lib.rs
TheArchitect108: #[derive(Debug, Clone)]
pub struct Block {
	pub hash: Hash,
	pub parent_hash: Hash,
	pub number: BlockNumber,
}

substrate/in-instructions/pallet/src/lib.rs
TheArchitect108: Maybe we should hash parts or whole of the batch somehow for better authenticity?

substrate/in-instructions/primitives/src/lib.rs
TheArchitect108: since validation is simple here we should to it every step but wont fight you on it.

substrate/in-instructions/primitives/src/lib.rs
TheArchitect108: Incoming is too ambiguous. I'm fine with OutputsIn or CoinsIn, just something other than incoming.

substrate/in-instructions/primitives/src/lib.rs
TheArchitect108: To understand shorthand someone has to look up the definition in our context. This in contrast to the transaction memo TC uses, TC wins on ease of understanding. We should add something to shorthand to make it more clear.

substrate/in-instructions/primitives/src/shorthand.rs
TheArchitect108: Why not a specific mint call then?

substrate/in-instructions/primitives/src/shorthand.rs
TheArchitect108: Ok but if we are not doing single sided add there's no swap interaction and the user must have/send TXs for both.

substrate/node/src/chain_spec.rs
TheArchitect108: The less we break from convention the better, this is a big one IMO. We should call this SeraiAddress, native is confusing.

substrate/runtime/src/lib.rs
TheArchitect108: I replied to the diff. using _pallet makes more sense to me.

substrate/serai/client/src/lib.rs
kayabaNerve: We don't have such a struct right now. My reason against it is IDs are optimally Copy-able. That exceeds intended Copy bounds. It's also shotgunning data and suggests it's fine to use the number, when it frequently isn't.

substrate/node/src/chain_spec.rs
kayabaNerve: Yes, will call it SeraiAddress. Are we cool not then re-exporting SeraiAddress as AccountId?

The main issue is when we use this in the processor/coordinator, and that sees AccountId due to grabbing Substrate integration code, yet it's using it in a multi-chain context. That's why I put this naming even into Substrate itself.

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: > requires context of the coin we're running against and requires a provided BTC/ETH/XMR address format.

It's impossible here unless every single ExternalAddress is ExternalAddress<C: Coin>, which then forces every single Instruction to be C: Coin, which breaks how it's intended to be a single type usable for all integrations since it's now type-locked to a specific coin. We do not have the context and cannot safely provide the context at this level.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: Mint isn't meaningfully smaller if we make a dedicated call, does increase our complexity, and if it replaces our exhaustive option, potentially leaves us incomplete.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: Right, yet that'd be a mint from BTC/XMR, not a Shorthand AddLiquidity, then a direct call on Serai. Else we need to code pending liquidity which is an abomination.

substrate/serai/primitives/src/lib.rs
kayabaNerve: The issue is we don't just verify the block number to be sequential. We verify a validator isn't saying block 9,999,999,999,999 just occured by ensuring the claimed block number <= our perceived block number. Replicating that check with block hashes would require making *every single hash* available under inherent data.

That's... fine? It's only actually the hashes since the last update? Though I really don't like it as we're dumping a Vec in we have to do an O(n) search on. I'd rather have In Instruction remain BlockNumber, yet use block hash for batch ID.

substrate/in-instructions/pallet/src/lib.rs
kayabaNerve: Resolved by using the block hash as the batch ID.

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: I feel every protocol has to be looked up to be properly understood. "Shorthand" is a long-standing term for a quick notation which expands to something. I'd argue "memo" is ambiguous since it suggests a message, not an instruction. THORChain isn't a chat app.

So yes, you have to look up what Shorthand expands to. That's called reading the docs to see how the protocol works. We can't handhold people through every single step of the protocol by how we name things alone. Even if it "InInstructionShorthand", they'd still have to lookup the options/fields. Even if it was "txencoded", they'd have to look up what the encoding format is and what the TX is.

"txencoded" is wrong for suggesting it's THE encoding when it isn't. These also aren't TXs. I'd like to close this and move on unless you can suggest a better name.

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: Except these aren't outputs and they're not coins. They're instructions. It's just annoying as hell to write `in_instructions::primitives::in_instructions::InInstruction` versus `in_instructions::primitives::in::InInstruction`. The members within, and the crate name, provide sufficient context IMO.

Also, this discussion is moot. This is literally a file name. It's written as `mod incoming; pub use incoming::*;` Accordingly, our public API is just `in_instructions::primitives::{InInstruction, OutInstruction}`. "incoming" is never in the public API. Can we drop this accordingly, please?

substrate/in-instructions/primitives/src/lib.rs
kayabaNerve: This is also an internal file name. It just has its members exported (`in_instructions::primitives::Shorthand`).

substrate/serai/client/src/lib.rs
kayabaNerve: To be clear, the processor is currently coded for numbers. I wouldn't mind moving it to block hashes. Since it's only operating past the confirmation depth it... *should* be fine? to use numbers, if we want to save the dev effort.

I'd also note that the recent Substrate upgrades have been removing usage *and* usability of numbers as IDs. I only believe in their usage for finalized data, and then as a project, we assume things sufficiently confirmed are final. Shotgunning them provides their accessibility where they arguably shouldn't be (and one of the ways Substrate removed their usability was by removing a variant of hash OR number).

I'd also note I did rewrite the Monero RPC to be hash-based when I wrote the Scanner scratch: https://github.com/serai-dex/serai/commit/1d6df0099ce62a5f80c00791668e8b7a46e43ffd

substrate/serai/primitives/src/lib.rs
TheArchitect108: Some blockchains reset their block number.

substrate/serai/client/src/lib.rs
TheArchitect108: We should use hashes.

substrate/node/src/chain_spec.rs
TheArchitect108: I vote for different types for external/native addresses. Not sure how we will use this in the processor/coordinator.

substrate/in-instructions/primitives/src/shorthand.rs
TheArchitect108: This does not make sense. You want liquidity to be asym deposit only?

substrate/in-instructions/primitives/src/shorthand.rs
TheArchitect108: This is an unnecessary optimization in my opinion.

substrate/in-instructions/primitives/src/shorthand.rs
TheArchitect108: We should be verbose, Raw tells us nothing.

substrate/node/src/chain_spec.rs
kayabaNerve: We have that :+1:

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: No. The flow for adding sym liq should be as follows:

- Mint sriX from X
- From an account with sriX and SRI, add liquidity.

This instruction is for the path where you don't have SRI and only want to add liquidity for X from X.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: This could be dropped for Mint alone and then the Serai call.

substrate/in-instructions/primitives/src/shorthand.rs
kayabaNerve: It's actually going to be `Shorthand::Raw(InInstruction::Transfer(serai_address))`. I do agree Raw isn't exactly descriptive there... We could offer a function `Shorthand::transfer(SeraiAddress)` to construct that?

PR #250, opened 2023-02-02
Coordinator

kayabaNerve:


kayabaNerve:
This still needs large amounts of cleaning to be considered.

Notable refactors:
- New MessageBox
- Removal of git subxt for serai-client
- Overall hygeine

This was a PR made off incomplete work, so that is to be expected.

PR #252, opened 2023-02-14
add monero seed support

akildemir:


coins/monero/Cargo.toml
```diff
@@ -15,6 +15,7 @@ rustdoc-args = ["--cfg", "docsrs"]
 hex-literal = "0.3"
 lazy_static = "1"
 thiserror = "1"
+crc = "3.0"
```
kayabaNerve: This should just be "3" so we use the latest release in the 3 series.

coins/monero/src/wallet/seed/classic/languages.rs
```diff
@@ -0,0 +1,67 @@
...
+const SPANISH_WORDS: &str = include_str!("./words/es.json");
+const RUSSIAN_WORDS: &str = include_str!("./words/ru.json");
+const JAPANESE_WORDS: &str = include_str!("./words/ja.json");
+const ESPERANTO_WORDS: &str = include_str!("./words/eo.json");
+const ENGLISH_OLD_WORDS: &str = include_str!("./words/ang.json");
+const PORTUGUESE_WORDS: &str = include_str!("./words/pt.json");
+const CHINESE_SIMPLIFIED_WORDS: &str = include_str!("./words/zh.json");
+
+/// returns all supported languages.
+pub fn all() -> Result<Vec<Language>, SeedError> {
```
kayabaNerve: This should be a lazy_static so it's only run once.

coins/monero/src/wallet/seed/classic/languages.rs
```diff
@@ -0,0 +1,67 @@
...
+      "Russian" => (LanguageName::Russian, RUSSIAN_WORDS),
+      "Japanese" => (LanguageName::Japanese, JAPANESE_WORDS),
+      "Esperanto" => (LanguageName::Esperanto, ESPERANTO_WORDS),
+      "EnglishOld" => (LanguageName::EnglishOld, ENGLISH_OLD_WORDS),
+      "Portuguese" => (LanguageName::Portuguese, PORTUGUESE_WORDS),
+      "ChineseSimplified" => (LanguageName::ChineseSimplified, CHINESE_SIMPLIFIED_WORDS),
+      _ => Err(SeedError::UnknownLanguage)?,
+    };
+    let mut lang = Language {
+      word_list: serde_json::from_str::<Vec<String>>(words).unwrap(),
```
kayabaNerve: from_str::<Vec<String> should be an unnecessary type specification.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+        utf8_prefix(word, self.unique_prefix_length)
+      } else {
+        word.to_string()
+      };
+
+      if self.trimmed_word_map.contains_key(&trimmed) && !allow_duplicate_prefix {
+        Err(SeedError::DuplicatePrefix)?;
+      }
+
+      self.trimmed_word_map.insert(trimmed, i);
```
kayabaNerve: This will override the previous value and make it impossible to properly use trimmed seeds. Is that intended behavior? For these languages to not support trimmed seeds?

If so, they should have None for trimmed_word_map, not an invalid map.

Also, since this is solely internal, this may be re-writable as something like
```rs
if self.trimmed_word_map.contains_key(&trimmed) {
  self.trimmed_word_map = None;
}
```

Then we can use tests to ensure languages which should support trimmed seeds do in fact support trimmed seeds.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  ) -> Result<(), SeedError> {
+    if self.word_list.len() != LANGUAGE_WORD_COUNT {
+      Err(SeedError::InvalidLanguageWordList)?;
+    }
+
+    for (i, word) in self.word_list.iter().enumerate() {
+      self.word_map.insert(word.clone(), i);
+
+      if word.len() < self.unique_prefix_length && !allow_short_words {
+        Err(SeedError::InvalidWordInLanguage)?;
```
kayabaNerve: If we're able to load all languages right now, this error is never hit. Accordingly, we don't need to have this error.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  word_list: Vec<String>,
+  word_map: HashMap<String, usize>,
+  trimmed_word_map: HashMap<String, usize>,
+  language_name: LanguageName,
+  unique_prefix_length: usize,
+}
+
+impl Language {
+  /// populates the word_map and trimmed_word_maps
+  fn populate_maps(
```
kayabaNerve: Should this be part of a Language constructor so it's impossible to have a Language which doesn't have populated maps?

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+
+      self.trimmed_word_map.insert(trimmed, i);
+    }
+
+    Ok(())
+  }
+}
+
+/// returns a new seed for a given lang.
+pub fn new(lang: LanguageName) -> Result<String, SeedError> {
```
kayabaNerve: Why can this fail? If it legitimately *can*, it should run until it doesn't fail and return that.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  let spend = random_scalar(&mut OsRng);
```
kayabaNerve: OsRng -> <R: RngCore + CryptoRng>

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  }
+  prefix
+}
+
+fn create_checksum_index(words: &[String], lang: &Language) -> Result<usize, SeedError> {
+  let mut trimmed_words = String::new();
+
+  for w in words {
+    let word = utf8_prefix(w, lang.unique_prefix_length);
+    if !lang.trimmed_word_map.contains_key(&word) {
```
kayabaNerve: Earlier, an invalid trimmed_word_map was created. Is this supposed to use trimmed_word_map? I thought the intent, for languages supporting trimmed words, was that wallets would expand to full words and then use those.

coins/monero/src/tests/seed.rs
```diff
@@ -0,0 +1,22 @@
+use crate::wallet::seed::{Seed, LanguageName};
+
+#[test]
+fn test_classic() {
+  // keys
+  let spend = "c0af65c0dd837e666b9d0dfed62745f4df35aed7ea619b2798a709f0fe545403";
+  let _view = "513ba91c538a5a9069e0094de90e927c0cd147fa10428ce3ac1afd49f63e3b01";
+
+  // seed string
+  let mut seed = String::from("washing thirsty occur lectures tuesday fainted toxic adapt ");
```
kayabaNerve: We need to test all languages, and their trimmed variants (for languages supporting trimming).

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  if seed_length != CLASSIC_SEED_LENGTH && seed_length != CLASSIC_SEED_LENGTH_WITH_CHECKSUM {
+    Err(SeedError::InvalidSeedLength)?;
+  }
+
+  // find the language
+  let has_checksum = seed_length == CLASSIC_SEED_LENGTH_WITH_CHECKSUM;
+  let (matched_indices, lang) = find_seed_language(&seed_words, has_checksum)?;
+
+  // convert to  bytes
+  let mut result: [u8; 32] = [b'0'; 32];
```
kayabaNerve: Why is this '0' and not 0?

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  let word_list_len = lang.word_list.len();
+  for i in 0 .. 8 {
+    let mut indices: [usize; 4] = [0; 4];
+    indices[1] = matched_indices[i * 3];
+    indices[2] = matched_indices[i * 3 + 1];
+    indices[3] = matched_indices[i * 3 + 2];
+
+    indices[0] = indices[1] +
```
kayabaNerve: Please use parentheses for all of these cases. It's a mess to read right now, partially due to the mental requirement for OoO, partially because rustfmt doesn't know how to handle it.

kayabaNerve: A comment explaining this would also be appreciated.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+      word_list_len *
+        word_list_len *
+        (((word_list_len - indices[2]) + indices[3]) % word_list_len);
+
+    if indices[0] % word_list_len != indices[1] {
+      Err(SeedError::InvalidSeed)?;
+    }
+
+    let pos = i * 4;
+    result[pos .. pos + 4].copy_from_slice(&(indices[0] as u32).to_le_bytes());
```
kayabaNerve: u32::try_from, not as, please

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+  }
+
+  Ok(result)
+}
+
+pub fn bytes_to_words(key: &[u8; 32], lang_name: LanguageName) -> Result<String, SeedError> {
+  // Grab the language, we can call unwrap because all language names must exist.
+  let lang = languages::all()?.into_iter().find(|l| l.language_name == lang_name).unwrap();
```
kayabaNerve: Should languages be a HashMap instead?

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+
+  // get the language words
+  let words = &lang.word_list;
+  let word_list_length = words.len() as u32;
```
kayabaNerve: u32::try_from

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+      }
+    }
+
+    if full_match && has_checksum && !checksum_test(seed, lang)? {
+      full_match = false;
+    }
+
+    if full_match {
+      matched_language = lang.clone();
+      return Ok((matched_indices, matched_language));
```
kayabaNerve: The matched_language variable doesn't need to exist. You can just return the language here.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    // no match for this language clear the indices
+    matched_indices.clear();
+  }
+
+  Err(SeedError::UnknownLanguage)?
+}
+
+fn utf8_prefix(word: &str, mut prefix_len: usize) -> String {
+  let mut prefix = String::new();
+  for a in word.chars() {
```
kayabaNerve: This isn't returning bytes yet Unicode scalars. Those may be what you want already? If they're not, and you do want bytes, use `as_bytes`.

coins/monero/src/wallet/seed/classic/mod.rs
```diff
@@ -0,0 +1,250 @@
...
+    if (a as u8) & 0xc0 == 0x80 {
+      prefix.push(a);
+      continue;
+    }
+    prefix.push(a);
+    prefix_len -= 1;
+
+    if prefix_len == 0 {
```
kayabaNerve: Use a for loop instead of this? `for (_, a) in (0 .. prefix_len).iter().zip(word.chars())` should work?

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+  Dutch,
+  Portuguese,
+  Russian,
+  ChineseSimplified,
+  Esperanto,
+  Lojban,
+}
+
+// TODO: Add polyseed to enum
+#[derive(PartialEq, Eq, Debug)]
```
kayabaNerve: This should derive Clone and not derive Debug. It should implement Debug, yet not print the String when it does.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  /// constructs a seed from a given key.
+  pub fn from_key(key: &[u8; 32], lang_name: LanguageName) -> Result<Seed, SeedError> {
+    Ok(Seed::Classic(classic::bytes_to_words(key, lang_name)?))
+  }
+
+  /// returns seed as String.
+  pub fn as_string(&self) -> &String {
```
kayabaNerve: This should be to_string, returning Zeroizing<String>, or a trait impl of ToString.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+        // convert to bytes to make sure it is valid
+        classic::words_to_bytes(&words)?;
+        Ok(Seed::Classic(words))
+      }
+      _ => Err(SeedError::InvalidSeedLength)?,
+    }
+  }
+
+  /// constructs a seed from a given key.
+  pub fn from_key(key: &[u8; 32], lang_name: LanguageName) -> Result<Seed, SeedError> {
```
kayabaNerve: Should this be able to fail?

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+impl Seed {
+  /// creates a new seed.
+  pub fn new(lang: LanguageName) -> Seed {
+    // TODO: This should return Polyseed when implemented.
+    Seed::Classic(classic::new(lang).unwrap())
+  }
+
+  /// constructs a seed from seed words.
+  pub fn from_string(words: String) -> Result<Seed, SeedError> {
```
kayabaNerve: The argument should be Zeroizing<String>.

coins/monero/src/wallet/seed/mod.rs
kayabaNerve: Also, it should be from_entropy, not from_key, and &Zeroizing<[u8; 32]>.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -0,0 +1,88 @@
...
+
+  /// returns seed as String.
+  pub fn as_string(&self) -> &String {
+    match self {
+      Seed::Classic(words) => words,
+    }
+  }
+
+  /// returns the seed spend key.
+  pub fn spend(&self) -> [u8; 32] {
```
kayabaNerve: Rename to entropy, return Zeroizing<[u8; 32]>, please.

kayabaNerve (changes requested):
Preliminary review. Happy to see this supported :D

coins/monero/src/wallet/seed/classic/mod.rs
akildemir: These errors are there to check the language words sanity, since anyone can just go to any language words file(e.g. `en.json`) and modify it. Only then it would throw the error. Otherwise yes it never hit.  If there is a way to make them completion errors instead of runtime, then yes we could remove them.

coins/monero/src/wallet/seed/classic/mod.rs
akildemir: we can fail when creating the checksum of the seed. And we can fail there because https://github.com/serai-dex/serai/pull/252#discussion_r1109453248

coins/monero/src/wallet/seed/classic/mod.rs
akildemir: As far as I read and understand the code, using trimmed words isn't about the language, it is about the specific seed type. Namely, one with checksum or without. It always uses the trimmed words for seed with checksum, uses the full words if there is not checksum in the seed. And it always make a seed with checksum when creating a new random seed, so it almost always uses trimmed words. I think the only reason using full words is there for historic seeds that doesn't have a checksum. Or you can create a seed without a checksum by will. It is just to have as support I guess.

coins/monero/src/tests/seed.rs
akildemir: Thought we would add all the tests when I switch back to tests.

coins/monero/src/wallet/seed/classic/mod.rs
kayabaNerve: Considering we're not modifying them, we're grabbing them from Monero, that shouldn't be necessary.

coins/monero/src/tests/seed.rs
kayabaNerve: If we're adding new features, they should be tested.

coins/monero/src/wallet/seed/classic/mod.rs
kayabaNerve: Per BIP39, all words are distinct after some amount of characters. That enables recovery of a seed, even if you only have 4 chars from each word. I don't believe it's about being forced to use trimmed/untrimmed. There's just no specific benefit to internally loading untrimmed seeds.

My primary comment is about how this assumes trimmed_word_map is valid when it was placed in an invalid state earlier, and how this shouldn't work on languages not supporting trimming.

kayabaNerve: We may need to
```rs
let mut word
if trimmed_word_map.is_some() {
  word = trimmed_word_map[word[.. unique_prefix_length]]
}
```
?

coins/monero/src/wallet/seed/classic/mod.rs
kayabaNerve: See my other comment which means this shouldn't fail.

PR #257, opened 2023-03-13
add polyseed support

akildemir:
add support for polyseed to monero coin. Implementation taken from https://github.com/tevador/polyseed/.


kayabaNerve:
Clippy failure is unrelated.

coins/monero/src/tests/seed.rs
```diff
@@ -6,27 +6,27 @@ use curve25519_dalek::scalar::Scalar;
...
+    language: classic::Language,
     seed: String,
     spend: String,
     view: String,
   }
 
   let vectors = [
     Vector {
-      language: Language::Chinese,
+      language: classic::Language::ChineseSimp,
```
kayabaNerve: Please spell out Simplified.

coins/monero/src/tests/seed.rs
```diff
@@ -163,15 +163,113 @@ fn test_classic_seed() {
...
+////// POLYSEED TESTS ///////////////
+///
+///
+#[test]
+fn test_polyseed() {
+  struct Vector {
+    language: polyseed::Language,
+    seed: String,
+    key: String,
+    birtday: u64,
```
kayabaNerve: Mispelling of birthday.

coins/monero/src/tests/seed.rs
```diff
@@ -163,15 +163,113 @@ fn test_classic_seed() {
...
+      duck valid someone little harsh puppy airport language"
+        .into(),
+      key: "dd76e7359a0ded37cd0ff0f3c829a5ae01673300000000000000000000000000".into(),
+      birtday: 1638446400,
+      formal: true,
+    },
+    Vector {
+      language: polyseed::Language::English,
+      seed: "raven tail swear infant grief assist regular lamp \
+      duck valid someone little harsh puppy airport language "
```
kayabaNerve: Can we automate testing with non-standard whitespace?

coins/monero/src/tests/seed.rs
```diff
@@ -163,15 +163,113 @@ fn test_classic_seed() {
...
+      seed: "eje fin parte celebre tabu pestana lienzo puma \
+      prision hora regalo lengua existir lapiz lote sonoro"
+        .into(),
+      key: "5a2b02df7db21fcbe6ec6df137d54c7b20fd2b00000000000000000000000000".into(),
+      birtday: 3118651200,
+      formal: false,
+    },
+    Vector {
+      language: polyseed::Language::Spanish,
+      seed: "eje fin part cele tabu pest lien puma \
```
kayabaNerve: Can we automate testing trimmed seeds?

coins/monero/src/tests/seed.rs
```diff
@@ -163,15 +163,113 @@ fn test_classic_seed() {
...
+      pris hora rega leng exis lapi lote sono"
+        .into(),
+      key: "5a2b02df7db21fcbe6ec6df137d54c7b20fd2b00000000000000000000000000".into(),
+      birtday: 3118651200,
+      formal: false,
+    },
+  ];
```
kayabaNerve, line 298: We need vectors for all languages. Feather Wallet should be usable for them?

coins/monero/src/tests/seed.rs
```diff
@@ -163,15 +163,113 @@ fn test_classic_seed() {
...
+    // string -> key
+    let s1 = Seed::from_string(Zeroizing::new(v.seed)).unwrap();
+    let key_bytes = Zeroizing::new(hex::decode(v.key).unwrap().try_into().unwrap());
+    let data = s1.polyseed_data();
+    assert_eq!(data.entropy, key_bytes);
+    assert_eq!(data.features, features);
+    assert!(data.birthday <= v.birtday);
+    assert!(data.birthday + polyseed_time_step > v.birtday);
+
+    if v.formal {
```
kayabaNerve: I ran into this when testing trimmed classic seeds. The seed parts should always trim and then expand, where possible, in order to always generate the full string value.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -21,44 +24,41 @@ pub enum SeedError {
...
-  Portuguese,
-  Japanese,
-  Russian,
-  Esperanto,
-  Lojban,
-  EnglishOld,
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+pub enum SeedType {
+  Classic(classic::Language),
+  PolySeed(polyseed::Language),
```
kayabaNerve: Polyseed has a lower case "s".

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -67,26 +67,52 @@ impl Seed {
...
       }
+      POLYSEED_LENGTH => PolySeed::from_string(words).map(Seed::PolySeed),
       _ => Err(SeedError::InvalidSeedLength)?,
     }
   }
 
   /// Create a Seed from entropy.
-  pub fn from_entropy(lang: Language, entropy: Zeroizing<[u8; 32]>) -> Option<Seed> {
-    ClassicSeed::from_entropy(lang, entropy).map(Seed::Classic)
+  pub fn from_entropy(seed_type: SeedType, entropy: Zeroizing<[u8; 32]>) -> Option<Seed> {
```
kayabaNerve: My intent for the entropy function was for it to return the bytes usable as a key (it being a rename of spend_key which is somewhat inaccurate under Polyseed). I now realize I've messed up in that the entropy *is* defined as the original 19-bytes?

We should add a key function to return the generated *key*. This would be the entropy for classic seeds yet the PBKDF output for Polyseed.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -67,26 +67,52 @@ impl Seed {
...
   /// Return the entropy for this seed.
   pub fn entropy(&self) -> Zeroizing<[u8; 32]> {
     match self {
       Seed::Classic(seed) => seed.entropy(),
+      Seed::PolySeed(seed) => seed.polyseed_data().entropy,
+    }
+  }
+
+  /// Return the polyseed data for this seed.
+  /// Return 0 values if it is a classic seed.
```
kayabaNerve: This should return an Option, not 0.

I'd also rather not have "polyseed_data", yet a dedicated birthday/features function. The birthday for a Classic seed should be 0 (not because it's unset but because no other value is safe). For features, it'd likely also be 0, signifying no features? Yet we may want to explicitly handle features in our API instead of relying on a uint expected to be interpreted per Polyseed. This may mean features returns Vec<Feature> or similar.

coins/monero/src/wallet/seed/mod.rs
```diff
@@ -67,26 +67,52 @@ impl Seed {
...
+        let birthday = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+        let data = PolySeedData { birthday, features: 0, entropy, checksum: 0 };
+        PolySeed::from_polyseed_data(data, lang).map(Seed::PolySeed)
+      }
+    }
+  }
+
+  /// Create a Seed from given data.
+  pub fn from_polyseed_data(data: PolySeedData, lang: polyseed::Language) -> Option<Seed> {
+    PolySeed::from_polyseed_data(data, lang).map(Seed::PolySeed)
```
kayabaNerve: Is it safe for the user to provide the checksum with this function?

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+  English,
+  Spanish,
+  French,
+  Italian,
+  Japanese,
+  Korean,
+  Czech,
+  Portuguese,
+  ChineseSimp,
+  ChineseTrad,
```
kayabaNerve: Please spell these out.

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+const IT_WORDS: &str = include_str!("./polyseed/it.json");
+const ES_WORDS: &str = include_str!("./polyseed/es.json");
+const JA_WORDS: &str = include_str!("./polyseed/ja.json");
+const PT_WORDS: &str = include_str!("./polyseed/pt.json");
+const ZH_S_WORDS: &str = include_str!("./polyseed/zh_simp.json");
+const ZH_T_WORDS: &str = include_str!("./polyseed/zh_trad.json");
+
+lazy_static! {
+  static ref LANGUAGES: HashMap<Language, WordList> = HashMap::from([
+    (Language::Czech, WordList::new(serde_json::from_str(CS_WORDS).unwrap(), true, false)),
```
kayabaNerve: You can have WordList::new perform this conversion to de-duplicate this.

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+fn poly_eval(poly: &Poly) -> u16 {
+  /* Horner's method at x = 2 */
+  let mut result = poly[POLYSEED_LENGTH - 1];
+  for i in (0 .. (POLYSEED_LENGTH - 1)).rev() {
+    result = elem_mul2(result) ^ poly[i];
+  }
+  result
+}
+
+fn poly_check(poly: &Poly) -> bool {
```
kayabaNerve: Does this really have to be a function when it only has one caller?

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+  pub fn new<R: RngCore + CryptoRng>(rng: &mut R) -> PolySeedData {
+    // get birthday
+    let time = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs();
+    let birthday = birthday_encode(time);
+
+    // get the entropy
+    let mut key = random_scalar(rng).to_bytes();
+    key[(SECRET_SIZE + 1) ..].fill(0);
+    key[SECRET_SIZE - 1] &= CLEAR_MASK;
+    let entropy = Zeroizing::new(key);
```
kayabaNerve:
```rs
let entropy = Zeroizing::new([0; 32]);
rng.fill_bytes(entropy.as_mut());
```

(and then the fill/clear) :)

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+
+    // make polyseed data
+    let mut data = PolySeedData { birthday, features: 0, entropy, checksum: 0 };
+
+    // set the  checksum
+    data.checksum = create_checksum(&data);
```
kayabaNerve: Add a constructor for PolyseedData from these parts so the checksum can't be unset?

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+    let features = make_features(data.features);
+    if !polyseed_features_supported(features) {
+      return None;
+    }
+
+    // get birthday
+    let birthday = birthday_encode(data.birthday);
+
+    // make sure it is a valid scalar
+    Scalar::from_canonical_bytes(*data.entropy)?;
```
kayabaNerve: ... why is this here? Anything which was created with

```rs
    let mut key = random_scalar(rng).to_bytes();
    key[(SECRET_SIZE + 1) ..].fill(0);
    key[SECRET_SIZE - 1] &= CLEAR_MASK;
    let entropy = Zeroizing::new(key);
```

will match this (as that's a 19-byte value needing to fit into a 31.5-byte space).

The concern I have is the non-19-byte values fitting into that 31.5 space.

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+  pub fn new<R: RngCore + CryptoRng>(rng: &mut R, lang: Language) -> PolySeed {
+    poly_encode(PolySeedData::new(rng), lang)
+  }
+
+  /// returns a new `PolySeed` for a given string.
+  pub fn from_string(words: Zeroizing<String>) -> Result<PolySeed, SeedError> {
+    let data = poly_decode(&words)?;
+
+    // Make sure this is a valid scalar
+    let mut scalar = Scalar::from_canonical_bytes(*data.entropy);
```
kayabaNerve: See prior comment on Scalar validity/the entropy's expected form.

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -0,0 +1,419 @@
...
+    // make polyseed data
+    let mut data = PolySeedData { birthday, features: 0, entropy, checksum: 0 };
+
+    // set the  checksum
+    data.checksum = create_checksum(&data);
+
+    data
+  }
+
+  pub fn from(data: PolySeedData) -> Option<PolySeedData> {
```
kayabaNerve: This really should be new(features, birthday, entropy). PolyseedData -> Option<PolyseedData> suggests not all PolyseedDatas are valid. If they're invalid, they shouldn't be creatable.

kayabaNerve (changes requested):

coins/monero/src/wallet/seed/mod.rs
akildemir: I think so because checksum is ignored and always overwrited according to other field values.

coins/monero/src/wallet/seed/polyseed.rs
akildemir: you mean `serde_json::from_str(CS_WORDS).unwrap()` this conversion?

coins/monero/src/wallet/seed/mod.rs
akildemir: the problem with having separate birthday/features function is that it is inefficient. We will need to re-decode the seed per call. In one scenario, user needs all polyseed data and makes 1 call to `polyseed_data`, in another user makes 3 call to get the data they want and we will re-decode the same seed 3 times, although all the data was already available after the first call.

coins/monero/src/wallet/seed/mod.rs
akildemir: function is converted into from(..args) now

coins/monero/src/wallet/seed/polyseed.rs
kayabaNerve: Yep!

PR #270, opened 2023-03-28
Attempt to the fix Bitcoin CI's cache statement

kayabaNerve:


kayabaNerve:
Works. Will merge after https://github.com/serai-dex/serai/actions/runs/4541546651/jobs/8003847145 finishes running, so it can take advantage of the cache that built.

PR #294, opened 2023-05-15
[Draft] Slash malevolent validators 

akildemir:
This is a **draft** pr to slash validators with bad behavior. Comments and/or reviews are appreciated.

in tributary, tendermint is used for the consensus. And there are bunch of defined invalid or bad behaviors for validators that participate in the tendermint consensus algorithm. We slash those nodes to disincentive those behaviors.

In this case, once a node catches an invalid/bad behavior from another validator, it either sends an evidence for this bad  behavior or a vote to slash the bad node. If an evidence provided and valid, it is solely enough to slash the node by the network participants even if other nodes didn't see this behavior. Once the evidence is on the chain, the bad node is slashed. If a vote is sent, then we wait vote from 2/3 of the nodes to slash the node.

For now following points are completed;

- create a special tx type to publish bad validator behavior data and/or votes for slashing
- modify the tributary to hold the both app txs and this integral tx
- ignore unsigned tendermint tx replays
- verify that provided evidence was actually valid(sender isn't falsely try to claim someone else with invalid evidence).
- scan the slash evidence/votes in the chain
- support unsigned app txs while at it(not really related to this pr but hey! why not)
- enforce provided, unsigned, signed tx ordering within a block
- add unit tests for tendermint tx validation
- update existing mempool/blockchain tests for new tx types

### DIFF
--- a/coordinator/src/tributary/db.rs
+++ b/coordinator/src/tributary/db.rs
@@ -39,14 +39,14 @@ impl<D: Db> TributaryDb<D> {
     Self::tributary_key(b"slash_point", [genesis, id].concat())
   }
 
-  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
-    Self::tributary_key(b"slash_vote", [genesis, id].concat())
+  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32], target: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_vote", [genesis, id, target].concat())
   }
 
   pub fn set_fatally_slashed(txn: &mut D::Transaction<'_>, genesis: [u8; 32], id: [u8; 32]) {
     let key = Self::fatal_slash_key(genesis);
     let mut existing = txn.get(&key).unwrap_or(vec![]);
-    
+
     // don't append if we already have it.
     if !existing.is_empty() && existing.chunks(32).any(|ex_id| ex_id == id) {
       return;
--- a/coordinator/src/tributary/db.rs
+++ b/coordinator/src/tributary/db.rs
@@ -31,6 +31,25 @@ impl<D: Db> TributaryDb<D> {
     self.0.get(Self::block_key(genesis)).map(|last| last.try_into().unwrap()).unwrap_or(genesis)
   }
 
+  fn fatal_slash_key(genesis: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"fatal_slash", genesis)
+  }
+
+  pub fn slash_point_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_point", [genesis, id].concat())
+  }
+
+  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_vote", [genesis, id].concat())
+  }
+
+  pub fn set_fatally_slashed(txn: &mut D::Transaction<'_>, genesis: [u8; 32], id: [u8; 32]) {
+    let key = Self::fatal_slash_key(genesis);
+    let mut existing = txn.get(&key).unwrap_or(vec![]);
+    existing.extend(id);
+    txn.put(key, existing);
+  }
+
   fn plan_ids_key(genesis: &[u8], block: u64) -> Vec<u8> {
     Self::tributary_key(b"plan_ids", [genesis, block.to_le_bytes().as_ref()].concat())
   }
--- a/coordinator/src/tributary/db.rs
+++ b/coordinator/src/tributary/db.rs
@@ -46,6 +46,12 @@ impl<D: Db> TributaryDb<D> {
   pub fn set_fatally_slashed(txn: &mut D::Transaction<'_>, genesis: [u8; 32], id: [u8; 32]) {
     let key = Self::fatal_slash_key(genesis);
     let mut existing = txn.get(&key).unwrap_or(vec![]);
+    
+    // don't append if we already have it.
+    if !existing.is_empty() && existing.chunks(32).any(|ex_id| ex_id == id) {
+      return;
+    }
+
     existing.extend(id);
     txn.put(key, existing);
   }
--- a/coordinator/src/tributary/db.rs
+++ b/coordinator/src/tributary/db.rs
@@ -31,25 +31,25 @@ impl<D: Db> TributaryDb<D> {
     self.0.get(Self::block_key(genesis)).map(|last| last.try_into().unwrap()).unwrap_or(genesis)
   }
 
-  fn fatal_slash_key(genesis: [u8; 32]) -> Vec<u8> {
-    Self::tributary_key(b"fatal_slash", genesis)
-  }
-
-  #[allow(unused)] // TODO
+  /* TODO
   pub fn slash_point_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
     Self::tributary_key(b"slash_point", [genesis, id].concat())
   }
+  */
 
-  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32], target: [u8; 32]) -> Vec<u8> {
-    Self::tributary_key(b"slash_vote", [genesis, id, target].concat())
+  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 13], target: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_vote", [genesis.as_slice(), &id, &target].concat())
   }
 
+  fn fatal_slash_key(genesis: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"fatal_slash", genesis)
+  }
   pub fn set_fatally_slashed(txn: &mut D::Transaction<'_>, genesis: [u8; 32], id: [u8; 32]) {
     let key = Self::fatal_slash_key(genesis);
     let mut existing = txn.get(&key).unwrap_or(vec![]);
 
     // don't append if we already have it.
-    if !existing.is_empty() && existing.chunks(32).any(|ex_id| ex_id == id) {
+    if existing.chunks(32).any(|ex_id| ex_id == id) {
       return;
     }
 
--- a/coordinator/src/tributary/db.rs
+++ b/coordinator/src/tributary/db.rs
@@ -35,6 +35,7 @@ impl<D: Db> TributaryDb<D> {
     Self::tributary_key(b"fatal_slash", genesis)
   }
 
+  #[allow(unused)] // TODO
   pub fn slash_point_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
     Self::tributary_key(b"slash_point", [genesis, id].concat())
   }
--- a/coordinator/tributary/src/block.rs
+++ b/coordinator/tributary/src/block.rs
@@ -180,9 +180,18 @@ impl<T: TransactionTrait> Block<T> {
   ) -> Result<(), BlockError> {
     #[derive(Clone, Copy, PartialEq, Eq, Debug)]
     enum Order {
-      Provided = 0,
-      Unsigned = 1,
-      Signed = 2,
+      Provided,
+      Unsigned,
+      Signed,
+    }
+    impl From<Order> for u8 {
+      fn from(order: Order) -> u8 {
+        match order {
+          Order::Provided => 0,
+          Order::Unsigned => 1,
+          Order::Signed => 2,
+        }
+      }
     }
 
     if self.serialize().len() > BLOCK_SIZE_LIMIT {
--- a/coordinator/tributary/src/block.rs
+++ b/coordinator/tributary/src/block.rs
@@ -152,6 +152,7 @@ impl<T: TransactionTrait> Block<T> {
     self.header.hash()
   }
 
+  #[allow(clippy::too_many_arguments)]
   pub(crate) fn verify<N: Network>(
     &self,
     genesis: [u8; 32],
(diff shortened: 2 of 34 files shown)


coordinator/tributary/src/blockchain.rs
```diff
@@ -65,17 +69,33 @@ impl<D: Db, T: TransactionTrait> Blockchain<D, T> {
       block_number: 0,
       tip: genesis,
       next_nonces,
+      unsigned_included: vec![],
```
kayabaNerve: Why is this here?

1) The mempool checks it for a TX's existence. That means it should be a HashSet, not a vec, if kept.
2) We shouldn't store in RAM a full list of every unsigned TX of all time. This should remain at the DB level. Pass the Mempool a lambda, or have it wrap its DB handle with BlockchainDb to get access to the function in BcDb.

coordinator/tributary/src/blockchain.rs
```diff
@@ -173,7 +193,21 @@ impl<D: Db, T: TransactionTrait> Blockchain<D, T> {
...
+        TransactionKind::Unsigned => {
+          let hash = tx.hash();
+          let key = Self::unsigned_included_key(&self.genesis);
+
+          // add to the included unsigned 
+          let mut existing = txn.get(&key).unwrap_or(vec![]);
+          existing.extend(hash);
+          txn.put(key, existing);
+
+          // add to memory
```
kayabaNerve: If it's in the DB, it shouldn't also be in memory.

coordinator/tributary/src/mempool.rs
```diff
@@ -45,6 +46,18 @@ impl<D: Db, T: TransactionTrait> Mempool<D, T> {
     self.txs.insert(tx_hash, tx);
   }
 
+  fn unsigned_already_exist(&self, unsigned_included: &[[u8; 32]], hash: [u8; 32]) -> bool {
+    // check we have the tx in the pool/chain
+    let res = unsigned_included.iter().find(|&&tx_hash| {
```
kayabaNerve: Should check the DB, not memory.

kayabaNerve (commented):
Just a few comments while I had a second. I don't think GH is showing me the full edits, so I'll have to re-open this to make the rest of my comments...

coordinator/src/tributary/db.rs
```diff
@@ -27,6 +27,25 @@ impl<D: Db> TributaryDb<D> {
...
+    Self::tributary_key(b"slash_point", [genesis, id].concat())
+  }
+
+  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_vote", [genesis, id].concat())
+  }
+
+  pub fn set_fatally_slashed(txn: &mut D::Transaction<'_>, genesis: [u8; 32], id: [u8; 32]) {
+    let key = Self::fatal_slash_key(genesis);
+    let mut existing = txn.get(&key).unwrap_or(vec![]);
```
kayabaNerve, line 49: This should check if the entity is already present.

Also, what's ID? Don't we want to slash keys, represented by Ristretto::G?

coordinator/src/tributary/db.rs
```diff
@@ -27,6 +27,25 @@ impl<D: Db> TributaryDb<D> {
...
 
+  fn fatal_slash_key(genesis: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"fatal_slash", genesis)
+  }
+
+  pub fn slash_point_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"slash_point", [genesis, id].concat())
+  }
+
+  pub fn slash_vote_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
```
kayabaNerve: We do vote on slashes by ID, yet we also need to track voted on slash ID -> underlying key.

coordinator/src/tributary/db.rs
```diff
@@ -27,6 +27,25 @@ impl<D: Db> TributaryDb<D> {
     self.0.get(Self::block_key(genesis)).map(|last| last.try_into().unwrap()).unwrap_or(genesis)
   }
 
+  fn fatal_slash_key(genesis: [u8; 32]) -> Vec<u8> {
+    Self::tributary_key(b"fatal_slash", genesis)
+  }
+
+  pub fn slash_point_key(genesis: [u8; 32], id: [u8; 32]) -> Vec<u8> {
```
kayabaNerve, line 35: Shouldn't this be Ristretto::G?

coordinator/tributary/src/block.rs
```diff
@@ -146,12 +152,14 @@ impl<T: Transaction> Block<T> {
     self.header.hash()
   }
 
-  pub(crate) fn verify(
+  pub(crate) fn verify<N: Network>(
     &self,
     genesis: [u8; 32],
     last_block: [u8; 32],
     mut locally_provided: HashMap<&'static str, VecDeque<T>>,
```
kayabaNerve, line 173: We also need to verify unsigned TXs aren't being included multiple times.

Preferably, we enforce an order of Provided, Unsigned, Signed as well.

coordinator/tributary/src/blockchain.rs
```diff
@@ -102,16 +129,42 @@ impl<D: Db, T: Transaction> Blockchain<D, T> {
...
   }
 
+  pub(crate) fn block_hash(&self, block: u32) -> Option<Vec<u8>> {
+    Self::block_hash_from_db(self.db.as_ref().unwrap(), self.genesis, block)
+  }
+
+  pub(crate) fn commit_by_block_number(&self, block: u32) -> Option<Vec<u8>> {
+    let hash = self.block_hash(block);
+    hash.as_ref()?;
+    self.commit(hash.unwrap().as_slice().try_into().unwrap())
```
kayabaNerve: This unwrap can be replaced with `?`, removing the prior line as well.

kayabaNerve: Also, why not have block_hash return [u8; 32]?

coordinator/tributary/src/blockchain.rs
```diff
@@ -102,16 +129,42 @@ impl<D: Db, T: Transaction> Blockchain<D, T> {
...
+  }
+
   pub(crate) fn block_after(db: &D, genesis: [u8; 32], block: &[u8; 32]) -> Option<[u8; 32]> {
     db.get(Self::block_after_key(&genesis, block)).map(|bytes| bytes.try_into().unwrap())
   }
 
-  pub(crate) fn add_transaction(&mut self, internal: bool, tx: T) -> bool {
-    self.mempool.add(&self.next_nonces, internal, tx)
+  pub(crate) fn add_transaction<N: Network>(&mut self, internal: bool, tx: Transaction<T>, schema: N::SignatureScheme) -> bool {
+
```
kayabaNerve: Random new line.

coordinator/tributary/src/blockchain.rs
```diff
@@ -102,16 +129,42 @@ impl<D: Db, T: Transaction> Blockchain<D, T> {
...
+    let db = self.db.as_ref().unwrap();
+    let genesis = self.genesis;
+    let commit = |block: u32| -> Option<Commit<N::SignatureScheme>> {
+      let hash = Self::block_hash_from_db(db, genesis, block);
+      hash.as_ref()?;
+      // we must have a commit per valid hash
+      let commit = Self::commit_from_db(db, genesis, hash.unwrap().as_slice().try_into().unwrap()).unwrap();
```
kayabaNerve: Same comment on block_hash handling as prior comment discussing it.

coordinator/tributary/src/blockchain.rs
```diff
@@ -156,10 +217,12 @@ impl<D: Db, T: Transaction> Blockchain<D, T> {
 
     self.tip = block.hash();
     txn.put(self.tip_key(), self.tip);
-
+
```
kayabaNerve: Random tab.

coordinator/tributary/src/mempool.rs
```diff
@@ -4,108 +4,165 @@ use ciphersuite::{Ciphersuite, Ristretto};
...
+        Transaction::Application(tx) => {
+          match tx.kind() {
+            TransactionKind::Signed(Signed { signer, nonce, .. }) => {
+              if let Some(prev) = res.next_nonces.insert(*signer, nonce + 1) {
+                // These mempool additions should've been ordered
+                assert!(prev < *nonce);
+              }
+              res.txs.insert(hash, Transaction::Application(tx));
+            }
+            _ => panic!("mempool database had a non-signed application transaction"),
```
kayabaNerve: Unsigned transactions are now mempool routed, so we need to handle Transaction::Unsigned and update this error to "non-provided".

coordinator/tributary/src/mempool.rs
```diff
@@ -4,108 +4,165 @@ use ciphersuite::{Ciphersuite, Ristretto};
...
-        // If we have too many transactions from this sender, don't add this yet UNLESS we are
-        // this sender
-        if !internal && (nonce >= &(blockchain_next_nonce + ACCOUNT_MEMPOOL_LIMIT)) {
+    match &tx {
+      Transaction::Tendermint(tendermint_tx) => {
+        assert_eq!(TransactionKind::Unsigned, tendermint_tx.kind());
+
+        // check we have the tx in the pool/chain
+        if self.unsigned_already_exist(unsigned_included, tx.hash()) {
+          // TODO: should this return true? function comment says return false
```
kayabaNerve: Returning false is correct.

coordinator/tributary/src/mempool.rs
```diff
@@ -118,27 +175,31 @@ impl<D: Db, T: Transaction> Mempool<D, T> {
...
     for hash in self.txs.keys().cloned().collect::<Vec<_>>() {
       let tx = &self.txs[&hash];
+
       // Verify this hasn't gone stale
-      match tx.kind() {
-        TransactionKind::Signed(Signed { signer, nonce, .. }) => {
-          if blockchain_next_nonces[signer] > *nonce {
-            self.remove(&hash);
-            continue;
+      if let Transaction::Application(tx) =  tx {
```
kayabaNerve: `=  tx`? There's two spaces there.

coordinator/tributary/src/mempool.rs
```diff
@@ -118,27 +175,31 @@ impl<D: Db, T: Transaction> Mempool<D, T> {
...
+        match tx.kind() {
+          TransactionKind::Signed(Signed { signer, nonce, .. }) => {
+            if blockchain_next_nonces[signer] > *nonce {
+              self.remove(&hash);
+              continue;
+            }
           }
+          _ => panic!("non-signed transaction entered mempool"),
```
kayabaNerve: Same commentary on above about handling unsigned TXs here. We just need to check they weren't already included.

coordinator/tributary/src/mempool.rs
```diff
@@ -118,27 +175,31 @@ impl<D: Db, T: Transaction> Mempool<D, T> {
...
         }
-        _ => panic!("non-signed transaction entered mempool"),
       }
 
       // Since this TX isn't stale, include it
+      // TODO: This also includes non-signed Tendermint Evidence txs. should it?
```
kayabaNerve: Yes, yet preferably in a seperate vec so we can return `unsigned || signed`, and enforce that ordering (unsigned before signed, just as we currently have provided before signed).

kayabaNerve (commented):
I reviewed about half of this. Looks solid so far.

coordinator/tributary/src/mempool.rs
akildemir: Don't we already check whether a given unsigned is already in chain when adding to the pool? why do we need to re check here?

coordinator/tributary/src/mempool.rs
akildemir: why do we have that ordering? can't a block contain all 3 types at the same time? or you mean the tx ordering within the same block?

kayabaNerve: Within a block, ideally the transactions are ordered

Provided
Unsigned
Signed

coordinator/tributary/src/mempool.rs
kayabaNerve: @akildemir Application transactions can be unsigned and are allowed in the mempool so long as they're valid.

coordinator/tributary/src/mempool.rs
kayabaNerve: This still has an extra space.

coordinator/tributary/src/mempool.rs
kayabaNerve: This block is checking if the transaction didn't go stale (included on chain). It has to be run on all TXs, not just application TXs. Unsigned TXs are stale if they were already included on chain.

Technically, if they're included on chain, they should be removed from the mempool and accordingly should never be present in the mempool if they're stale. This is distinct from Signed where another TX with the same nonce may enter, making the TX in our mempool stale, despite not being on-chain.

The reason I still want you to handle unsigned TXs here is:
1) It's assuming a lack of a race condition which is exceptionally easy to add here. I'd rather be robust.
2) You iterate over all TXs. If any unsigned TX is in the pool, this will panic. While you limited it to Application TXs, so it wouldn't panic on unsigned Tendermint TXs, application TXs are allowed to be unsigned. That means an unsigned application TX will hit this panic.

coordinator/src/tributary/db.rs
akildemir: ID is the ValidatorId from `msg.sender` which I believe is already Ristretto::G.

coordinator/src/tributary/db.rs
akildemir: you mean who sent this vote? IF so, that is already a TODO atm.

akildemir: you mean who sent this vote? If so, that is a TODO atm.

Issue #291, opened 2023-05-09
Tributary sync test intermittently fails


kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/4921632496/jobs/8791653466 failed, with the next run passing.

kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/4930494640/jobs/8811477210 with a more descriptive error.

kayabaNerve:
Unfortunately, this is frequently failing.

kayabaNerve:
Supposedly, the odds of participating in the next commit were much lower than expected. We may have to re-open this if that theory proves incorrect but for now...

PR #301, opened 2023-06-02
monero-serai: fee calculation parity with Monero's wallet2

j-berman:
The goal of this PR is to ensure `monero-serai` constructs transactions using fees that are indistinguishable from Monero's wallet2. This PR accomplishes this goal to the best of my knowledge. I haven't tested it with mainnet yet but plan on doing so (while working on the decoy selection algorithm next).

The core internal difference this PR introduces to `monero-serai` transaction construction logic is that decoys must be selected *before* first calculating the fee. This is because a ring's `key_offsets` can vary in byte length between decoy selection attempts. By fixing the selected decoys at the start of tx construction, the tx builder can immediately calculate the tx's weight, calculate the fee based on that weight, and settle on the user's change.

Of note, there is a theoretical discrepancy to wallet2 that I noted in a comment in the function `calculate_weight_and_fee`. This discrepancy should never apply in practice. In short, wallet2 can theoretically use an estimate that is significantly different from what the calculated tx fee should be, which could cause wallet2 to use a fee that is not calculated from the final tx's weight. This does not occur in practice to the best of my knowledge; wallet2 *always* under-estimates the fee today by a small margin (it could have happened in the past when there were very few outputs in the chain). The reason I kept this theoretical discrepancy in `monero-serai` is because handling it would require copying brittle code from wallet2 that has oddities (and what look like bugs) that would be a pain to integrate cleanly for no practical gain (e.g. wallet2 does not factor in the expected final weight of tx extra in its initial estimate, while the CLI, RPC, and GUI do factor in the estimated weight of a payment ID in tx extra only if sending to an integrated address).

### DIFF
--- a/coins/monero/src/lib.rs
+++ b/coins/monero/src/lib.rs
@@ -55,7 +55,13 @@ pub(crate) fn INV_EIGHT() -> Scalar {
 pub enum Protocol {
   v14,
   v16,
-  Custom { ring_len: usize, bp_plus: bool, optimal_rct_type: RctType },
+  Custom {
+    ring_len: usize,
+    bp_plus: bool,
+    optimal_rct_type: RctType,
+    view_tags: bool,
+    v16_fee: bool,
+  },
 }
 
 impl Protocol {
@@ -88,16 +94,37 @@ impl Protocol {
     }
   }
 
+  /// Whether or not the specified version uses view tags.
+  pub fn view_tags(&self) -> bool {
+    match self {
+      Protocol::v14 => false,
+      Protocol::v16 => true,
+      Protocol::Custom { view_tags, .. } => *view_tags,
+    }
+  }
+
+  /// Whether or not the specified version uses the fee algorithm from Monero
+  /// hard fork version 16 (released in v18 binaries).
+  pub fn v16_fee(&self) -> bool {
+    match self {
+      Protocol::v14 => false,
+      Protocol::v16 => true,
+      Protocol::Custom { v16_fee, .. } => *v16_fee,
+    }
+  }
+
   pub(crate) fn write<W: io::Write>(&self, w: &mut W) -> io::Result<()> {
     match self {
       Protocol::v14 => w.write_all(&[0, 14]),
       Protocol::v16 => w.write_all(&[0, 16]),
-      Protocol::Custom { ring_len, bp_plus, optimal_rct_type } => {
+      Protocol::Custom { ring_len, bp_plus, optimal_rct_type, view_tags, v16_fee } => {
         // Custom, version 0
         w.write_all(&[1, 0])?;
         w.write_all(&u16::try_from(*ring_len).unwrap().to_le_bytes())?;
         w.write_all(&[u8::from(*bp_plus)])?;
-        w.write_all(&[optimal_rct_type.to_byte()])
+        w.write_all(&[optimal_rct_type.to_byte()])?;
+        w.write_all(&[u8::from(*view_tags)])?;
+        w.write_all(&[u8::from(*v16_fee)])
       }
     }
   }
@@ -121,6 +148,16 @@ impl Protocol {
           },
           optimal_rct_type: RctType::from_byte(read_byte(r)?)
             .ok_or_else(|| io::Error::new(io::ErrorKind::Other, "invalid RctType serialization"))?,
+          view_tags: match read_byte(r)? {
+            0 => false,
+            1 => true,
+            _ => Err(io::Error::new(io::ErrorKind::Other, "invalid bool serialization"))?,
+          },
+          v16_fee: match read_byte(r)? {
+            0 => false,
+            1 => true,
+            _ => Err(io::Error::new(io::ErrorKind::Other, "invalid bool serialization"))?,
+          },
         },
         _ => {
           Err(io::Error::new(io::ErrorKind::Other, "unrecognized custom protocol serialization"))?
--- a/coins/monero/src/ringct/bulletproofs/mod.rs
+++ b/coins/monero/src/ringct/bulletproofs/mod.rs
@@ -37,14 +37,16 @@ pub enum Bulletproofs {
 }
 
 impl Bulletproofs {
-  pub(crate) fn fee_weight(plus: bool, outputs: usize) -> usize {
-    #[allow(non_snake_case)]
-    let (bp_clawback, LR_len) = Bulletproofs::calculate_bp_clawback(plus, outputs);
-    32 * (Bulletproofs::bp_fields(plus) + (2 * LR_len)) + 2 + bp_clawback
+  fn bp_fields(plus: bool) -> usize {
+    if plus {
+      6
+    } else {
+      9
+    }
   }
 
   // https://github.com/monero-project/monero/blob/94e67bf96bbc010241f29ada6abc89f49a81759c/
-  // src/cryptonote_basic/cryptonote_format_utils.cpp#L106-L124
+  //   src/cryptonote_basic/cryptonote_format_utils.cpp#L106-L124
   pub(crate) fn calculate_bp_clawback(plus: bool, n_outputs: usize) -> (usize, usize) {
     #[allow(non_snake_case)]
     let mut LR_len = 0;
@@ -66,12 +68,10 @@ impl Bulletproofs {
     (bp_clawback, LR_len)
   }
 
-  fn bp_fields(plus: bool) -> usize {
-    if plus {
-      6
-    } else {
-      9
-    }
+  pub(crate) fn fee_weight(plus: bool, outputs: usize) -> usize {
+    #[allow(non_snake_case)]
+    let (bp_clawback, LR_len) = Bulletproofs::calculate_bp_clawback(plus, outputs);
+    32 * (Bulletproofs::bp_fields(plus) + (2 * LR_len)) + 2 + bp_clawback
   }
 
   /// Prove the list of commitments are within [0 .. 2^64).
--- a/coins/monero/src/ringct/bulletproofs/mod.rs
+++ b/coins/monero/src/ringct/bulletproofs/mod.rs
@@ -38,23 +38,40 @@ pub enum Bulletproofs {
 
 impl Bulletproofs {
   pub(crate) fn fee_weight(plus: bool, outputs: usize) -> usize {
-    let fields = if plus { 6 } else { 9 };
-
-    // TODO: Shouldn't this use u32/u64?
     #[allow(non_snake_case)]
-    let mut LR_len = usize::try_from(usize::BITS - (outputs - 1).leading_zeros()).unwrap();
-    let padded_outputs = 1 << LR_len;
+    let (bp_clawback, LR_len) = Bulletproofs::calculate_bp_clawback(plus, outputs);
+    32 * (Bulletproofs::bp_fields(plus) + (2 * LR_len)) + 2 + bp_clawback
+  }
+
+  // https://github.com/monero-project/monero/blob/94e67bf96bbc010241f29ada6abc89f49a81759c/
+  // src/cryptonote_basic/cryptonote_format_utils.cpp#L106-L124
+  pub(crate) fn calculate_bp_clawback(plus: bool, n_outputs: usize) -> (usize, usize) {
+    #[allow(non_snake_case)]
+    let mut LR_len = 0;
+    let mut n_padded_outputs = 1;
+    while n_padded_outputs < n_outputs {
+      LR_len += 1;
+      n_padded_outputs = 1 << LR_len;
+    }
     LR_len += LOG_N;
 
-    let len = (fields + (2 * LR_len)) * 32;
-    len +
-      if padded_outputs <= 2 {
-        0
-      } else {
-        let base = ((fields + (2 * (LOG_N + 1))) * 32) / 2;
-        let size = (fields + (2 * LR_len)) * 32;
-        ((base * padded_outputs) - size) * 4 / 5
-      }
+    let mut bp_clawback = 0;
+    if n_padded_outputs > 2 {
+      let fields = Bulletproofs::bp_fields(plus);
+      let base = ((fields + (2 * (LOG_N + 1))) * 32) / 2;
+      let size = (fields + (2 * LR_len)) * 32;
+      bp_clawback = ((base * n_padded_outputs) - size) * 4 / 5;
+    }
+
+    (bp_clawback, LR_len)
+  }
+
+  fn bp_fields(plus: bool) -> usize {
+    if plus {
+      6
+    } else {
+      9
+    }
   }
 
   /// Prove the list of commitments are within [0 .. 2^64).
--- a/coins/monero/src/serialize.rs
+++ b/coins/monero/src/serialize.rs
@@ -11,8 +11,18 @@ use curve25519_dalek::{
 
 const VARINT_CONTINUATION_MASK: u8 = 0b1000_0000;
 
-pub(crate) fn varint_len(varint: usize) -> usize {
-  ((usize::try_from(usize::BITS - varint.leading_zeros()).unwrap().saturating_sub(1)) / 7) + 1
+mod sealed {
+  pub trait VarInt: TryInto<u64> {}
+  impl VarInt for u8 {}
+  impl VarInt for u32 {}
+  impl VarInt for u64 {}
+  impl VarInt for usize {}
+}
+
+/// This will panic if the varint exceeds u64::MAX
+pub(crate) fn varint_len<U: sealed::VarInt>(varint: U) -> usize {
+  let varint_u64: u64 = varint.try_into().map_err(|_| "varint exceeded u64").unwrap();
+  ((usize::try_from(u64::BITS - varint_u64.leading_zeros()).unwrap().saturating_sub(1)) / 7) + 1
 }
 
 pub(crate) fn write_byte<W: Write>(byte: &u8, w: &mut W) -> io::Result<()> {
--- a/coins/monero/src/serialize.rs
+++ b/coins/monero/src/serialize.rs
@@ -19,7 +19,7 @@ mod sealed {
   impl VarInt for usize {}
 }
 
-/// This will panic if the varint exceeds u64::MAX
+// This will panic if the VarInt exceeds u64::MAX
 pub(crate) fn varint_len<U: sealed::VarInt>(varint: U) -> usize {
   let varint_u64: u64 = varint.try_into().map_err(|_| "varint exceeded u64").unwrap();
   ((usize::try_from(u64::BITS - varint_u64.leading_zeros()).unwrap().saturating_sub(1)) / 7) + 1
--- a/coins/monero/src/transaction.rs
+++ b/coins/monero/src/transaction.rs
@@ -14,7 +14,7 @@ use curve25519_dalek::{
 use crate::{
   Protocol, hash,
   serialize::*,
-  ringct::{RctBase, RctPrunable, RctSignatures, bulletproofs::Bulletproofs, RctType},
+  ringct::{bulletproofs::Bulletproofs, RctType, RctBase, RctPrunable, RctSignatures},
 };
 
 #[derive(Clone, PartialEq, Eq, Debug)]
@@ -409,17 +409,16 @@ impl Transaction {
     }
   }
 
-  /// Calculate the transaction's final weight
+  /// Calculate the transaction's weight.
   pub fn weight(&self) -> usize {
     let blob_size = self.serialize().len();
 
     let bp = self.is_rct_bulletproof();
     let bp_plus = self.is_rct_bulletproof_plus();
-    if !bp && !bp_plus {
+    if !(bp || bp_plus) {
       blob_size
     } else {
-      let n_outputs = self.prefix.outputs.len();
-      blob_size + Bulletproofs::calculate_bp_clawback(bp_plus, n_outputs).0
+      blob_size + Bulletproofs::calculate_bp_clawback(bp_plus, self.prefix.outputs.len()).0
     }
   }
 }
--- a/coins/monero/src/transaction.rs
+++ b/coins/monero/src/transaction.rs
@@ -14,7 +14,7 @@ use curve25519_dalek::{
 use crate::{
   Protocol, hash,
   serialize::*,
-  ringct::{RctBase, RctPrunable, RctSignatures},
+  ringct::{RctBase, RctPrunable, RctSignatures, bulletproofs::Bulletproofs, RctType},
 };
 
 #[derive(Clone, PartialEq, Eq, Debug)]
@@ -24,11 +24,10 @@ pub enum Input {
 }
 
 impl Input {
-  // Worst-case predictive len
-  pub(crate) fn fee_weight(ring_len: usize) -> usize {
+  pub(crate) fn fee_weight(offsets_weight: usize) -> usize {
+    // Uses 1 byte for the input type
     // Uses 1 byte for the VarInt amount due to amount being 0
-    // Uses 1 byte for the VarInt encoding of the length of the ring as well
-    1 + 1 + 1 + (8 * ring_len) + 32
+    1 + 1 + offsets_weight + 32
   }
 
   pub fn write<W: Write>(&self, w: &mut W) -> io::Result<()> {
@@ -86,8 +85,10 @@ pub struct Output {
 }
 
 impl Output {
-  pub(crate) fn fee_weight() -> usize {
-    1 + 1 + 32 + 1
+  pub(crate) fn fee_weight(view_tags: bool) -> usize {
+    // Uses 1 byte for the output type
+    // Uses 1 byte for the VarInt amount due to amount being 0
+    1 + 1 + 32 + if view_tags { 1 } else { 0 }
   }
 
   pub fn write<W: Write>(&self, w: &mut W) -> io::Result<()> {
@@ -185,13 +186,19 @@ pub struct TransactionPrefix {
 }
 
 impl TransactionPrefix {
-  pub(crate) fn fee_weight(ring_len: usize, inputs: usize, outputs: usize, extra: usize) -> usize {
+  pub(crate) fn fee_weight(
+    decoy_weights: &[usize],
+    outputs: usize,
+    view_tags: bool,
+    extra: usize,
+  ) -> usize {
     // Assumes Timelock::None since this library won't let you create a TX with a timelock
+    // 1 input for every decoy weight
     1 + 1 +
-      varint_len(inputs) +
-      (inputs * Input::fee_weight(ring_len)) +
-      1 +
-      (outputs * Output::fee_weight()) +
+      varint_len(decoy_weights.len()) +
+      decoy_weights.iter().map(|&offsets_weight| Input::fee_weight(offsets_weight)).sum::<usize>() +
+      varint_len(outputs) +
+      (outputs * Output::fee_weight(view_tags)) +
       varint_len(extra) +
       extra
   }
@@ -253,12 +260,13 @@ pub struct Transaction {
 impl Transaction {
   pub(crate) fn fee_weight(
     protocol: Protocol,
-    inputs: usize,
+    decoy_weights: &[usize],
     outputs: usize,
     extra: usize,
+    fee: u64,
   ) -> usize {
-    TransactionPrefix::fee_weight(protocol.ring_len(), inputs, outputs, extra) +
-      RctSignatures::fee_weight(protocol, inputs, outputs)
+    TransactionPrefix::fee_weight(decoy_weights, outputs, protocol.view_tags(), extra) +
+      RctSignatures::fee_weight(protocol, decoy_weights.len(), outputs, fee)
   }
 
   pub fn write<W: Write>(&self, w: &mut W) -> io::Result<()> {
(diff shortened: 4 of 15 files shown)


coins/monero/src/rpc.rs
```diff
@@ -513,16 +513,27 @@ impl<R: RpcConnection> Rpc<R> {
...
       quantization_mask: u64,
     }
 
     let res: FeeResponse = self.json_rpc_call("get_fee_estimate", None).await?;
-    Ok(Fee { per_weight: res.fee, mask: res.quantization_mask })
+
+    let priority_idx =
+      if priority >= 4 { 3 } else { priority.saturating_sub(1).try_into().unwrap() };
+
+    debug_assert!(priority_idx < res.fees.len(), "Unexpected size of fee response");
```
kayabaNerve: This allows a malicious node to crash any app with debug assertions. Please use an error here since this can't be stated as an invariant.

coins/monero/src/rpc.rs
```diff
@@ -513,16 +513,27 @@ impl<R: RpcConnection> Rpc<R> {
...
+
+    // Using core Monero wallet2.cpp defaults if it can't use a fee from response
```
kayabaNerve: Mind explaining the safety of this? Should we just return an error instead?

coins/monero/src/rpc.rs
```diff
@@ -513,16 +513,27 @@ impl<R: RpcConnection> Rpc<R> {
...
+    if priority_idx >= res.fees.len() {
+      Ok(Fee { per_weight: 300000 /* wallet2 FEE_PER_BYTE */, mask: 1 })
```
kayabaNerve: `res.fees.get(priority_idx).unwrap_or(300000)`

Also, is this meant to be static for all priorities? And a mask of 1???

coins/monero/src/transaction.rs
```diff
@@ -164,12 +165,13 @@ pub struct TransactionPrefix {
 }
 
 impl TransactionPrefix {
-  pub(crate) fn fee_weight(ring_len: usize, inputs: usize, outputs: usize, extra: usize) -> usize {
+  pub(crate) fn fee_weight(decoy_weights: &Vec<usize>, outputs: usize, extra: usize) -> usize {
```
kayabaNerve: &[], not &Vec

coins/monero/src/transaction.rs
```diff
@@ -214,12 +216,18 @@ pub struct Transaction {
 impl Transaction {
   pub(crate) fn fee_weight(
     protocol: Protocol,
+    decoy_weights: &Vec<usize>,
```
kayabaNerve: &[], not &Vec<>

coins/monero/src/transaction.rs
```diff
@@ -214,12 +216,18 @@ pub struct Transaction {
...
     inputs: usize,
     outputs: usize,
     extra: usize,
+    fee: u64,
   ) -> usize {
-    TransactionPrefix::fee_weight(protocol.ring_len(), inputs, outputs, extra) +
-      RctSignatures::fee_weight(protocol, inputs, outputs)
+    if inputs != decoy_weights.len() {
```
kayabaNerve: Replace inputs with decoy_weights.len()? Why take in the variable twice if it just introduces a potential inconsistency?

coins/monero/src/wallet/decoys.rs
```diff
@@ -16,6 +16,8 @@ use crate::{
   rpc::{RpcError, RpcConnection, Rpc},
 };
 
+use crate::serialize::{varint_len};
```
kayabaNerve: Please merge this use with the prior use crate.

coins/monero/src/wallet/decoys.rs
```diff
@@ -131,7 +133,12 @@ pub struct Decoys {
   pub ring: Vec<[EdwardsPoint; 2]>,
 }
 
+#[allow(clippy::len_without_is_empty)]
 impl Decoys {
+  pub fn fee_weight(offsets: &Vec<u64>) -> usize {
```
kayabaNerve: &[], not &Vec<>

coins/monero/src/wallet/send/builder.rs
```diff
@@ -48,6 +50,13 @@ impl SignableTransactionBuilderInternal {
     self.inputs.extend(inputs.iter().cloned());
   }
 
+  fn add_decoy(&mut self, decoy: Decoys) {
+    self.decoys.push(decoy);
+  }
+  fn add_decoys(&mut self, decoys: &[Decoys]) {
```
kayabaNerve: Why not add these with inputs?

coins/monero/src/wallet/send/mod.rs
```diff
@@ -147,27 +147,14 @@ pub enum TransactionError {
   FrostError(FrostError),
 }
 
-async fn prepare_inputs<R: RngCore + CryptoRng, RPC: RpcConnection>(
-  rng: &mut R,
-  rpc: &Rpc<RPC>,
-  ring_len: usize,
+fn prepare_inputs(
   inputs: &[SpendableOutput],
```
kayabaNerve: Why not `(SpendableOutput, Decoys)`?

coins/monero/src/wallet/send/mod.rs
```diff
@@ -197,6 +184,55 @@ async fn prepare_inputs<R: RngCore + CryptoRng, RPC: RpcConnection>(
...
+fn calculate_weight_and_fee(
+  protocol: Protocol,
+  decoy_weights: &Vec<usize>,
+  n_inputs: usize,
+  n_outputs: usize,
+  extra: usize,
+  fee_rate: Fee,
+) -> (usize, u64) {
+  let mut weight: usize = 0;
+  let mut fee: u64 = 0;
```
kayabaNerve: FYI, 0usize, 0u64 are valid Rust.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -197,6 +184,55 @@ async fn prepare_inputs<R: RngCore + CryptoRng, RPC: RpcConnection>(
...
+    weight = Transaction::fee_weight(protocol, decoy_weights, n_inputs, n_outputs, extra, fee);
+
+    let fee_calculated_from_weight = fee_rate.calculate_fee_from_weight(weight);
+
+    // Continue trying to use the fee calculated from the tx's weight
+    done = fee_calculated_from_weight == fee;
+
+    fee = fee_calculated_from_weight;
+
+    debug_assert!(iters < max_iters, "Reached max fee calculation attempts");
```
kayabaNerve, line 233: This isn't an actual invariant and shouldn't be a debug assert.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -235,9 +278,11 @@ pub struct SignableTransaction {
   protocol: Protocol,
   r_seed: Option<Zeroizing<[u8; 32]>>,
   inputs: Vec<SpendableOutput>,
+  decoys: Vec<Decoys>,
```
kayabaNerve: Would it be better to do (SpendableOutput, Decoys) in a single vec? This is internal so I don't actually care, just noting.

processor/src/coins/monero.rs
```diff
@@ -536,7 +549,7 @@ impl Coin for Monero {
 
   #[cfg(test)]
   async fn get_fee(&self) -> Self::Fee {
-    self.rpc.get_fee().await.unwrap()
+    self.rpc.get_fee(0).await.unwrap()
```
kayabaNerve: Is 0 highest or lowest priority? Lowest, right?

coins/monero/src/wallet/send/mod.rs
```diff
@@ -276,6 +321,49 @@ impl Change {
...
+  }
+
+  (subaddresses, additional)
+}
+
+fn sanity_check_change_payment(payments: &[InternalPayment], has_change_address: bool) {
+  assert_eq!(
+    payments
+      .iter()
+      .filter(|payment| match *payment {
```
kayabaNerve, line 389: You can use `any` to return if any payments in the iter are change. Just FYI.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -332,56 +422,89 @@ impl SignableTransaction {
...
+    // Modify the amount of the change output
+    if change_address.is_some() {
+      let change_payment = payments.last_mut().unwrap();
+      *change_payment =
+        InternalPayment::Change(change_address.clone().unwrap(), in_amount - out_amount - fee);
     }
 
-    Ok(SignableTransaction { protocol, r_seed, inputs, payments, data, fee })
+    // Sanity check the change again after modifying
+    sanity_check_change_payment(&payments, change_address.is_some());
```
kayabaNerve, line 527: What's the concern here?

coins/monero/src/wallet/send/multisig.rs
```diff
@@ -153,22 +145,7 @@ impl SignableTransaction {
...
-      // committed to. They'll also be committed to later via the TX message as a whole
-      &mut ChaCha20Rng::from_seed(transcript.rng_seed(b"decoys")),
-      rpc,
-      self.protocol.ring_len(),
-      height,
-      &self.inputs,
-    )
-    .await
-    .map_err(TransactionError::RpcError)?;
+    let decoys = self.decoys.clone();
```
kayabaNerve: You shouldn't clone decoys out of self when self is in the Machine. Just access them via `.signable`.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -332,56 +422,89 @@ impl SignableTransaction {
...
+    // Get the outgoing amount ignoring fees
+    let out_amount = payments.iter().map(|payment| payment.1).sum::<u64>();
+
     let outputs = payments.len() + usize::from(change_address.is_some());
+    if outputs > MAX_OUTPUTS {
+      Err(TransactionError::TooManyOutputs)?;
+    }
+
+    // Collect payments in a container that includes a change output if a change address is provided
+    let mut payments: Vec<InternalPayment> =
```
kayabaNerve: I don't believe this type specification is needed.

coins/monero/tests/runner.rs
```diff
@@ -74,6 +75,18 @@ pub async fn get_miner_tx_output(rpc: &Rpc<HttpRpc>, view: &ViewPair) -> Spendab
   scanner.scan(rpc, &block).await.unwrap().swap_remove(0).ignore_timelock().swap_remove(0)
 }
 
+/// Makes sure the weight and fee match the expected calculation
+pub fn check_weight_and_fee(tx: &Transaction, fee_rate: Fee) {
```
kayabaNerve, line 77: This should also be run on TXs sent via monero-wallet-rpc, which we should have in lib.

kayabaNerve (commented):
utACK. Overall LGTM.

coins/monero/src/rpc.rs
j-berman: Fair.

coins/monero/src/rpc.rs
j-berman: I was trying to match wallet2's behavior where if the RPC response is bad, it returns these defaults. I'm fine with erroring here instead. Easier to reason through and it's a sensible error case.

coins/monero/src/rpc.rs
j-berman: I'm fine with erroring here instead as noted above

coins/monero/src/transaction.rs
j-berman: > Why take in the variable twice if it just introduces a potential inconsistency?

Because I didn't think of this here, good spot :)

coins/monero/src/wallet/send/mod.rs
j-berman: This is there to pick up an unexpected result. It should never happen unless there is a mistake in the code. Plus I figure an actual invariant should be an assert. What is `debug_assert` for if not for a situation like this?

coins/monero/src/wallet/send/mod.rs
j-berman: I like this suggestion. Done in the latest.

processor/src/coins/monero.rs
j-berman: 0 is lowest. FWIW, upon working through the v17 fixes, I noticed the CLI, RPC, and GUI all end up defaulting this to 1 in the normal case thanks to a separate code path in those wallets that adjusts the priority down to 1 if there is no backlog in the pool. I defaulted to 1 in the latest and included a TODO to implement an `adjust_priority`

coins/monero/src/wallet/send/mod.rs
j-berman: Think you're aware, but just clarifying, it's testing that the count is exactly as expected here

coins/monero/src/wallet/send/mod.rs
j-berman: That someone will accidentally break this invariant when modifying the code in the future. I could make the code cleaner to avoid any value in adding this assert here, but figured this is solid to ship and it's worth me moving on

coins/monero/tests/runner.rs
j-berman: See 4905e6fd3c2eaf6a2264bab2ce00ea2b7e1b8790

If https://github.com/monero-project/monero/pull/8882 is merged then that commit could be reverted

j-berman:
- In revisions, I noticed a missing parameter to the `get_fee_estimate` RPC that impacts the fee returned when tx volume fluctuates and dynamic fees trigger (the `grace_blocks` parameter). In the latest, the default RPC  request now matches wallet2 as of the latest. FWIW fairly positive I would have caught this when testing this PR with mainnet and am still planning to test with mainnet. Can look into implementing a test to trigger this scenario too.
- Ack all your comments I didn't explicitly respond to. I implemented those comments in the latest.
- v17 tests should be working now.
- Also calling out that there is an `adjust_priority` function that executes by default in the CLI, RPC, and GUI wallets that automatically adjusts the priority of a user-provided priority downward if there is no backlog in the transaction pool (and the user-provided priority would cause an unnecessarily expensive fee). I added a TODO for a future PR.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -622,7 +721,7 @@ impl SignableTransaction {
       &mut self.data,
     );
 
-    let mut fee = self.inputs.iter().map(|input| input.commitment().amount).sum::<u64>();
+    let mut fee = self.inputs.iter().map(|(input, _)| input.commitment().amount).sum::<u64>();
     let mut tx_outputs = Vec::with_capacity(outputs.len());
     let mut ecdh_info = Vec::with_capacity(outputs.len());
     for output in &outputs {
```
kayabaNerve, line 769: L633 in develop, 732 in your PR at time of comment is

`Some(output.view_tag).filter(|_| matches!(self.protocol, Protocol::v16))`

This should be consolidated with Protocol::view_tags.

coins/monero/src/lib.rs
```diff
@@ -67,15 +67,36 @@ impl Protocol {
...
+    match self {
+      Protocol::v14 => false,
+      Protocol::v16 => true,
+      Protocol::Custom { view_tags, .. } => *view_tags,
+    }
+  }
+
+  /// Whether or not the specified version uses the fee algorithm released in
+  /// Monero v18.
+  pub fn v18_fee(&self) -> bool {
```
kayabaNerve: In the future, it may make sense for the fee algorithm to be an enum. Since we don't plan to support legacy fee algorithms though, I'm not asking for it here.

coins/monero/src/lib.rs
```diff
@@ -67,15 +67,36 @@ impl Protocol {
...
+    match self {
+      Protocol::v14 => false,
+      Protocol::v16 => true,
```
kayabaNerve, line 111: Why do you have Protocol::v16 using the v18 fee algorithm?

coins/monero/src/ringct/bulletproofs/mod.rs
```diff
@@ -35,23 +35,38 @@ pub enum Bulletproofs {
...
-    let len = (fields + (2 * LR_len)) * 32;
-    len +
-      if padded_outputs <= 2 {
-        0
-      } else {
-        let base = ((fields + (2 * (LOG_N + 1))) * 32) / 2;
-        let size = (fields + (2 * LR_len)) * 32;
-        ((base * padded_outputs) - size) * 4 / 5
-      }
+    let mut bp_clawback = 0;
```
kayabaNerve, line 60: Cite the BP clawback algorithm with a link to Monero's?

coins/monero/src/rpc.rs
```diff
@@ -18,6 +18,9 @@ use crate::{
   wallet::Fee,
 };
 
+// Number of blocks the fee estimate will be valid for
+const GRACE_BLOCKS_FOR_FEE_ESTIMATE: u64 = 10;
```
kayabaNerve: This should be the same constant as wallet2. Cite source?

coins/monero/src/rpc.rs
```diff
@@ -513,16 +520,69 @@ impl<R: RpcConnection> Rpc<R> {
   /// Get the currently estimated fee from the node. This may be manipulated to unsafe levels and
   /// MUST be sanity checked.
   // TODO: Take a sanity check argument
-  pub async fn get_fee(&self) -> Result<Fee, RpcError> {
+  pub async fn get_fee(&self, protocol: Protocol, priority: u32) -> Result<Fee, RpcError> {
```
kayabaNerve: Should priority be a u32 or an enum? Match Protocol and do enum with Custom(u32)?

Also, just double checking it's u32 and not i32.

coins/monero/src/rpc.rs
```diff
@@ -513,16 +520,69 @@ impl<R: RpcConnection> Rpc<R> {
...
+
+      let res: FeeResponse = self
+        .json_rpc_call(
+          "get_fee_estimate",
+          Some(json!({ "grace_blocks": GRACE_BLOCKS_FOR_FEE_ESTIMATE })),
+        )
+        .await?;
+
+      let priority_idx =
+        if priority >= 4 { 3 } else { priority.saturating_sub(1).try_into().unwrap() };
```
kayabaNerve: Error InvalidPriority?

coins/monero/src/rpc.rs
```diff
@@ -513,16 +520,69 @@ impl<R: RpcConnection> Rpc<R> {
...
+
+      if res.status != "OK" {
+        Err(RpcError::InvalidFee)
+      } else if priority_idx >= res.fees.len() {
```
kayabaNerve: `res.fees.get(priority.try_into().unwrap())`, moving the `.try_into().unwrap()` off saturating sub for clarity, then `.ok_or(RpcError::InvalidPriority)?` for Option to Result.

coins/monero/src/serialize.rs
```diff
@@ -7,8 +7,18 @@ use curve25519_dalek::{
 
 const VARINT_CONTINUATION_MASK: u8 = 0b1000_0000;
 
-pub(crate) fn varint_len(varint: usize) -> usize {
-  ((usize::try_from(usize::BITS - varint.leading_zeros()).unwrap().saturating_sub(1)) / 7) + 1
+mod sealed {
+  pub trait Varint: TryInto<u64> {}
```
kayabaNerve: VarInt

coins/monero/src/rpc.rs
```diff
@@ -513,16 +520,69 @@ impl<R: RpcConnection> Rpc<R> {
...
-    struct FeeResponse {
+    struct FeeResponseV17 {
+      status: String,
       fee: u64,
       quantization_mask: u64,
     }
 
-    let res: FeeResponse = self.json_rpc_call("get_fee_estimate", None).await?;
-    Ok(Fee { per_weight: res.fee, mask: res.quantization_mask })
+    let priority_idx = if priority == 0 { 1 } else { priority - 1 };
```
kayabaNerve: So 0 -> 1, and 2 -> 1, yet 1 -> 0? Please document this monstrosity better, ideally citing Monero RPC docs yet alternatively the code it uses for this.

coins/monero/src/rpc.rs
```diff
@@ -513,16 +520,69 @@ impl<R: RpcConnection> Rpc<R> {
...
+    let multipliers: Vec<u64> = vec![1, 5, 25, 1000];
+    if priority_idx >= multipliers.len().try_into().unwrap() {
+      // though not an RPC error, it seems sensible to treat as such
+      Err(RpcError::InvalidPriority)?;
+    }
+    let fee_multiplier = multipliers[priority_idx as usize];
```
kayabaNerve: try_into, please.

coins/monero/src/transaction.rs
```diff
@@ -331,4 +339,26 @@ impl Transaction {
 
     hash(&sig_hash)
   }
+
+  fn is_rct_bulletproof(&self) -> bool {
+    matches!(self.rct_signatures.prunable.rct_type(), 3 | 4 | 5)
```
kayabaNerve: Can we match prunable, returning true/false for each possible enum variant?

coins/monero/src/transaction.rs
```diff
@@ -331,4 +339,26 @@ impl Transaction {
...
+  }
+
+  fn is_rct_bulletproof_plus(&self) -> bool {
+    matches!(self.rct_signatures.prunable.rct_type(), 6)
```
kayabaNerve: Same comment as is_rct_bulletproof.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -33,7 +33,7 @@ use crate::{
     RctBase, RctPrunable, RctSignatures,
   },
   transaction::{Input, Output, Timelock, TransactionPrefix, Transaction},
-  rpc::{RpcError, RpcConnection, Rpc},
+  rpc::{RpcError},
```
kayabaNerve: Unnecessary curly bois

coins/monero/src/wallet/send/mod.rs
kayabaNerve: Agreed it should never happen. Statement as an invariant requires effective certainty this won't be hit, no matter the arguments/variables at play. We allow custom fee rate specification, and that introduces a massive surface scope.

Unlikely != impossible without error.

You can likely turn this into an invariant by stating that the fee will only go up, and will stop going up once the VarInt hits its maximum length in bytes. Additionally, any increases in the fee requires the VarInt having gone up in bytes, and accordingly, the maximum execution depth is 10 (preventing endless execution). But then the max_iters has to be the execution depth which it should already halt on, making this debug_assert pointless.

The reasons to remove this debug_assert, and unbound iterations which may allow ridiculous values, are:

1) So we can keep other debug_asserts
2) Because the fee already has to be inspected before signing continues

If the concern is programs not doing #2, we should emphasize the API around doing so like Timelock.

As for this specific debug_assert, you can rewrite as
```
#[cfg(test)]
debug_assert
```
So it's caught in testing yet will never effect actual programs, compiled with debug_asserts or not.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -276,6 +318,49 @@ impl Change {
...
+  // Unless the above change view key path is taken
+  if (payments.len() == 2) && has_change_view {
+    additional = false;
+  }
+
+  (subaddresses, additional)
+}
+
+fn sanity_check_change_payment(payments: &[InternalPayment], has_change_address: bool) {
+  assert_eq!(
```
kayabaNerve: debug_assert unless you're concerned over an API invariant.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -332,62 +418,98 @@ impl SignableTransaction {
...
-    // Calculate the fee.
-    let fee = fee_rate.calculate(estimated_tx_size);
-
     // Make sure we have enough funds
-    let in_amount = inputs.iter().map(|input| input.commitment().amount).sum::<u64>();
-    let out_amount = payments.iter().map(|payment| payment.1).sum::<u64>() + fee;
-    if in_amount < out_amount {
-      Err(TransactionError::NotEnoughFunds(in_amount, out_amount))?;
+    let in_amount = inputs.iter().map(|(input, _)| input.commitment().amount).sum::<u64>();
+    if in_amount < out_amount + fee {
```
kayabaNerve: Parens on rhs, please.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -332,62 +418,98 @@ impl SignableTransaction {
...
+      *change_payment =
+        InternalPayment::Change(change_address.clone().unwrap(), in_amount - out_amount - fee);
     }
 
-    Ok(SignableTransaction { protocol, r_seed, inputs, payments, data, fee })
+    // Sanity check the change again after modifying
+    sanity_check_change_payment(&payments, change_address.is_some());
+
+    // Sanity check outgoing amount + fee == incoming amount
+    assert_eq!(
```
kayabaNerve: debug_assert?

coins/monero/src/wallet/send/mod.rs
```diff
@@ -419,31 +541,8 @@ impl SignableTransaction {
...
-
-    // We need additional keys if we have any subaddresses
-    let mut additional = subaddresses;
-    // Unless the above change view key path is taken
-    if (payments.len() == 2) && has_change_view {
-      additional = false;
-    }
-    let modified_change_ecdh = subaddresses && (!additional);
+    let (subaddresses, additional) = need_additional(payments);
+    let modified_change_ecdh = subaddresses && !additional;
```
kayabaNerve: Parens on right term, please.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -286,10 +371,11 @@ impl SignableTransaction {
   /// specified, leftover funds will be sent to it.
   ///
   /// Each chunk of data must not exceed MAX_ARBITRARY_DATA_SIZE and will be embedded in TX extra.
+  #[allow(clippy::too_many_arguments)]
   pub fn new(
     protocol: Protocol,
     r_seed: Option<Zeroizing<[u8; 32]>>,
-    inputs: Vec<SpendableOutput>,
+    inputs: Vec<(SpendableOutput, Decoys)>,
```
kayabaNerve, line 412: These decoys need to be checked to have the proper ring_len.

coins/monero/src/wallet/send/multisig.rs
```diff
@@ -148,36 +134,17 @@ impl SignableTransaction {
       let clsag = ClsagMultisig::new(transcript.clone(), input.key(), inputs[i].clone());
       key_images.push((
         clsag.H,
-        keys.current_offset().unwrap_or(dfg::Scalar::ZERO).0 + self.inputs[i].key_offset(),
+        keys.current_offset().unwrap_or(dfg::Scalar::ZERO).0 + self.inputs[i].0.key_offset(),
       ));
       clsags.push(AlgorithmMachine::new(clsag, offset));
     }
 
-    // Select decoys
```
kayabaNerve, line 157: Transcript the decoys since they're no longer deterministic to the transcript, please.

coins/monero/tests/runner.rs
kayabaNerve: Add a TODO in codebase for when 8882 is merged?

kayabaNerve: Add a TODO in codebase for when 8882 is merged?

coins/monero/tests/send.rs
```diff
@@ -116,3 +148,126 @@ test!(
...
+        );
+        subaddresses.push(subaddress);
+      }
+
+      (builder.build().unwrap(), (scanner, subaddresses))
+    },
+    |_, tx: Transaction, _, mut state: (Scanner, Vec<SubaddressIndex>)| async move {
+      let mut scanned_tx = state.0.scan_transaction(&tx).not_locked();
+
+      use std::collections::HashMap;
```
kayabaNerve: Import at the top?

processor/src/coins/monero.rs
```diff
@@ -398,6 +399,28 @@ impl Coin for Monero {
...
+      protocol.ring_len(),
+      block_number + 1,
+      &spendable_outputs,
+    )
+    .await
+    .map_err(|_| CoinError::ConnectionError)
+    .unwrap();
+
+    let inputs: Vec<(SpendableOutput, Decoys)> =
+      spendable_outputs.into_iter().zip(decoys.into_iter()).collect();
```
kayabaNerve: `collect::<Vec<_>>();`

processor/src/coins/monero.rs
```diff
@@ -583,17 +598,32 @@ impl Coin for Monero {
...
+      &self.rpc,
+      protocol.ring_len(),
+      self.rpc.get_height().await.unwrap() - 1,
+      &outputs,
+    )
+    .await
+    .unwrap();
+
+    let inputs: Vec<(SpendableOutput, Decoys)> =
+      outputs.into_iter().zip(decoys.into_iter()).collect();
```
kayabaNerve: `collect::<Vec<_>>();`

kayabaNerve (commented):
Very thankful to have this :D

kayabaNerve:
CI failure due to #291.

coins/monero/src/wallet/send/mod.rs
j-berman: There was another one later in the file too, updated both.

coins/monero/src/lib.rs
j-berman: I was referring to Monero release version v18, but agreed this was confusing. Renamed the fn to `v16_fee` for consistency.

coins/monero/src/rpc.rs
j-berman: Implemented an enum.

It's u32: https://github.com/monero-project/monero/blob/94e67bf96bbc010241f29ada6abc89f49a81759c/src/wallet/wallet2.cpp#L7610

coins/monero/src/rpc.rs
j-berman: Leaving this one for you

coins/monero/src/rpc.rs
j-berman: Cited the code.

coins/monero/src/transaction.rs
j-berman: Used the nice new `rct_signatures.rct_type()` in the latest.

coins/monero/tests/send.rs
j-berman: I moved to the top of that function in the latest (since top of the file gives unused warnings). How's that? Or is the preference to suppress the warning?

kayabaNerve: Top of block is good.

PR #308, opened 2023-06-23
Monero: support for legacy transactions

### LINKED ISSUE #304: Monero V1 txs serialization issues
V1 miner txs don't have any signatures and (for non-miner V1 txs) signatures should be  `Vec<Vec<(Scalar, Scalar)>>` not `Vec<(Scalar, Scalar)>`. I have fixed this [here](https://github.com/Cuprate/serai/commit/d8b8ce9837eb1372d167d60ed8f758ed5480854f) which I will upstream with the rest of the legacy types just thought I would make this known. 

I have also began scanning the blockchain and, with that fix, all V1 txs/ blocks can be decoded.

### LINKED ISSUE #306: We need to support deserializing all v2 Monero transactions
This is due to decoy selection deserializing whatever transactions selected as decoys. Currently, we're bugged there.

Reported by @j-berman.

Boog900:
(should) add support for all txs types.

Fixes #304 
Fixes #306

With verifying functionality for Borromean sigs and MLSAG, I haven't done proving because no one using Monero on main net will need it.

I have also included a function to get the transaction header hash, and to compute the block hash. We discussed ways of doing POW hashes but I feel just exposing serialize_hashable is the best option. Also `merkle_root.rs` has been ripped from monero-rs so I will need to add a copyright header to that file.

I will continue scanning the blockchain to see if we can (de)serialize all blockchain data for now I will mark this as a draft.


kayabaNerve:
Thanks! I'll review it when I have a few moments to try and help there.

coins/monero/generators/src/lib.rs
```diff
@@ -30,6 +32,21 @@ lazy_static! {
...
+  pub static ref H2: [DalekPoint; 64] = generate_H2();
+}
+
+#[allow(non_snake_case)]
+fn generate_H2() -> [DalekPoint; 64] {
+  let mut temp = Vec::with_capacity(64);
+  for i in 0 .. 64 {
+    temp.push(Scalar::from(2_u128.pow(i)) * *H)
+  }
+  temp.try_into().unwrap()
```
kayabaNerve: Why does this use a vector when you can just index with `[i]`?

coins/monero/generators/src/lib.rs
```diff
@@ -30,6 +32,21 @@ lazy_static! {
       .decompress()
       .unwrap()
       .mul_by_cofactor();
+
+  /// Monero's `H` generator multiplied 2^i for each index, i.e. H, 2H, 4H, 8H, ...
+  /// used in old range proofs.
+  /// https://github.com/monero-project/monero/blob/94e67bf96bbc010241f29ada6abc89f49a81759c/src/
+  /// ringct/rctTypes.h#L628
+  pub static ref H2: [DalekPoint; 64] = generate_H2();
```
kayabaNerve: I'm unsure how much I appreciate this being here, as it's decently expensive to generate. I'm unsure 64 decompressions would be cheaper than 64 doubles however. I doubt it, so I guess there's  nothing else to do.

coins/monero/generators/src/lib.rs
```diff
@@ -30,6 +32,21 @@ lazy_static! {
...
+}
+
+#[allow(non_snake_case)]
+fn generate_H2() -> [DalekPoint; 64] {
+  let mut temp = Vec::with_capacity(64);
+  for i in 0 .. 64 {
+    temp.push(Scalar::from(2_u128.pow(i)) * *H)
```
kayabaNerve: This can be much more efficiently calculated with doublings.

coins/monero/src/block.rs
```diff
@@ -65,6 +68,29 @@ impl Block {
...
+
+    let mut out = vec![];
+    write_varint(&(blob.len() as u64), &mut out).unwrap();
+    out.append(&mut blob);
+
+    out
+  }
+
+  pub fn id(&self) -> [u8; 32] {
+    // TODO: block 202612?
```
kayabaNerve: Why is this block significant? Or is that why it has its own question mark (since we don't know why)?

coins/monero/src/block.rs
```diff
@@ -65,6 +68,29 @@ impl Block {
...
+  }
+
+  pub fn serialize_hashable(&self) -> Vec<u8> {
+    let mut blob = self.header.serialize();
+
+    blob.extend_from_slice(&self.tx_merkle_root());
+
+    write_varint(&(1 + self.txs.len() as u64), &mut blob).unwrap();
+
+    let mut out = vec![];
```
kayabaNerve: Vec::with_capacity since blob.len is known.

coins/monero/src/block.rs
```diff
@@ -65,6 +68,29 @@ impl Block {
...
+  pub fn tx_merkle_root(&self) -> [u8; 32] {
+    merkle_root::tree_hash(self.miner_tx.hash(), &self.txs)
+  }
+
+  pub fn serialize_hashable(&self) -> Vec<u8> {
+    let mut blob = self.header.serialize();
+
+    blob.extend_from_slice(&self.tx_merkle_root());
+
+    write_varint(&(1 + self.txs.len() as u64), &mut blob).unwrap();
```
kayabaNerve: `as` is unsafe. This would be u64::try_from, yet I thought we impl'd usize as an acceptable VarInt...

coins/monero/src/block.rs
```diff
@@ -65,6 +68,29 @@ impl Block {
...
+
+    let mut out = vec![];
+    write_varint(&(blob.len() as u64), &mut out).unwrap();
```
kayabaNerve: Use of as.

coins/monero/src/block/merkle_root.rs
```diff
@@ -0,0 +1,68 @@
+use crate::hash;
```
kayabaNerve: Is this the file from monero-rs? It seems small enough we could trivially rewrite it and not bother complicating the licensing at all.

coins/monero/src/block/merkle_root.rs
```diff
@@ -0,0 +1,68 @@
...
+
+  pow >> 1
+}
+
+fn hash_concat(a: [u8; 32], b: [u8; 32]) -> [u8; 32] {
+  let v = [a, b].concat();
+  hash(&v)
+}
+
+/// Compute tree hash as defined by Cryptonote
```
kayabaNerve: If this isn't simply pairing off, it should document how it's distinct. If it is simply pairing off, it does so in a very ugly way.

I believe there's a MRL PDF on the algorithm since it was exploited...

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
+#![allow(non_snake_case)]
+
+use std::fmt::Debug;
```
kayabaNerve: Personally, I prefer merged inputs consistently grouped, but I ack if this hasn't been the convention on your end and can do it after.

coins/monero/generators/src/lib.rs
kayabaNerve: Also, confusingly named. Is this a second H or the powers of H (`H_pow_2`)? We already have two `H` generators, the static amount blinding `H` and the variable key image `H`.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+use curve25519_dalek::traits::Identity;
+
+use monero_generators::H2;
+
+use crate::hash_to_scalar;
+use crate::serialize::*;
+
+/// A Borromean signature.
+///
+/// Note: This type keeps the data as raw bytes as Monero has
```
kayabaNerve: What length were these comments kept to? 80?

Also, the "Note:" label is redundant.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+/// some transactions with unreduced scalars in this field, we
+/// could use `from_bytes_mod_order` but then we would not be able
+/// to encode this back into it's original form.
```
kayabaNerve: Personally, I'd introduce a new scalar type entirely. UncheckedScalar. Then that would have methods for its reduced form, and it's invalidly reduced form (which dangerousfreedom identified).

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+///
```
kayabaNerve, line 19: Pointless comment line.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+  }
+  pub fn write<W: Write>(&self, w: &mut W) -> io::Result<()> {
+    self.asig.write(w)?;
+    write_raw_vec(write_point, &self.Ci, w)
+  }
+
+  pub fn verify(&self, commitment: &EdwardsPoint) -> bool {
+    let mut P1 = Vec::with_capacity(64);
+    let mut P2 = Vec::with_capacity(64);
+    let mut bbs0 = Vec::with_capacity(64);
```
kayabaNerve: Not to say this doesn't make sense to someone who's worked with the paper/Monero's impl, yet mind sharing with me what "bb" is?

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+    let mut bbs1 = Vec::with_capacity(64);
+
+    let bbee = Scalar::from_bytes_mod_order(self.asig.ee);
+
+    let mut C_temp = EdwardsPoint::identity();
+
+    for i in 0 .. 64 {
+      bbs0.push(Scalar::from_bytes_mod_order(self.asig.s0[i]));
```
kayabaNerve: I don't believe this is correct, per dangerousfreedom. At *some* point, one of these will bite you.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+      P1.push(self.Ci[i]);
+      P2.push(P1[i] - H2[i]);
+
+      C_temp += P1[i];
+    }
+
+    if &C_temp != commitment {
+      false
+    } else {
+      verify_borromean(P1, P2, bbee, bbs0, bbs1)
```
kayabaNerve: P1 == self.Ci. Just pass self.Ci.clone().

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+    }
+  }
+}
+
+fn verify_borromean(
+  P1: Vec<EdwardsPoint>,
+  P2: Vec<EdwardsPoint>,
+  bbee: Scalar,
+  bbs0: Vec<Scalar>,
+  bbs1: Vec<Scalar>,
```
kayabaNerve: AFAICT, &[]  would work instead of all of these vectors.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    if &C_temp != commitment {
+      false
+    } else {
+      verify_borromean(P1, P2, bbee, bbs0, bbs1)
+    }
+  }
+}
+
+fn verify_borromean(
```
kayabaNerve: I'll have to double-check consistency with Monero before I sign off on this.

coins/monero/src/ringct/borromean.rs
kayabaNerve: Also, shouldn't you early return and then inline verify_borromean? Why break it out? This *is* Borromean::verify.

coins/monero/src/ringct/borromean.rs
```diff
@@ -0,0 +1,108 @@
...
+  P1: Vec<EdwardsPoint>,
+  P2: Vec<EdwardsPoint>,
+  bbee: Scalar,
+  bbs0: Vec<Scalar>,
+  bbs1: Vec<Scalar>,
+) -> bool {
+  let mut LV: Vec<u8> = Vec::with_capacity(2048);
```
kayabaNerve: I don't think this type specification is necessary.

coins/monero/src/ringct/mlsag/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+
+  pub fn verify_rct_full(
+    &self,
+    msg: &[u8; 32],
+    pubs: &[[EdwardsPoint; 2]],
+    out_pks: &[EdwardsPoint],
+    fee: &EdwardsPoint,
+    I: &EdwardsPoint,
+  ) -> bool {
+    if pubs.is_empty() {
```
kayabaNerve: This check can be merged into verify_mlsag so it's not duplicated between these functions.

coins/monero/src/ringct/mlsag/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  pub fn read<R: Read>(mixins: usize, r: &mut R) -> io::Result<MgSig> {
+    Ok(MgSig {
+      ss: (0 .. mixins).map(|_| read_array(read_scalar, r)).collect::<Result<_, _>>()?,
+      cc: read_scalar(r)?,
+    })
+  }
+
+  pub fn verify_rct_full(
```
kayabaNerve: These functions should be documented since they're rather opaque as present here.

coins/monero/src/ringct/mlsag/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    &self,
+    msg: &[u8; 32],
+    pubs: &[[EdwardsPoint; 2]],
+    out_pks: &[EdwardsPoint],
+    fee: &EdwardsPoint,
```
kayabaNerve: Is the fee a pedersen commitment or a u64 * G?

coins/monero/src/ringct/mlsag/mod.rs
kayabaNerve: Also, are these actually relevant to MgSig? Shouldn't MgSig only have the underlying function? Not anything related to its RingCT composition?

coins/monero/src/ringct/mlsag/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::io::{Read, Write};
+
+use curve25519_dalek::scalar::Scalar;
+use curve25519_dalek::edwards::EdwardsPoint;
+use curve25519_dalek::traits::Identity;
+
+use crate::{hash_to_scalar, serialize::*};
+use crate::ringct::hash_to_point;
+
+/// MgSig part of MLSAG, as used in Monero.
```
kayabaNerve: This isn't the MgSig part. This IS the MLSAG, per https://eprint.iacr.org/2015/1098.pdf.

coins/monero/src/ringct/mod.rs
```diff
@@ -10,24 +10,54 @@ pub use hash_to_point::{raw_hash_to_point, hash_to_point};
 
 /// CLSAG struct, along with signing and verifying functionality.
 pub mod clsag;
+/// MLSAG struct, along with verifying functionality.
+pub mod mlsag;
+/// RangeSig struct, along with verifying functionality.
```
kayabaNerve: I'd rather identify this as Borromean than RangeSig.

coins/monero/src/ringct/mod.rs
```diff
@@ -10,24 +10,54 @@ pub use hash_to_point::{raw_hash_to_point, hash_to_point};
...
 };
 
 /// Generate a key image for a given key. Defined as `x * hash_to_point(xG)`.
 pub fn generate_key_image(secret: &Zeroizing<Scalar>) -> EdwardsPoint {
   hash_to_point(&ED25519_BASEPOINT_TABLE * secret.deref()) * secret.deref()
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum EcdhInfo {
+  Standard { mask: Scalar, amount: Scalar },
```
kayabaNerve: Sorry, but are these technically scalars or byte arrays?

coins/monero/src/ringct/mlsag/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+    let mut ring_matrix = Vec::with_capacity(pubs.len());
+
+    for member in pubs.iter() {
+      ring_matrix.push([member[0], member[1] - pseudo_out])
+    }
+
+    self.verify_mlsag(msg, &ring_matrix, I)
+  }
+
+  fn verify_mlsag(&self, msg: &[u8; 32], ring: &[[EdwardsPoint; 2]], I: &EdwardsPoint) -> bool {
```
kayabaNerve: I'll have to check if this aligns with Monero.

coins/monero/src/ringct/mod.rs
```diff
@@ -10,24 +10,54 @@ pub use hash_to_point::{raw_hash_to_point, hash_to_point};
...
+  Bulletproof { amount: [u8; 8] },
+}
+
+impl EcdhInfo {
+  pub fn read<R: Read>(rct_type: u8, r: &mut R) -> io::Result<EcdhInfo> {
+    Ok(match rct_type {
+      0 ..= 3 => EcdhInfo::Standard { mask: read_scalar(r)?, amount: read_scalar(r)? },
```
kayabaNerve: If we keep using rct_type like this, they should be turned into proper constants.

I'd rather have this take in a boolean of which to use though, to prevent distribution of such specificities.

coins/monero/src/ringct/mod.rs
```diff
@@ -10,24 +10,54 @@ pub use hash_to_point::{raw_hash_to_point, hash_to_point};
...
 
 /// Generate a key image for a given key. Defined as `x * hash_to_point(xG)`.
 pub fn generate_key_image(secret: &Zeroizing<Scalar>) -> EdwardsPoint {
   hash_to_point(&ED25519_BASEPOINT_TABLE * secret.deref()) * secret.deref()
 }
 
+#[derive(Clone, PartialEq, Eq, Debug)]
+pub enum EcdhInfo {
+  Standard { mask: Scalar, amount: Scalar },
+  Bulletproof { amount: [u8; 8] },
```
kayabaNerve: Is `Bulletproof` the proper label? `Standard` is already bs since they're no longer standard.

coins/monero/src/ringct/mod.rs
```diff
@@ -40,26 +70,31 @@ impl RctBase {
     w.write_all(&[rct_type])?;
     match rct_type {
       0 => Ok(()),
-      5 | 6 => {
+      _ => {
         write_varint(&self.fee, w)?;
+        if rct_type == 2 {
```
kayabaNerve: I have to double check this is correct.

coins/monero/src/ringct/mod.rs
```diff
@@ -40,26 +70,31 @@ impl RctBase {
...
     let rct_type = read_byte(r)?;
     Ok((
       if rct_type == 0 {
-        RctBase { fee: 0, ecdh_info: vec![], commitments: vec![] }
+        RctBase { fee: 0, ecdh_info: vec![], pseudo_outs: vec![], commitments: vec![] }
       } else {
         RctBase {
           fee: read_varint(r)?,
-          ecdh_info: (0 .. outputs).map(|_| read_bytes(r)).collect::<Result<_, _>>()?,
+          pseudo_outs: if rct_type == 2 { read_raw_vec(read_point, inputs, r)? } else { vec![] },
```
kayabaNerve: Time to make constants out of these? Or rather, should RctType should be an enum?

coins/monero/src/ringct/mod.rs
```diff
@@ -71,14 +106,43 @@ impl RctBase {
 #[derive(Clone, PartialEq, Eq, Debug)]
 pub enum RctPrunable {
   Null,
-  Clsag { bulletproofs: Vec<Bulletproofs>, clsags: Vec<Clsag>, pseudo_outs: Vec<EdwardsPoint> },
+  Borromean {
+    range_sigs: Vec<RangeSig>,
+    mlsags: Vec<MgSig>,
+    simple: bool,
+  },
+  BulletProof {
```
kayabaNerve: This isn't valid punctuation. The publication is officially `Bulletproofs`.

coins/monero/src/wallet/mod.rs
```diff
@@ -92,8 +93,24 @@ pub(crate) fn amount_encryption(amount: u64, key: Scalar) -> [u8; 8] {
...
 }
 
-fn amount_decryption(amount: [u8; 8], key: Scalar) -> u64 {
-  u64::from_le_bytes(amount_encryption(u64::from_le_bytes(amount), key))
+fn amount_decryption(amount: &EcdhInfo, key: Scalar) -> u64 {
+  match amount {
+    EcdhInfo::Standard { mask, amount } => {
+      let shared_sec1 = hash(key.as_bytes());
+      let shared_sec2 = hash(&shared_sec1);
+      let _mask_scalar = mask - Scalar::from_bytes_mod_order(shared_sec1);
```
kayabaNerve: This should be  removed.

coins/monero/src/wallet/mod.rs
```diff
@@ -92,8 +93,24 @@ pub(crate) fn amount_encryption(amount: u64, key: Scalar) -> [u8; 8] {
   (amount ^ u64::from_le_bytes(hash(&amount_mask)[.. 8].try_into().unwrap())).to_le_bytes()
 }
 
-fn amount_decryption(amount: [u8; 8], key: Scalar) -> u64 {
-  u64::from_le_bytes(amount_encryption(u64::from_le_bytes(amount), key))
+fn amount_decryption(amount: &EcdhInfo, key: Scalar) -> u64 {
+  match amount {
+    EcdhInfo::Standard { mask, amount } => {
+      let shared_sec1 = hash(key.as_bytes());
+      let shared_sec2 = hash(&shared_sec1);
```
kayabaNerve: If shared_sec1 isn't further used, then just use another `let shared_sec` binding.

coins/monero/src/wallet/mod.rs
```diff
@@ -92,8 +93,24 @@ pub(crate) fn amount_encryption(amount: u64, key: Scalar) -> [u8; 8] {
...
+      let _mask_scalar = mask - Scalar::from_bytes_mod_order(shared_sec1);
+
+      let amount_scalar = amount - Scalar::from_bytes_mod_order(shared_sec2);
+      // get first 64 bits (d2b in rctTypes.cpp)
+      let amount_significant_bytes =
+        amount_scalar.to_bytes()[0 .. 8].try_into().expect("Can't fail");
```
kayabaNerve: I'd just unwrap. An unwrap is saying it can't fail.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -701,6 +703,9 @@ impl SignableTransaction {
         clsags.append(&mut clsag_pairs.iter().map(|clsag| clsag.0.clone()).collect::<Vec<_>>());
         pseudo_outs.append(&mut clsag_pairs.iter().map(|clsag| clsag.1).collect::<Vec<_>>());
       }
+      _ => {
```
kayabaNerve: Should just be a panic. This isn't reachable.

coins/monero/src/wallet/send/mod.rs
```diff
@@ -750,7 +755,8 @@ impl Eventuality {
         view_tag: Some(expected.view_tag).filter(|_| matches!(self.protocol, Protocol::v16)),
       } != actual) ||
         (Some(&expected.commitment.calculate()) != tx.rct_signatures.base.commitments.get(o)) ||
-        (Some(&expected.amount) != tx.rct_signatures.base.ecdh_info.get(o))
+        (Some(&EcdhInfo::Bulletproof { amount: expected.amount }) !=
```
kayabaNerve: I think this is safe, yet I need to check/potentially update documentation.

coins/monero/src/wallet/send/multisig.rs
```diff
@@ -429,6 +429,7 @@ impl SignatureMachine<Transaction> for TransactionSignatureMachine {
           pseudo_outs.push(pseudo_out);
         }
       }
+      _ => todo!(""),
```
kayabaNerve: Should be a panic, not a todo, definitely not a todo which has an empty string.

kayabaNerve (changes requested):
Overall looks to be quality! I really appreciate you working on this, and submitting it back upstream. Feel free to handle these or I can do it on my end (just appreciating the work so far, yet not asking for more).

coins/monero/generators/src/lib.rs
Boog900: I have kept the same name as Monero: 

https://github.com/monero-project/monero/blob/94e67bf96bbc010241f29ada6abc89f49a81759c/src/ringct/rctTypes.h#L628

but I do like `H_pow_2` more ...

coins/monero/src/block.rs
Boog900: https://monero.stackexchange.com/questions/421/what-happened-at-block-202612

Do we handle this inside to lib ... I think we should but I didn't get round to it, it shouldn't be too hard

coins/monero/src/block/merkle_root.rs
Boog900: This is copied from monero-rs which it self was pretty much 1 to 1 copied from monero

coins/monero/src/ringct/borromean.rs
Boog900: This was taken from dangerousfreedom's page: https://www.moneroinflation.com/amounts_confidential_tx

coins/monero/src/ringct/mlsag/mod.rs
Boog900: Cryptography is not my strong point at all but I used this function:
https://github.com/monero-project/monero/blob/0a1eaf26f9dd6b762c2582ee12603b2a4671c735/src/ringct/rctSigs.cpp#L802

coins/monero/src/ringct/mlsag/mod.rs
Boog900: its a u64 * H

coins/monero/src/ringct/mlsag/mod.rs
Boog900: This is how its called in monero-rs

https://github.com/monero-project/monero/blob/0a1eaf26f9dd6b762c2582ee12603b2a4671c735/src/ringct/rctTypes.h#L168

so MgSig is just that + II ?

coins/monero/src/ringct/borromean.rs
Boog900: Yes I think you are right

Boog900: What should we do about the already stored bad signatures? should that be something users deal with?

coins/monero/src/block.rs
Boog900: yep will change that but wouldn't using `as` here be safe as long as there aren't any 128 bit CPUs

Boog900:
Im happy to continue working through this, just don't have a load of time available atm

coins/monero/src/ringct/mod.rs
Boog900: form this:

https://github.com/monero-project/monero/blob/0a1eaf26f9dd6b762c2582ee12603b2a4671c735/src/ringct/rctTypes.h#L140C15-L140C15

it looks like mask is a scalar and amount is a byte array

coins/monero/src/ringct/borromean.rs
Boog900: This is RangeSig::verify. verify_borromean just verifies each bit of the amount is a 1 or 0 whereas this also checks that Ci bits represent C

coins/monero/src/block.rs
kayabaNerve: There are quite a few execution units exceeding 64-bit widths. Rust compatibility for them, and memory which would cause a pointer to exceed 2**64, is a distinct countermeasure to this being an issue.

coins/monero/src/block/merkle_root.rs
kayabaNerve: Right, yet copying Monero line for line is pointless. We should follow its design, not its literalization. Solely being compatible with its literalization off our best work of the design.

coins/monero/src/ringct/borromean.rs
kayabaNerve: > Personally, I'd introduce a new scalar type entirely. UncheckedScalar. Then that would have methods for its reduced form, and it's invalidly reduced form (which dangerousfreedom identified).

The reduction used should be internal to the lib.

coins/monero/src/ringct/borromean.rs
kayabaNerve: Right, but this isn't a RangeSig because a RangeSig doesn't exist. These are Borromean ring signatures used to prove a value is in range. You can delineate Borromean, the academic, out from its usage (keeping the indepent function). I actually would like that. But in that case, you should ideally re-define this range proof as a Vec of Borromean signatures, instead of inlining the signatures' elements.

coins/monero/src/ringct/mlsag/mod.rs
kayabaNerve: Then why not pass in a u64 and do this conversion in-circuit? It's expecting the fee to be so-formed without guaranteeing it.

coins/monero/src/ringct/mlsag/mod.rs
kayabaNerve: Monero notates it MgSig. That doesn't change it is the MLSAG from Monero.

The II is the list of key images. Annoyingly, the IACR paper merged it (part of the statement) with the proof. I'd delineate the MLSAG proof as solely the c scalar and ss matrix, which is what's here, unless I'm missing something.

kayabaNerve:
I may pick it up if it becomes a priority, else I'll leave it to you. Thanks again! :)

kayabaNerve:
Taking over. Will resolve comments as I locally resolve them.

coins/monero/src/ringct/borromean.rs
kayabaNerve: No longer needed thanks to experimental label.

coins/monero/src/ringct/mlsag/mod.rs
kayabaNerve: Not needed since marked experimental.

coins/monero/src/wallet/send/mod.rs
kayabaNerve: Eventuality should likely embed an RctType into itself...

kayabaNerve: We do embed Protocol making this safe so long as Eventuality bounds the protocol.

kayabaNerve:
Full credit to @Boog900.

PR #310, opened 2023-06-28
fix tributary sync test

### LINKED ISSUE #291: Tributary sync test intermittently fails

akildemir:
fix https://github.com/serai-dex/serai/issues/291

### DIFF
--- a/coordinator/src/tests/tributary/sync.rs
+++ b/coordinator/src/tests/tributary/sync.rs
@@ -35,10 +35,11 @@ async fn sync_test() {
 
   // Have the rest form a P2P net
   let mut tributary_arcs = vec![];
+  let mut p2p_threads = vec![];
   for (i, (p2p, tributary)) in tributaries.drain(..).enumerate() {
     let tributary = Arc::new(RwLock::new(tributary));
     tributary_arcs.push(tributary.clone());
-    tokio::spawn(handle_p2p(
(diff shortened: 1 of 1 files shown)


coordinator/src/tests/tributary/sync.rs
```diff
@@ -112,21 +114,28 @@ async fn sync_test() {
...
-    {
-      return;
-    }
-    sleep(Duration::from_secs(block_time)).await;
+  // Because only `t` validators are used in a commit, take n - t nodes offline
+  // leaving only `t` nodes. Which should force it to participate in the consensus
+  // of next blocks.
+  let n = (spec.n() - spec.t()) as usize;
+  for t in p2p_threads.iter().take(n) {
+    t.abort();
```
kayabaNerve, line 122: I'd prefer
```
  let spares = usize::from(spec.n() - spec.t());
  for thread in p2p_threads.iter().take(spares) {
    thread.abort();
```

`as` violates safety, and then you call this `n` when it isn't `n`. It's `n-t`. Then the usage of a variable, `t`, for a thread when `t` already exists is in conflict. But it's not worth blocking the merge for if tests pass.

kayabaNerve:
Even if this doesn't consistently work, it is a tighter test and appreciated accordingly.

PR #314, opened 2023-07-02
std-shims: fix `Read` for &[u8]

Boog900:


kayabaNerve:
That's embarrassing on my end. Thanks.

PR #316, opened 2023-07-06
Improve batch handling

### LINKED ISSUE #280: DoS by too big batches
A malicious ETH validator can create a block flooding Serai with InInstructions. If the encoded size of the SignedBatch exceeds the block size, the batch will fail to be published. This will effectively halt all activity on ETH until a hard fork occurs.

If a SignedBatch exceeds some size, we need to split it into multiple batches.

akildemir:
fixes https://github.com/serai-dex/serai/issues/280 alongside some improvements

- restrict batch sizes to ~25kb
- rate limit batches to 1 batch per serai block per key group
- allow multiple batches per block

### DIFF
--- a/coordinator/src/tributary/scanner.rs
+++ b/coordinator/src/tributary/scanner.rs
@@ -496,7 +496,7 @@ async fn handle_block<
                 spec.set().network,
                 CoordinatorMessage::Coordinator(
                   coordinator::CoordinatorMessage::BatchPreprocesses {
-                    id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
+                    id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
                     preprocesses,
                   },
                 ),
@@ -519,7 +519,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Coordinator(coordinator::CoordinatorMessage::BatchShares {
-                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
                   shares: shares
                     .drain()
                     .map(|(validator, share)| (validator, share.try_into().unwrap()))
@@ -545,7 +545,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Sign(sign::CoordinatorMessage::Preprocesses {
-                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
                   preprocesses,
                 }),
               )
@@ -567,7 +567,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Sign(sign::CoordinatorMessage::Shares {
-                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
                   shares,
                 }),
               )
--- a/coordinator/src/tributary/scanner.rs
+++ b/coordinator/src/tributary/scanner.rs
@@ -496,7 +496,7 @@ async fn handle_block<
                 spec.set().network,
                 CoordinatorMessage::Coordinator(
                   coordinator::CoordinatorMessage::BatchPreprocesses {
-                    id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
+                    id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
                     preprocesses,
                   },
                 ),
@@ -519,7 +519,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Coordinator(coordinator::CoordinatorMessage::BatchShares {
-                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
                   shares: shares
                     .drain()
                     .map(|(validator, share)| (validator, share.try_into().unwrap()))
@@ -545,7 +545,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Sign(sign::CoordinatorMessage::Preprocesses {
-                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
                   preprocesses,
                 }),
               )
@@ -567,7 +567,7 @@ async fn handle_block<
               .send(
                 spec.set().network,
                 CoordinatorMessage::Sign(sign::CoordinatorMessage::Shares {
-                  id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
+                  id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
                   shares,
                 }),
               )
--- a/processor/messages/src/lib.rs
+++ b/processor/messages/src/lib.rs
@@ -14,8 +14,12 @@ use validator_sets_primitives::{ValidatorSet, KeyPair};
 #[derive(Clone, PartialEq, Eq, Debug, Zeroize, Serialize, Deserialize)]
 pub struct SubstrateContext {
   pub serai_time: u64,
+<<<<<<< HEAD
   pub network_latest_finalized_block: BlockHash,
   pub batches: Vec<u32>,
+=======
+  pub coin_latest_finalized_block: BlockHash,
+>>>>>>> 5a4d39d2 (fix review comments)
 }
 
 pub mod key_gen {
@@ -158,6 +162,7 @@ pub mod substrate {
       block: u64,
       key: Vec<u8>,
       burns: Vec<OutInstructionWithBalance>,
+      batches: Vec<u32>,
     },
   }
 
--- a/processor/messages/src/lib.rs
+++ b/processor/messages/src/lib.rs
@@ -11,10 +11,11 @@ use in_instructions_primitives::SignedBatch;
 use tokens_primitives::OutInstructionWithBalance;
 use validator_sets_primitives::{ValidatorSet, KeyPair};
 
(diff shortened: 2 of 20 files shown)


substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -22,6 +22,8 @@ use serai_primitives::{BlockHash, Balance, NetworkId, SeraiAddress, ExternalAddr
 mod shorthand;
 pub use shorthand::*;
 
+pub const MAX_BATCH_SIZE: usize = 25_000_000; // ~25kb
```
kayabaNerve: This is 25 MB, not 25 KB.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -82,6 +90,15 @@ pub mod pallet {
 
       let mut batch = batch.batch;
 
+      // check that this validator set already included a batch for this block
+      let validator_set = ValidatorSet { session: Session(0), network: batch.network };
+      let current_block = <frame_system::Pallet<T>>::block_number();
+      let last_block = LastBatchBlock::<T>::get(validator_set).unwrap_or(0);
+      if last_block >= current_block {
+        Err(DispatchError::Other("only 1 batch per set allowed per serai block"))?;
+      }
```
kayabaNerve: I believe this check should be in validate_unsigned and return ::Future if it fails. That'll keep it in the mempool yet not include it in blocks yet.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -57,6 +57,14 @@ pub mod pallet {
...
   pub(crate) type Batches<T: Config> = StorageMap<_, Blake2_256, NetworkId, u32, OptionQuery>;
 
+  // The last serai block that this validator set included a batch
+  #[pallet::storage]
+  #[pallet::getter(fn last_batch_block)]
+  pub(crate) type LastBatchBlock<T: Config> =
+    StorageMap<_, Blake2_256, ValidatorSet, u64, OptionQuery>;
+
+  // TODO: make this name explicit that this is an external network block
+  // LatestNetworkBlock maybe?
```
kayabaNerve: Sounds good.

processor/messages/src/lib.rs
```diff
@@ -61,6 +62,7 @@ pub mod sign {
   pub struct SignId {
     pub key: Vec<u8>,
     pub id: [u8; 32],
+    pub block: Option<BlockHash>,
```
kayabaNerve: Why is this an Option? Due to the SignId reuse for Batch?

processor/messages/src/lib.rs
```diff
@@ -11,10 +11,11 @@ use in_instructions_primitives::SignedBatch;
 use tokens_primitives::OutInstructionWithBalance;
 use validator_sets_primitives::{ValidatorSet, KeyPair};
 
-#[derive(Clone, Copy, PartialEq, Eq, Debug, Zeroize, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, Debug, Zeroize, Serialize, Deserialize)]
 pub struct SubstrateContext {
   pub serai_time: u64,
   pub coin_latest_finalized_block: BlockHash,
+  pub batches: Vec<u32>,
```
kayabaNerve: This shouldn't be in the SubstrateContext, yet in whichever specific message.

processor/src/main.rs
```diff
@@ -390,16 +390,16 @@ async fn handle_coordinator_msg<D: Db, C: Coin, Co: Coordinator>(
...
           let mut block_id = <C::Block as Block<C>>::Id::default();
           block_id.as_mut().copy_from_slice(&context.coin_latest_finalized_block.0);
+          let batches = &context.batches;
 
           let key = <C::Curve as Ciphersuite>::read_G::<&[u8]>(&mut key_vec.as_ref()).unwrap();
 
           // We now have to acknowledge every block for this key up to the acknowledged block
-          let (blocks, outputs) =
-            substrate_mutable.scanner.ack_up_to_block(txn, key, block_id).await;
+          let (_, outputs) = substrate_mutable.scanner.ack_up_to_block(txn, key, block_id).await;
```
kayabaNerve: We can have ack_up_to_block return the batches instead of the blocks.

processor/src/main.rs
```diff
@@ -657,45 +658,78 @@ async fn run<C: Coin, D: Db, Co: Coordinator>(mut raw_db: D, coin: C, mut coordi
...
-                let instruction = RefundableInInstruction::try_from(shorthand).ok()?;
-                // TODO2: Set instruction.origin if not set (and handle refunds in general)
-                Some(InInstructionWithBalance {
-                  instruction: instruction.instruction,
-                  balance: output.balance(),
-                })
-              }).collect()
-            };
+            let batch_id = substrate_mutable.scanner.last_batch_id(&txn);
+            let mut current_batch = batch_id + 1;
```
kayabaNerve: This can be simplified by using next_batch_id.

processor/src/main.rs
```diff
@@ -657,45 +658,78 @@ async fn run<C: Coin, D: Db, Co: Coordinator>(mut raw_db: D, coin: C, mut coordi
...
+              // TODO2: Set instruction.origin if not set (and handle refunds in general)
+              let in_ins = InInstructionWithBalance {
+                instruction: instruction.instruction,
+                balance: output.balance(),
+              };
+              let ins_size = in_ins.encode().len();
+
+              // batch has 38 bytes of other data
+              // TODO: should we make another const for this 38?
+              if batch_size + ins_size > MAX_BATCH_SIZE - 38 {
```
kayabaNerve: I'd just do batch.push(), if batch.encode().len() > MAX { batch.pop() }. It's slow but it's safe.

coordinator/src/tributary/scanner.rs
```diff
@@ -222,7 +222,7 @@ async fn handle_block<D: Db, Pro: Processors>(
                 spec.set().network,
                 CoordinatorMessage::Coordinator(
                   coordinator::CoordinatorMessage::BatchPreprocesses {
-                    id: SignId { key: todo!(), id: data.plan, attempt: data.attempt },
+                    id: SignId { key: todo!(), id: data.plan, block: None, attempt: data.attempt },
```
kayabaNerve: Here you set `block: None`, yet `substrate_signer.rs` sets `block: Some(_)` That means `substrate_signer.rs` and this have different IDs.

Why was the block added to the SignId? I may have already read it yet can't seem to place it exactly.

processor/src/substrate_signer.rs
```diff
@@ -204,16 +211,19 @@ impl<D: Db> SubstrateSigner<D> {
   }
 
   pub async fn sign(&mut self, txn: &mut D::Transaction<'_>, batch: Batch) {
-    if SubstrateSignerDb::<D>::completed(txn, batch.block.0) {
+    // Use the batch id as the ID
+    let mut id = [0u8; 32];
+    id.copy_from_slice(&batch.id.to_le_bytes());
```
kayabaNerve: I believe this will panic and you need to scope it as `id[.. 8].copy_from_slice`.

Regardless, we shouldn't use the u64, yet some hash of the batch (similar to how TXs have their IDs created).

processor/src/substrate_signer.rs
```diff
@@ -328,19 +338,24 @@ impl<D: Db> SubstrateSigner<D> {
       }
 
       CoordinatorMessage::BatchReattempt { id } => {
-        self.attempt(txn, id.id, id.attempt).await;
+        // TODO: this unwrap should be safe?
+        self.attempt(txn, id.id, id.block.unwrap(), id.attempt).await;
```
kayabaNerve: It's not, Coordinator used None for Batch messages.

processor/src/substrate_signer.rs
```diff
@@ -328,19 +338,24 @@ impl<D: Db> SubstrateSigner<D> {
...
       }
     }
   }
 
-  pub fn batch_signed(&mut self, txn: &mut D::Transaction<'_>, block: BlockHash) {
+  pub fn batch_signed(&mut self, txn: &mut D::Transaction<'_>, batch_id: u32) {
+    // convert into slice
+    let mut id = [0u8; 32];
+    id.copy_from_slice(&batch_id.to_le_bytes());
```
kayabaNerve: ID should be a Transcript of batch and there should be a common function for it.

processor/src/tests/substrate_signer.rs
```diff
@@ -25,12 +25,16 @@ async fn test_substrate_signer() {
...
+    key: keys[&participant_one].group_key().to_bytes().to_vec(),
+    id: [0u8; 32],
+    block: Some(block),
+    attempt: 0,
+  };
 
   let batch = Batch {
     network: NetworkId::Monero,
-    id: 5,
+    id: 0,
```
kayabaNerve: This was non-zero to ensure it worked with non-zero items.

processor/src/tests/substrate_signer.rs
```diff
@@ -25,12 +25,16 @@ async fn test_substrate_signer() {
   let participant_one = Participant::new(1).unwrap();
 
   let block = BlockHash([0xaa; 32]);
-  let actual_id =
-    SignId { key: keys[&participant_one].group_key().to_bytes().to_vec(), id: block.0, attempt: 0 };
+  let actual_id = SignId {
+    key: keys[&participant_one].group_key().to_bytes().to_vec(),
+    id: [0u8; 32],
```
kayabaNerve: ID should be non-zero.

kayabaNerve (changes requested):

coordinator/src/tributary/scanner.rs
akildemir: I believe the only reason block hash needed was due to https://github.com/akildemir/serai/blob/0eb4e2a57b287c555e1c52eb376fbe7c4543ce0a/processor/messages/src/lib.rs#L121 this member function on `CoordinatorMessage`

processor/messages/src/lib.rs
akildemir: Yes. It is used in SignId for batches but not in `coordinator/tributary/scanner` see my reply to https://github.com/serai-dex/serai/pull/316#discussion_r1283255738 as well

processor/src/main.rs
akildemir: I believe not needed since we get the acknowledged batches directly from serai anymore. Also that would require extra step of saving block_hash -> batches pair into db and reading etc etc..

PR #322, opened 2023-07-20
.rustmfmt.toml: add edition

Steven-Chang:
I get this error when I try to save:

Screenshot 2023-07-20 at 12 25 29 pm

Seems like it happens to a few people and this fixed it as per:
https://github.com/rust-lang/rust-analyzer/issues/1959


kayabaNerve:
Thanks.

PR #325, opened 2023-07-25
docs/protocol/Staking.md: Note regarding delegation being disabled.

Steven-Chang:
For your consideration. All good if you don't want to use it.

Screenshot 2023-07-25 at 12 32 44 pm


kayabaNerve:
Re-reading the code from 7m ago, delegation wasn't disabled: https://github.com/serai-dex/serai/blob/staking/substrate/staking/pallet/src/lib.rs

I'd rather delete this or leave it untouched until we have something accurate to move to (where nothing now is accurate).

Thanks for working on enuring it's up to date though :) Considering you have the perspective of a new user reading the docs, I'll let you decide which is better for now.

Steven-Chang:
Thanks. I think it'd be best to remove it. I'll close this PR and do a new PR deleting this file.

kayabaNerve:
We also could've edited the title and done it here :P

PR #336, opened 2023-08-05
cargo.toml: sort

Steven-Chang:


kayabaNerve:
As a FYI, my personal sort preference is generally std, libraries, first-party/crate, primitives to complex. It's a bit erratic, as it's opinionated, yet that explains most of why the tomls are as they are.

For me, despite alpha sorting intending to be unambiguous, I find it difficult to quickly locate objects. That's why it wasn't prior adopted. If there's sufficient interest, I'll adapt, but until then...

Steven-Chang:
Thanks for the info.  I accidentally did a PR and instant closed but will consider what you wrote here.

PR #363, opened 2023-08-30
initial dex pallet support

akildemir:


substrate/client/src/serai/dex.rs
```diff
@@ -0,0 +1,60 @@
...
+    Payload::new(
+      PALLET,
+      "create_pool",
+      scale_composite(dex::Call::<Runtime>::create_pool { asset1: asset, asset2: Coin::Serai }),
+    )
+  }
+
+  pub fn add_liquidity(
+    asset: Coin,
+    asset_amount: SubstrateAmount,
```
kayabaNerve: Let's call them `coin`, as they're presumed to be coins (3 out of 4 initially supported coins/tokens are coins), or tokens since they're tokens on Serai.

`asset` is a term with legal implications which do not apply to anything here, AFAIK.

substrate/client/tests/common/dex.rs
```diff
@@ -0,0 +1,45 @@
...
+
+#[allow(dead_code)]
+pub async fn create_pool(asset: Coin, pair: Pair) -> [u8; 32] {
+  let serai = serai().await;
+
+  let tx = serai
+    .sign(&PairSigner::new(pair), &Serai::create_pool(asset), 0, BaseExtrinsicParamsBuilder::new())
+    .unwrap();
+
+  // TODO: this func should panic if can't found the tx?
```
kayabaNerve: Doesn't publish_tx panic on failure to publish? Are you referring to successful publish, failure to include in block? That shouldn't happen, not that I'd mind extending publish_tx to handle such an edge case if it does.

substrate/client/tests/common/dex.rs
```diff
@@ -0,0 +1,45 @@
...
+  let serai = serai().await;
+  let address = pair.public();
+
+  let tx = serai
+    .sign(
+      &PairSigner::new(pair),
+      &Serai::add_liquidity(asset, asset_amount.0, sri_amount.0, 1, 1, address.into()),
+      // TODO: we should handle the nonces better in tests. This nonce is
+      // 1 because we have to make create_pool call first and nonce of that is 0.
+      // maybe take it as parameter but just seems like delaying the problem.
```
kayabaNerve: Yes, just take it as a param and let the test track nonces.

You can also ask Serai for the nonce? Though may require an extra method in the Serai client.

substrate/client/tests/common/in_instructions.rs
```diff
@@ -45,3 +47,23 @@ pub async fn provide_batch(batch: Batch) -> [u8; 32] {
 
   block
 }
+
+#[allow(dead_code)]
+pub async fn mint_coin(balance: Balance, network: NetworkId, address: SeraiAddress) -> [u8; 32] {
+  let id = 0;
```
kayabaNerve: id should be a parameter OR gotten from Serai. The client now has a method for the last-executed batch ID (as of ~3 days ago?).

substrate/client/tests/common/in_instructions.rs
```diff
@@ -45,3 +47,23 @@ pub async fn provide_batch(batch: Batch) -> [u8; 32] {
...
+    network,
+    id,
+    block: block_hash,
+    instructions: vec![InInstructionWithBalance {
+      instruction: InInstruction::Transfer(address),
+      balance,
+    }],
+  };
+
+  provide_batch(batch).await
```
kayabaNerve, line 82: Cursed, but it works without us having to add the `sudo` pallet for tests.

:+1:

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -16,20 +16,33 @@ use sp_application_crypto::sr25519::Signature;
...
+use serai_primitives::{
+  BlockHash, Balance, NetworkId, SeraiAddress, ExternalAddress, Coin, Amount, pallet_address,
+};
 
 mod shorthand;
 pub use shorthand::*;
 
 pub const MAX_BATCH_SIZE: usize = 25_000; // ~25kb
 
+pub const IN_INSTRUCTION_EXECUTOR: SeraiAddress = pallet_address(b"InInstructionExecutor");
```
kayabaNerve, line 30: Add a comment clarifying the distinction?

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -16,20 +16,33 @@ use sp_application_crypto::sr25519::Signature;
...
+  Dex(DexCall),
+}
+
+#[derive(
+  Clone, PartialEq, Eq, Debug, Serialize, Deserialize, Encode, Decode, MaxEncodedLen, TypeInfo,
+)]
+#[cfg_attr(feature = "std", derive(Zeroize))]
+pub enum DexCall {
+  AddLiquidity(SeraiAddress),          // address to sent the lp tokens to
+  Swap(Coin, ExternalAddress, Amount), // to_coin, to_address, min_out_amount
```
kayabaNerve: You can swap to SRI, which would specify a SeraiAddress. I think you should use Data for the middle portion, though I have to double check the documentation for what's already specified (though those docs may be invalid under current considerations).

kayabaNerve: Also, to not have to bother with dated docs, it's perhaps best just to add
`// TODO: Sync with docs`
so it's labelled to be done later. I'm fine kicking it.

substrate/tokens/pallet/src/lib.rs
```diff
@@ -37,7 +37,7 @@ pub mod pallet {
   pub struct Pallet<T>(PhantomData<T>);
 
   impl<T: Config> Pallet<T> {
-    fn burn_internal(
+    pub fn burn_internal(
```
kayabaNerve, line 40: My general convention is when we have multiple paths over the same core functionality, to add the two paths *with a single `_internal` execution block*. The higher paths sometimes just fill in any pre-set variables, *yet sometimes do path specific execution*.

I'm not an advocate for making an `_internal` function `pub` unless we make it not an `_internal` function. We'd have to

1) Name it something else, which shows it's a complete function (not one with a prerequisites/assumptions on the caller)
2) Double check its current calls

As for the name, I'd suggest calling it burn but my guess is that'd conflict with the extrinsic-reachable Call? It may not due to the macro transformations which Call undergoes? Feel free to try renaming this to burn.

substrate/runtime/src/lib.rs
```diff
@@ -388,9 +502,12 @@ construct_runtime!(
     TransactionPayment: transaction_payment,
 
     Assets: assets,
+    PoolAssets: assets::<Instance1>::{Pallet, Call, Storage, Event<T>},
```
kayabaNerve: Actually probably the best way to do this now that I think about it :+1: Congrats for figuring this out.

Prob best to rename it `LiquidityTokens` though? The current one is ambiguous about tokens in-pool vs tokens of pool.

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -70,10 +80,110 @@ pub mod pallet {
     StorageMap<_, Blake2_256, NetworkId, BlockHash, OptionQuery>;
 
   impl<T: Config> Pallet<T> {
-    fn execute(instruction: InInstructionWithBalance) -> Result<(), ()> {
+    fn execute(instruction: InInstructionWithBalance) -> DispatchResult {
```
kayabaNerve, line 92: This type may overlap with the `DispatchResult` type alias, but it's not used as a `DispatchResult`. `execute_batch` (the dispatchable) gets it, performs is_err, but won't `?`.

substrate/runtime/src/lib.rs
```diff
@@ -603,4 +720,21 @@ sp_api::impl_runtime_apis! {
...
+    fn quote_price_exact_tokens_for_tokens(asset1: Coin, asset2: Coin, amount: SubstrateAmount, include_fee: bool) -> Option<SubstrateAmount> {
+      Dex::quote_price_exact_tokens_for_tokens(asset1, asset2, amount, include_fee)
+    }
+    fn quote_price_tokens_for_exact_tokens(asset1: Coin, asset2: Coin, amount: SubstrateAmount, include_fee: bool) -> Option<SubstrateAmount> {
+      Dex::quote_price_tokens_for_exact_tokens(asset1, asset2, amount, include_fee)
+    }
+    fn get_reserves(asset1: Coin, asset2: Coin) -> Option<(SubstrateAmount, SubstrateAmount)> {
+      Dex::get_reserves(&asset1, &asset2).ok()
+    }
+  }
```
kayabaNerve: I think this only has benefit if you additionally pull in the RPC hooks for the DEX.

substrate/runtime/src/lib.rs
```diff
@@ -304,6 +361,63 @@ impl validator_sets::Config for Runtime {
...
+  type LPFee = ConstU32<3>; // 0.3%
+  type PoolSetupFee = ConstU64<0>; // Asset class deposit fees are sufficient to prevent spam
+  type LiquidityWithdrawalFee = LiquidityWithdrawalFee;
+  type MintMinLiquidity = ConstU64<100>;
+
+  type MaxSwapPathLength = ConstU32<3>; // asset1 -> SRI -> asset2
+
+  type PalletId = AssetConversionPalletId;
+  type PoolSetupFeeReceiver = AssetConversionOrigin;
+  type AllowMultiAssetPools = AllowMultiAssetPools;
```
kayabaNerve: It'd be simpler to inline this. `AllowMultiAssetPools = ConstBool<false>`.

substrate/runtime/src/lib.rs
```diff
@@ -143,6 +149,16 @@ parameter_types! {
...
+
+  pub const AssetConversionPalletId: PalletId = PalletId(*b"DexPalet"); // TODO: this requires exactly 8 byte?
+  pub const LiquidityWithdrawalFee: Permill = Permill::from_percent(1);
+  pub const AllowMultiAssetPools: bool = false;
+}
+
+ord_parameter_types! {
+  // TODO: normally we should do AssetConversionPalletId::get() instead of b"DexPalet"
+  // but that gives a lifetime error for now.
+  pub const AssetConversionOrigin: PublicKey = PublicKey::from(pallet_address(b"DexPalet"));
```
kayabaNerve, line 161: Pallet, not Palet.

substrate/runtime/src/lib.rs
```diff
@@ -304,6 +361,63 @@ impl validator_sets::Config for Runtime {
...
+  type MultiAssetId = Coin;
+  type MultiAssetIdConverter = AssetConverter;
+  type PoolAssetId = u32;
+
+  type Assets = Assets;
+  type PoolAssets = PoolAssets;
+
+  type LPFee = ConstU32<3>; // 0.3%
+  type PoolSetupFee = ConstU64<0>; // Asset class deposit fees are sufficient to prevent spam
+  type LiquidityWithdrawalFee = LiquidityWithdrawalFee;
```
kayabaNerve, line 407: This should be inlined as 0.

substrate/runtime/src/lib.rs
```diff
@@ -304,6 +361,63 @@ impl validator_sets::Config for Runtime {
...
+  type MintMinLiquidity = ConstU64<100>;
```
kayabaNerve: I believe Uniswap used 10k to prevent their rounding errors. Without my own understanding of the exact issues, I'd mirror them.

substrate/runtime/src/lib.rs
```diff
@@ -304,6 +361,63 @@ impl validator_sets::Config for Runtime {
...
+    *asset
+  }
+}
+
+impl dex::Config for Runtime {
+  type RuntimeEvent = RuntimeEvent;
+  type Currency = Balances;
+  type Balance = SubstrateAmount;
+  type AssetBalance = SubstrateAmount;
+  type HigherPrecisionBalance = u128;
```
kayabaNerve, line 395: When this was written for Substrate, they probably expected u128/u256, not u64/u128. I'd add a TODO to review if this should be u64/u128 or u64/u256 (and rounding in general).

substrate/runtime/src/lib.rs
```diff
@@ -304,6 +361,63 @@ impl validator_sets::Config for Runtime {
...
+
+  type AssetId = Coin;
+  type MultiAssetId = Coin;
+  type MultiAssetIdConverter = AssetConverter;
+  type PoolAssetId = u32;
```
kayabaNerve, line 400: ... we _can_ reuse Coin here. I don't like that, from a safety perspective, but the next alternative would be to copy the Coin enum to also define a LiquidityToken enum (minus the Serai entry).

substrate/runtime/src/lib.rs
```diff
@@ -292,6 +319,36 @@ impl assets::Config for Runtime {
   type BenchmarkHelper = SeraiAssetBenchmarkHelper;
 }
 
+pub type PoolAssetsInstance = assets::Instance1;
+impl assets::Config<PoolAssetsInstance> for Runtime {
+  type RuntimeEvent = RuntimeEvent;
+  type Balance = SubstrateAmount;
+  type RemoveItemsLimit = ConstU32<1000>; // 0?
```
kayabaNerve: Add a TODO if this is an active question?

substrate/runtime/src/lib.rs
```diff
@@ -143,6 +149,16 @@ parameter_types! {
     );
 
   pub const MaxAuthorities: u32 = 100;
+
+  pub const AssetConversionPalletId: PalletId = PalletId(*b"DexPalet"); // TODO: this requires exactly 8 byte?
```
kayabaNerve: Pallet, not palet.

kayabaNerve (commented):
I have yet to review substrate/in-instructions/pallet/src/lib.rs and substrate/client/tests/dex.rs. Overall looks as expected :+1:

substrate/client/src/serai/dex.rs
akildemir: yea make sense 👍  Just used assets because substrate uses it like that.

substrate/in-instructions/pallet/src/lib.rs
akildemir: `?`would return the `execute_batch` but we have multiple in_ins in a batch and `Self::execute()` is called in a for loop per instruction. So returning there would  prevents us from executing remanning instructions in the batch. I thought that is why we deposit instruction failure event if that is what you mean. But yea maybe event can have a `reason` field that holds the `DispatchError`.

substrate/runtime/src/lib.rs
akildemir: This TODO explains that we should normally use `AssetConversionPalletId` to construct the pallet address and if you have a look https://github.com/serai-dex/serai/pull/363#discussion_r1310780660 there is another TODO there that say it should be exactly 8 byte? The reason is pallet ids are fixed 8 bytes. So it was intentionally named `DexPalet` to make it 8 byte because `DexPallet` is 9 byte.

substrate/runtime/src/lib.rs
akildemir: That is what I tried initially but lp tokens are constructed on the fly. So pallet should be able to auto make ids for them. Hence, `PoolAssetId` should be something that is incremental, something that implements `Incrementable`. 

I guess we can still implement this for an enum? But enum will only have certain elements and this will create a new lp token per pool automatically.  I mean we will only have certain amount of pools right? so enum still makes sense? yea I was confused there too.

substrate/runtime/src/lib.rs
akildemir: yes it was a question lol. Same type `RemoveItemsLimit` also exist in assets pallet config. So should they be same ?

substrate/tokens/pallet/src/lib.rs
akildemir: doesn't compile with naming it `burn.  `duplicate definitions`  error so yea prob have to name it something else.

substrate/runtime/src/lib.rs
kayabaNerve: Making the same is fine. They may have different costs of removal, so the actual limit on execution may be distinct, but I'm not concerned about it at this time.

substrate/runtime/src/lib.rs
kayabaNerve: And then we can't guarantee the increment matches the intended order of Coin.

Got it. This is fine for now :+1:

PR #373, opened 2023-09-16
Staking pallet

akildemir:
- Implements stake/unstake
 - Allocate/deallocate to specific network validator set
 - Validator set rotation for serai network


substrate/client/tests/validator_sets.rs
```diff
@@ -38,7 +38,7 @@ serai_test!(
         .get_new_set_events(serai.get_block_by_number(0).await.unwrap().unwrap().hash())
         .await
         .unwrap(),
-      [NetworkId::Bitcoin, NetworkId::Ethereum, NetworkId::Monero]
+      [NetworkId::Bitcoin, NetworkId::Ethereum, NetworkId::Monero, NetworkId::Serai]
```
kayabaNerve: Serai should be first, not last in these cases (a few follow in the files changed view).

substrate/staking/pallet/LICENSE
```diff
@@ -0,0 +1,15 @@
+AGPL-3.0-only license
+
+Copyright (c) 2022 Luke Parker
```
kayabaNerve: It's now 2023, so this would need an extension.

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+  use serai_validator_sets_primitives::{ValidatorSet, Session};
+  use staking_primitives::AllocatedStaking;
+
+  use validator_sets_pallet::{Config as VsConfig, Pallet as VsPallet};
+  use pallet_session::{Config as SessionConfig, SessionManager, Pallet as SessionPallet};
+
+  #[pallet::error]
+  pub enum Error<T> {
+    BondUnavailable,
+    InSufficientAllocation,
```
kayabaNerve: "Insufficient"

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+      network: NetworkId,
+      #[pallet::compact] amount: u64,
+    ) -> DispatchResult {
+      let account = ensure_signed(origin)?;
+
+      // add to amount bonded
+      Self::allocate_internal(&account, amount)?;
+
+      // add to participants list for the network
+      let result = VsPallet::<T>::add_participant(account, Amount(amount), network);
```
kayabaNerve: Hm. It's not adding if they were already added.

`increase_allocation`? What do you think would be a more apt name?

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+      if result.is_err() {
+        Self::deallocate_internal(&account, amount).unwrap();
```
kayabaNerve: If this returns an error, all state changes from the call will be discarded (since https://github.com/paritytech/substrate/pull/11431). Accordingly, you should simply be able to `?` the error, without this restoration code (which may have its own bug).

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+        return result;
+      }
+
+      Ok(())
+    }
+
+    /// Allocate `amount` to a given validator set.
+    #[pallet::call_index(3)]
+    #[pallet::weight((0, DispatchClass::Operational))] // TODO
+    pub fn deallocate(
```
kayabaNerve: Incorrect doc comment.

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+      network: NetworkId,
+      #[pallet::compact] amount: u64,
+    ) -> DispatchResult {
+      let account = ensure_signed(origin)?;
+
+      // remove the participant if necessary.
+      // we can't directly deallocate here, since the leaving validator
+      // will be removed after the next session. We only deallocate then
+      // on `end_session` for the right index.
+      VsPallet::<T>::maybe_remove_participant(account, Amount(amount), network)
```
kayabaNerve: `decrease_allocation`?

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+    }
+  }
+
+  /// Call order is end_session(i - 1) -> start_session(i) -> new_session(i + 1)
+  /// new_session(i + 1) is called immediately after start_session(i) returns then
+  /// we wait until the session ends then get a call to end_session(i) and so on.
```
kayabaNerve: The `///` signifies doc comment, yet this won't generate documentation. `//`?

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+  impl<T: Config> SessionManager<T::ValidatorId> for Pallet<T> {
+    fn new_session(new_index: u32) -> Option<Vec<T::ValidatorId>> {
+      let next_validators = VsPallet::<T>::next_validator_set(new_index, NetworkId::Serai);
+
+      // Returning None will keep the previous set going.
+      if next_validators.is_empty() {
+        return None;
+      }
```
kayabaNerve, line 165: Why will next_validator_set return an empty list?

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+
+    fn end_session(end_index: u32) {
+      // do the deallocation of those validator funds
+      // who will not be in the set next session.
+      let key = ValidatorSet { session: Session(end_index + 1), network: NetworkId::Serai };
+      let deallocating_validators = VsPallet::<T>::deallocating_validators(key);
+      for (account, amount, _) in deallocating_validators {
+        // we can unwrap because we are not deallocating more than allocated.
+        <Self as AllocatedStaking<T>>::deallocate(&account, amount.0).unwrap();
+      }
```
kayabaNerve: This is O(n) (meaning `end_session` can be nuked into an extremely expensive call).

Can we have it where, once a session expires (Serai or external network), the validator submits a TX to refresh their allocation (instead of handling it here for them)? We can even do it automatically when `allocate`/`unstake` is called so multiple TXs aren't needed.

It does require storing deallocations as:

`Queued (amount, session)`

And if further deallocation is queued in a new session, shifting the existing deallocation from `Queued (amount, session)` to `Complete amount`.

substrate/staking/pallet/src/lib.rs
```diff
@@ -0,0 +1,219 @@
...
+      VsPallet::<T>::end_session(end_index, NetworkId::Serai);
+    }
+
+    fn start_session(start_index: u32) {
+      let validators = SessionPallet::<T>::validators();
+      VsPallet::<T>::start_session(start_index, NetworkId::Serai, validators)
+    }
+  }
+
+  impl<T: Config> AllocatedStaking<T> for Pallet<T> {
```
kayabaNerve: Why did we introduce this new trait? Because else we had validator-sets pallet import staking pallet and staking pallet implement validator-sets pallet?

substrate/validator-sets/pallet/Cargo.toml
```diff
@@ -39,13 +41,17 @@ std = [
...
+
   "serai-primitives/std",
   "validator-sets-primitives/std",
 ]
 
 runtime-benchmarks = [
   "frame-system/runtime-benchmarks",
   "frame-support/runtime-benchmarks",
+
+	"pallet-babe/runtime-benchmarks",
```
kayabaNerve: Is this a tab instead of two spaces?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -63,13 +70,51 @@ pub mod pallet {
   #[pallet::getter(fn keys)]
   pub type Keys<T: Config> = StorageMap<_, Twox64Concat, ValidatorSet, KeyPair, OptionQuery>;
 
+  /// The Validators that has enough bond allocated
+  /// and set to be joining to the validator set
+  /// in the next session.
+  #[pallet::storage]
+  #[pallet::getter(fn joining_validators)]
+  #[pallet::unbounded]
+  pub type JoiningValidators<T: Config> =
```
kayabaNerve: What about validators who aren't newly joining yet have allocated more bond?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -63,13 +70,51 @@ pub mod pallet {
...
+  /// some funds.
+  #[pallet::storage]
+  #[pallet::getter(fn deallocating_validators)]
+  #[pallet::unbounded]
+  pub type DeallocatingValidators<T: Config> =
+    StorageMap<_, Twox64Concat, ValidatorSet, Vec<(T::AccountId, Amount, bool)>, ValueQuery>;
+
+  #[pallet::storage]
+  #[pallet::getter(fn current_session)]
+  pub type CurrentSessionIndex<T> = StorageValue<_, Session, ValueQuery>;
```
kayabaNerve: https://paritytech.github.io/polkadot-sdk/master/pallet_session/pallet/struct.Pallet.html#method.current_index ?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -63,13 +70,51 @@ pub mod pallet {
...
+    /// Validator Set already generated keys.
+    AlreadyGeneratedKeys,
+    /// An invalid MuSig signature was provided.
+    BadSignature,
+    /// Not enough bond to participate in a set.
+    InSufficientBond,
+    /// Validator wasn't registered or active.
+    NonExistentValidator,
+    /// Trying to deallocate more than allocated.
+    InSufficientAllocation,
```
kayabaNerve: "Insufficient"

substrate/runtime/src/lib.rs
```diff
@@ -176,6 +178,20 @@ impl Contains<RuntimeCall> for CallFilter {
...
+        call,
+        staking::Call::stake { .. } |
+          staking::Call::unstake { .. } |
+          staking::Call::allocate { .. } |
+          staking::Call::deallocate { .. }
+      );
+    }
+
+    if let RuntimeCall::Session(call) = call {
+      return matches!(call, session::Call::set_keys { .. } | session::Call::purge_keys {});
```
kayabaNerve: Do we bother exposing purge_keys? Seems like something it only makes sense for us to call internally.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -97,17 +142,32 @@ pub mod pallet {
...
+  impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    /// Called when a block is initialized.
+    fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+      // fire new_set 100 blocks prior to session change.
+      // we can "predict" next session start by getting the
+      // next epoch time from babe since sessions rotate on epoch change.
+      // This is only approximation since slots != blocks.
+      let next_start: u64 = Babe::<T>::next_epoch().start_slot.into();
+      if next_start - 100 == n.saturated_into::<u64>() {
+        // TODO: only for serai or for all networks?
```
kayabaNerve: I'd just do all of them here.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -127,6 +187,227 @@ pub mod pallet {
...
+    ) -> DispatchResult {
+      let key = ValidatorSet { session: Self::current_session(), network };
+      if amount < Self::validator_set(key).unwrap().bond {
+        Err(Error::<T>::InSufficientBond)?;
+      }
+
+      // Since this participant will be included after the next session, save it for the set
+      // where it will be active.
+      let participant_set =
+        ValidatorSet { session: Session(Self::current_session().0 + 2), network };
```
kayabaNerve: `+ 1` if not Serai, as only Serai has this limitation?

Though also, it's frustrating that it'll take two weeks to join the Serai validator set. Do we want to consider halving the discussed session time from 7 days to 3.5, yet only rotating external networks every other session? That'd set a maximum of 7 days to join the Serai set (not 2 weeks), while maintaining the 7 day interval for external networks.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -127,6 +187,227 @@ pub mod pallet {
...
+      JoiningValidators::<T>::try_mutate_exists(participant_set, |existing| {
+        if existing.is_some() {
+          existing.as_mut().unwrap().push((account, amount));
```
kayabaNerve: They may already be present?

kayabaNerve: Also, this is an O(n) algorithm. If we move to keying off ValidatorSet *and account*, I believe you can have O(1) insertion (well, the complexity would be the underlying DBs which should be O(log n), yet we can handwave it) yet still trivially do an O(n) iteration over the storage prefix. We should at least `TODO` for this.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -127,6 +187,227 @@ pub mod pallet {
...
+        }
+      };
+
+      // check validator has enough to deallocate
+      if amount > pair.1 {
+        Err(Error::<T>::InSufficientAllocation)?;
+      }
+
+      // if the remaining bond is not enough remove from the validator set and
+      // deallocate all bond instead of just "amount" since they will be leaving.
```
kayabaNerve: I'd error, unless they deallocate the entire amount. Then someone trying to deallocate 10 doesn't undo a 100k bond position.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -127,6 +187,227 @@ pub mod pallet {
...
+      let current_set =
+        Self::validator_set(ValidatorSet { session: Self::current_session(), network }).unwrap();
+      let key = ValidatorSet { session: Session(Self::current_session().0 + 2), network };
+
+      let pair = match current_set.participants.iter().find(|p| p.0 == account) {
+        Some(p) => *p,
+        None => {
+          // check whether it is still in the joining set before getting active
+          let joining_set = JoiningValidators::<T>::get(key);
+          *joining_set.iter().find(|p| p.0 == account).ok_or(Error::<T>::NonExistentValidator)?
```
kayabaNerve: This seems error-prone (removing EITHER from current or joining, yet still going through `DeallocatingValidators`). If they have yet to join, can't we deallocate in place without DV?

kayabaNerve (changes requested):

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -127,6 +187,227 @@ pub mod pallet {
...
+        .filter(|(_, _, leaving)| !leaving)
+        .map(|(id, amount, _)| (*id, *amount))
+        .collect::<Vec<_>>();
+      let prev =
+        Self::validator_set(ValidatorSet { session: Session(new_index - 1), network }).unwrap();
+
+      // append them all together
+      joining.extend(prev.participants);
+      let last_add_index = joining.len() - 1;
+      joining.extend(deallocating);
```
kayabaNerve: Why is `deallocating` merged into `joining`?

substrate/staking/pallet/src/lib.rs
akildemir: `handle_allocation`? since then inside of the function, we either add or increase.

substrate/staking/pallet/src/lib.rs
akildemir: you mean ` VsPallet::<T>::add_participant(account, Amount(amount), network)` (directly returns the `DispatchResult` like in the `deallocate`) and this will revert the allocation we did if it returns error?

substrate/staking/pallet/src/lib.rs
akildemir: Actually no handle is very general, it might also mean the decrease. So I guess best is `increase_allocation`, since also can assume that in the case of add, validator had `0` allocation an now has more than `0`.

substrate/staking/pallet/src/lib.rs
akildemir: I was actually wondering that myself lol. Why do we have it? it was just there in the original staking folder so I just imported it lol. But will delete because doesn't seem necessary now 👍

akildemir: uhh but that makes the whole `staking/primitives` crate empty? Should we nuke the whole crate then?

substrate/validator-sets/pallet/src/lib.rs
akildemir: They are added to the same set, yea renaming is necessary. It should be `AllocatingValidators` .

substrate/validator-sets/pallet/src/lib.rs
akildemir: I mean we would be importing the whole pallet to vs pallet just for this call? ValidatorSet pallet doesn't deal with session pallet at all currently. But don't mind, whichever is best.

substrate/runtime/src/lib.rs
akildemir: umm right. Purge keys takes effect in the next session, so exposing it would remove a validator from the set in the session pallet although we think it is  still in the set in the vs pallet. 👍

substrate/validator-sets/pallet/src/lib.rs
akildemir: yea we can set the rotation time to 3.5 days, but it might still be better for other networks to follow the serai net? that would purify the design, since all the commands would be coming from the session pallet. Otherwise  we are kinda looking at  implementing our own session pallet for other networks. Which if we were gonna do that, we can also use it for the serai net as well. But yea migration taking 6 hrs every 3.5 days is also a problem in that case.

substrate/validator-sets/pallet/src/lib.rs
akildemir: I didn't understand why it would be O(n), we are not iterating through anything if I am not mistaken.

substrate/validator-sets/pallet/src/lib.rs
akildemir: uhh we should be able to through a callback? The problem was since staking -> vs-pallet, we couldn't call the staking pallet function. Idk how desirable a callback tho.

substrate/validator-sets/pallet/src/lib.rs
akildemir: all 3 vectors merged into 1 to not repeat [following](https://github.com/akildemir/serai/blob/baaf96f59c06b1157e75059d275cc69de5625f1a/substrate/validator-sets/pallet/src/lib.rs#L348) block of code.

substrate/staking/pallet/src/lib.rs
akildemir: Hmm yes that is possible. We can do the the refresh whenever we get a call to `allocate`/`unstake`.  We wouldn't need to store additional info for it(as far as immediately thinking about it.) we could just check the active bond amount, and release any amount that is more than necessary. The main concern would be future complications due to not having refreshed state immediately.

PR #377, opened 2023-09-24
Add support for multiple multisigs to the processor

kayabaNerve:


kayabaNerve:
Since this shouldn't break anything, I'm fine merging it without the matching coordinator code yet.

Issue #366, opened 2023-08-31
Consider having `Provided` solely block the scanner, not the Tributary


kayabaNerve:
`Provided` halting the Tributary was a safety feature, yet it has the DoS of preventing the 33% who have yet to receive the Provided Transaction from participating in consensus. By having verify_block still reject it, yet add_block handle it, we prevent this DoS.

There'd still be the guarantee 67% have come to sync on the item, and the scanner still waiting for the `Provided` transaction to be locally provided will ensure safety.

This should reduce risk of stutters in Tendermint.

kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/6238640448/job/16934913654 failed, likely due to a proposer who was late on a Provided failing to propose, adding several seconds of latency to the run.

kayabaNerve:
To clarify, a proposer late on a Provided did fail to propose. The "likely" is referring to that being THE reason the run failed. Not due to uncertainty about that event occurring.

kayabaNerve:
Fixed by #381.

PR #381, opened 2023-09-26
Improve provided handling

akildemir:
Instead of preventing a node adding a tributary block that contains provided tx(s) that doesn't locally available, we add the block to tributary(since it would be acknowledged the by the network anyway) and let the node participates in the consensus as usual. But we stop handling(scanning) the block by the coordinator until it has the provided txs in the block locally available.

So provided tx completion now happens on coordinator scanner instead of within the tributary.

Tackles https://github.com/serai-dex/serai/issues/366.

### DIFF
--- a/coordinator/src/tributary/scanner.rs
+++ b/coordinator/src/tributary/scanner.rs
@@ -12,6 +12,7 @@ use tributary::{
     tx::{TendermintTx, decode_evidence},
     TendermintNetwork,
   },
+  TransactionKind,
 };
 
 use serai_db::DbTxn;
@@ -121,8 +122,18 @@ pub(crate) async fn handle_new_blocks<
   while let Some(next) = tributary.block_after(&last_block) {
     let block = tributary.block(&next).unwrap();
 
-    if !tributary.provided_waiting_list_empty() {
-      return;
+    for tx in &block.transactions {
+      // since we know provided txs are the first in the block, we can assume that
+      // all of them were ok if we haven't returned yet and got a new kind, so we can
+      // break and continue to scan the block.
+      let TransactionKind::Provided(order) = tx.kind() else {
+        break;
+      };
+
+      // make sure we have all the provided txs in this block locally
+      if !tributary.provided_txs_ok_for_block(&block.hash(), order) {
+        return;
+      }
     }
 
     handle_block::<_, _, _, _, _, _, P>(
--- a/coordinator/src/tributary/scanner.rs
+++ b/coordinator/src/tributary/scanner.rs
@@ -120,6 +120,11 @@ pub(crate) async fn handle_new_blocks<
   let mut last_block = db.last_block(genesis);
   while let Some(next) = tributary.block_after(&last_block) {
     let block = tributary.block(&next).unwrap();
+
+    if !tributary.provided_waiting_list_empty() {
+      return;
+    }
+
     handle_block::<_, _, _, _, _, _, P>(
       db,
       key,
--- a/coordinator/src/tributary/scanner.rs
+++ b/coordinator/src/tributary/scanner.rs
@@ -7,12 +7,11 @@ use ciphersuite::{Ciphersuite, Ristretto};
 use serai_client::{validator_sets::primitives::ValidatorSet, subxt::utils::Encoded};
 
 use tributary::{
-  Transaction as TributaryTransaction, Block, TributaryReader,
+  TransactionKind, Transaction as TributaryTransaction, Block, TributaryReader,
   tendermint::{
     tx::{TendermintTx, decode_evidence},
     TendermintNetwork,
   },
-  TransactionKind,
 };
 
 use serai_db::DbTxn;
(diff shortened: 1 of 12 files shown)


coordinator/src/tributary/scanner.rs
```diff
@@ -114,12 +117,38 @@ pub(crate) async fn handle_new_blocks<
   processors: &Pro,
   publish_serai_tx: PST,
   spec: &TributarySpec,
-  tributary: &TributaryReader<D, Transaction>,
+  tributary: &Arc<Tributary<D, Transaction, P>>,
```
kayabaNerve: You shouldn't need to take `&Arc`.  `&Arc::new(x)` should deref to `&x`.

I'm unsure the impact of this (taking the Tributary) on latency. It shouldn't be atrocious with the recent localization I did, yet even if it's perfectly fine (which I'd hope), it shouldn't be necessary. We should add a method to the reader to check if an on-chain provided transaction was provided into the tributary ourselves yet.

IIRC, `Provided` are perfectly ordered within some queue, yet aren't guaranteed to have distinct hashes. Accordingly, on local provision, you'd save `Provided queue i hash`. Then, on-chain when you see a Provided, you'd check `Provided queue i hash`.

There is a pseudo-unsafety here, where `Provided queue i` is insufficient as what other validators came to may be distinct from what we came to. Under this, we'd likely enter a wait loop. The question is if the tributary should return an error in such a case. It likely should.

coordinator/src/tributary/scanner.rs
```diff
@@ -114,12 +117,38 @@ pub(crate) async fn handle_new_blocks<
...
+      if let TransactionKind::Provided(_) = tx.kind() {
+        let tx_hash = tx.hash();
+        if !provided_transactions.chunks_exact(32).any(|hash| {
+          let h: [u8; 32] = hash.try_into().unwrap();
+          h == tx_hash
+        }) {
+          return;
+        }
+      }
+    }
```
kayabaNerve: This is an O(n^2) algorithm which breaks safety. Provided are expected to be ordered, yet this disregards order.

IF we were going to have this unsafe algorithm, though no, we do need to keep the ordering checks, we could at least make it O(n) with a HashSet...

coordinator/src/tributary/scanner.rs
```diff
@@ -114,12 +117,38 @@ pub(crate) async fn handle_new_blocks<
...
+    // of the block until we have all of them.
+    let provided_transactions = tributary.provided_transactions().await;
+    for tx in &block.transactions {
+      if let TransactionKind::Provided(_) = tx.kind() {
+        let tx_hash = tx.hash();
+        if !provided_transactions.chunks_exact(32).any(|hash| {
+          let h: [u8; 32] = hash.try_into().unwrap();
+          h == tx_hash
+        }) {
+          return;
```
kayabaNerve: I'm unsure if this should be a return or a break. I believe a break.

coordinator/tributary/src/blockchain.rs
```diff
@@ -166,6 +166,16 @@ impl<D: Db, T: TransactionTrait> Blockchain<D, T> {
     self.provided.provide(tx)
   }
 
+  pub(crate) fn complete_provided(&mut self, order: &str, hash: [u8; 32]) {
+    let mut txn = self.db.as_mut().unwrap().txn();
+    self.provided.complete(&mut txn, order, hash);
+    txn.commit();
+  }
+
+  pub(crate) fn provided_transactions(&self) -> Vec<u8> {
```
kayabaNerve: While it's irrelevant, due to my above commentary on how this *should* be handled, we shouldn't return a raw byte buffer with no meaning no context. The Tributary APIs should be self contained and safe, meaning this should've at least been a `Vec<[u8; 32]>`.

coordinator/tributary/src/blockchain.rs
```diff
@@ -217,7 +227,11 @@ impl<D: Db, T: TransactionTrait> Blockchain<D, T> {
     commit: Vec<u8>,
     schema: N::SignatureScheme,
   ) -> Result<(), BlockError> {
-    self.verify_block::<N>(block, schema)?;
+    // verify the block, ignore the NonLocalProvided error.
+    let res = self.verify_block::<N>(block, schema);
+    if res.as_ref().is_err_and(|e| !matches!(e, BlockError::NonLocalProvided(_))) {
```
kayabaNerve: This is unsafe as a `NonLocalProvided` error may be raised midway through. If it is, the block would pass as valid here, despite half of the checks not actually executing.

We need to pass in a flag to not perform these checks specifically, yet still perform all other checks, and check that doesn't have any error.

coordinator/tributary/src/blockchain.rs
```diff
@@ -248,9 +262,9 @@ impl<D: Db, T: TransactionTrait> Blockchain<D, T> {
 
     for tx in &block.transactions {
       match tx.kind() {
-        TransactionKind::Provided(order) => {
-          self.provided.complete(&mut txn, order, tx.hash());
-        }
+        // We used to call the provided.complete() here
+        // but that is now moved to coordinator scanner.
```
kayabaNerve: The `coordinator` should not be acknowledge by the `tributary`, a self-contained piece of code.

coordinator/tributary/tendermint/src/lib.rs
```diff
@@ -796,6 +796,9 @@ impl<N: Network + 'static> TendermintMachine<N> {
       if self.block.log.has_consensus(self.block.round().number, Data::Prevote(Some(block.id()))) {
         match self.network.validate(block).await {
           Ok(_) => (),
+          // BlockError::Temporal is due to us not having the locally provided tx yet,
```
kayabaNerve: Tendermint should not acknowledge Tributary.

Though yes, the logic

"If there's a temporal error yet no one else has it, we assume it's our fault"

is perfectly sound :)

kayabaNerve (changes requested):

kayabaNerve:
Will merge upon CI passing :)

Issue #351, opened 2023-08-21
Monero CI failure


kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/5923190999/job/16058367418

https://github.com/serai-dex/serai/blob/34397b31b1a77e9924ba407649768eb79a7b4590/tests/processor/src/lib.rs#L287 returned invalid node.

This can only happen if:

1) The Monero response wasn't UTF-8.
2) It was an error variant (which it shouldn't be since we're getting a block it gave us the hash for).
3) Our parsing code has an error.
4) Our hash code has an error.

Unfortunately, we do not have the granularity to know what the answer is.

kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/6378881560/job/17310298162?pr=388, same spot.

kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/6377392652/job/17305972503 same spot, different error. Since it was `ConnectionError`, I initially thought the node simply hadn't booted before the test ran, but no, prior calls executed.

Maybe it's some issue with long-lived Monero connections? If I run the binary to reserialize the blockchain for long enough, I eventually get similar behavior...

Boog900:
I'm getting a similar issue in Cuprate every now and then I get error requesting a block or transaction, no other RPC request is giving me issues. I don't think it's because of long-lived connections as I was holding the connections for a while without requesting any blocks with no issues but when I start requesting blocks/ txs I start getting issues.

I did a bit of investigating exposed some errors and found that [here](https://github.com/serai-dex/serai/blob/4ee65ed243afb7ed629b04b3c1bbbd53df9b665f/coins/monero/src/rpc/mod.rs#L150) keeps erroring with `expected value at line 1 column 1`

At the moment I have no clue why

Boog900:
well I found my issue:
```
<html>
<head><title>429 Too Many Requests</title></head>
<body>
<center><h1>429 Too Many Requests</h1></center>
<hr><center>nginx</center>
</body>
</html>
```
guessing this is unrelated to the orignal issue

kayabaNerve:
If it's from `nginx`, yeah, it's unrelated to the CI failures. I wouldn't be surprised if the Monero node has some similar bs causing out own ConnectionError/InvalidNode reports though.

Our RPC code really needs a large rewrite re: error handling...

kayabaNerve:
[...]

PR #388, opened 2023-10-02
- Added db macro

econsta:
- implemented db macro for KeyGenDb

Here is our implementation of the db macro for review, as well as a modified key_gen.rs file that uses it. We are using bincode instead of SCALE right now, as we ran into two problems when we tried to use SCALE. First HashMap is not compatible with SCALE, but it used in Commitments. Second even when we made a work around for this, we ran into a ``overflow evaluating the requirement`` problem that we could not find a work around for. Here is an issue we think might have been related as it has the exact same error message we got: https://github.com/paritytech/scale-info/issues/65. Let us know what you would like us to do moving forward, if you think this is sufficient we can start implementing it in other places in the code base.


kayabaNerve:
The idea LGTM :)

As for the PR itself,

1) `createDb` should be `create_db`, per Rust style.
2) Instead of `&impl serde::Serialize`, it probably makes sense to require the field type with the field name in the macro usage. This would replace `ParamsDb::get::<ThresholdParams>` with `ParamsDb::get`. Ideally, we also ensure the key has a consistent type between `get` and `set`. To that end, it may make sense to replace keys (`impl AsRef<[u8]>` with `K: Encode + Decode` (or `bincode` again for consistency)).
3) It probably makes sense to impl `read_keys` directly into `GeneratedKeysDb`, if I'm not missing a scoping issue.

This definitely should remove a lot of boiler-plate though, so it's much appreciated ^_^ Thank you.

kayabaNerve:
Also, what was the issue with `cargo +nightly fmt` causing the commit to be reverted?

davidjohnbell:
For some reason more tests failed after we had formatted

kayabaNerve:
The two actual tests failing here aren't your fault. One is #351, the other was intermittent a week ago as it assumes network latency properties GH's CI likes to prove invalid to assume.

clippy and fmt are on you though. I'll look at the other failed runs from the prior commits...

davidjohnbell:
I see, thank you for the heads up. econsta and I are new to OSS I was meaning to ask you if making a PR is the proper protocol here, or is there some other means by which we show you our progress on an issue?

kayabaNerve:
The fmt commit had the same issue as #351, just with a different error 0_o I'll update the meta for it.

Making a PR, and pushing updates to it, is definitely the best way to go! Feel free to chime in on Discord as well, if you prefer a quicker response.

Sorry the inconsistent tests threw you off. Since you're editing the DB code, every service which uses a DB has all of their tests run. The larger test suites run 10+ Docker containers at once and I'm still working on resolving their stability, as shown here.

kayabaNerve:
What was the reason for closing this?

PR #397, opened 2023-10-13
Alternate handover batch TOCTOU fix

kayabaNerve:
This may be a cleaner impl and accordingly preferable.


kayabaNerve:
This PR is ~2x as large, yet doesn't sprawl out over several locations. Instead, it's contained to the coordinator. I find it preferable accordingly.

Issue #337, opened 2023-08-06
serai-client rewrite


kayabaNerve:
serai-client has three issues right now:

1) Inefficient RPC routes (`get_block_hash(u64)` doesn't exist because we only return finalized blocks, so instead we offer the full `get_block_by_number(u64)`)
2) Lack of scoping. Each section shares a namespace under the global Serai client. Like subxt, we should use wrapper structs to diversify.
3) Only operating on finalized blocks. While this is intended behavior, and good for Serai, we need to offer a mode which does expose un-finalized blocks. Ideally, finalized/unfinalized is just a config option on construction.

Most of this can be done by re-organizing serai-client yet some will require adding RPC routes to our node.

kayabaNerve:
We also should look at removing subxt entirely, just as we're looking at removing metadata.

kayabaNerve:
Part of the goal for this rewrite should also be making the client MIT licensed, which means dropping all AGPL dependencies. See https://github.com/serai-dex/serai/issues/13#issuecomment-1384826126 for info on what this would look like.

kayabaNerve:
I'm fine closing this issue after https://github.com/serai-dex/serai/commit/c511a54d18453814719155b30777c99bcf355f7f, which moved serai-client to MIT.

Remaining tweaks and nits are part of #474.

PR #398, opened 2023-10-14
serai-client reorganization

kayabaNerve:
Instead of functions taking a block hash, has a scope to a block hash before functions can be called.

Separates functions by pallets.


kayabaNerve:
Relevant #337.

kayabaNerve:
Bah. I think I hit rebase instead of squash by accident...

PR #399, opened 2023-10-16
Coins pallet

akildemir:
Replaces `pallet_tokens`, `pallet_balances` and `tokens_pallet` with a custom single `coins_pallet` .


substrate/client/src/serai/coins.rs
```diff
@@ -22,32 +20,29 @@ impl<'a> SeraiCoins<'a> {
...
-    > = self.0.storage("System", "Account", Some(vec![scale_value(address)])).await?;
-    Ok(data.map(|data| data.data.free).unwrap_or(0))
+    let amount: Option<SubstrateAmount> = self
+      .0
+      .storage(PALLET, "Coins", Some(vec![scale_value(address), scale_value(Coin::Serai)]))
+      .await?;
+    Ok(amount.unwrap_or(0))
   }
 
+  // TODO: shouldn't these token functions be renamed to coins now?
```
kayabaNerve: Yep. We should remove `sri_` specific methods and solely have coin_*method* methods (which can be passed Coin::Serai as an argument).

substrate/coins/pallet/Cargo.toml
```diff
@@ -0,0 +1,47 @@
+[package]
+name = "serai-coins-pallet"
+version = "0.1.0"
+description = "Coins pallet for Serai"
+license = "AGPL-3.0-only"
+repository = "https://github.com/serai-dex/serai/tree/develop/substrate/staking/pallet"
+authors = ["Luke Parker <lukeparker5132@gmail.com>"]
```
kayabaNerve: You should be the author, actually :p

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  pub enum Error<T> {
+    AmountOverflowed,
+    NotEnoughFunds,
+  }
+
+  #[pallet::event]
+  #[pallet::generate_deposit(fn deposit_event)]
+  pub enum Event<T: Config> {
+    Mint { address: SeraiAddress, balance: Balance },
+    Burn { address: SeraiAddress, balance: Balance, instruction: OutInstruction },
```
kayabaNerve: Do we have an `OutInstructionWithBalance`? We do for `InInstruction`.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    SriBurn { address: SeraiAddress, amount: Amount },
+    Transfer { from: SeraiAddress, to: SeraiAddress, balance: Balance },
+  }
+
+  #[pallet::pallet]
+  pub struct Pallet<T>(PhantomData<T>);
+
+  /// The amount of funds each account has.
```
kayabaNerve: "amount of coins"

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  #[pallet::storage]
+  #[pallet::getter(fn coins)]
+  pub type Coins<T: Config> = StorageDoubleMap<
+    _,
+    Blake2_128Concat,
+    T::AccountId,
+    Identity,
```
kayabaNerve, line 67: We should have a comment justifying the use of Identity due to Coin being a non-manipulatable fixed-space ID.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  #[pallet::genesis_build]
+  impl<T: Config> BuildGenesisConfig for GenesisConfig<T> {
+    fn build(&self) {
+      // initialize the supply of the coins
+      for c in COINS.iter() {
+        Supply::<T>::set(c, 0);
+      }
+
+      // initialize the genesis accounts
+      for (account, coin, amount) in self.accounts.iter() {
```
kayabaNerve: Use the Balance type?

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        Coins::<T>::set(account, coin, Some(*amount));
+        let current_supply = Supply::<T>::get(coin);
+        Supply::<T>::set(coin, current_supply.checked_add(*amount).unwrap());
```
kayabaNerve: These lines should be replaced with `mint`. They won't emit events right now *and* overwrite balance due to non-usage of `mint`.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    Transfer { from: SeraiAddress, to: SeraiAddress, balance: Balance },
+  }
+
+  #[pallet::pallet]
+  pub struct Pallet<T>(PhantomData<T>);
+
+  /// The amount of funds each account has.
+  #[pallet::storage]
+  #[pallet::getter(fn coins)]
+  pub type Coins<T: Config> = StorageDoubleMap<
```
kayabaNerve: Would `Balances` be a better name for this? It's not a StorageValue of all active Coins.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  impl<T: Config> Pallet<T> {
+    /// Returns the balance of a given account for `coin`.
+    pub fn balance(of: &T::AccountId, coin: Coin) -> SubstrateAmount {
+      Self::coins(of, coin).unwrap_or(0)
+    }
+
+    fn increase_supply(coin: Coin, by: SubstrateAmount) -> Result<(), Error<T>> {
+      // update the supply
+      let new_supply = Self::supply(coin).checked_add(by).ok_or(Error::<T>::AmountOverflowed)?;
+      Supply::<T>::set(coin, new_supply);
```
kayabaNerve: Shouldn't we inline this into mint? I'm unsure why we'd have another caller.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      Ok(())
+    }
+
+    fn decrease_supply(coin: Coin, by: SubstrateAmount) -> Result<(), Error<T>> {
```
kayabaNerve: Saqme comment as increase_supply.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      // update the supply
+      let new_supply = Self::supply(coin).checked_sub(by).ok_or(Error::<T>::AmountOverflowed)?;
+      Supply::<T>::set(coin, new_supply);
+      Ok(())
+    }
+
+    /// Mints amount at the given account, errors if amount overflows.
+    pub fn mint(at: &T::AccountId, coin: Coin, amount: SubstrateAmount) -> Result<(), Error<T>> {
```
kayabaNerve: at -> to?

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      // don't waste time if amount 0.
+      if amount == 0 {
+        return Ok(());
+      }
```
kayabaNerve: Ironically enough, I think this check will waste more time than not having this check would. This'll be called by emissions, which should be non-zero, and InInstructions. The processor will drop anything that's dust (such as 0 transfers).

You're welcome to keep it as an idea of good practice though.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      Coins::<T>::set(at, coin, Some(new_amount));
+      Self::increase_supply(coin, amount)?;
+
+      let balance = Balance { coin, amount: Amount(amount) };
+      Self::deposit_event(Event::Mint { address: SeraiAddress(at.0), balance });
+      Ok(())
+    }
+
+    /// Burns amount at the given account, errors if not enough funds to burn.
+    pub fn burn_internal(
```
kayabaNerve: Shouldn't be pub.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  impl<T: Config> Default for GenesisConfig<T> {
+    fn default() -> Self {
+      GenesisConfig { accounts: Default::default() }
+    }
+  }
+
+  #[pallet::error]
+  pub enum Error<T> {
+    AmountOverflowed,
+    NotEnoughFunds,
```
kayabaNerve: NotEnoughCoins

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+      // sub amount from account
+      let new_amount =
+        Self::coins(at, coin).unwrap_or(0).checked_sub(amount).ok_or(Error::<T>::NotEnoughFunds)?;
+
+      if new_amount == 0 {
+        Coins::<T>::remove(at, coin);
+      } else {
+        Coins::<T>::set(at, coin, Some(new_amount));
+      }
```
kayabaNerve: decrease_balance_internal to share this with transfer? Same commentary for increase_balance_internal.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      // something should have gone horribly wrong if we error on this.
+      if Self::decrease_supply(coin, amount).is_err() {
+        panic!("we tried to burn more assets than what we have");
+      }
+
+      if coin == Coin::Serai {
+        Self::deposit_event(Event::SriBurn { address: SeraiAddress(at.0), amount: Amount(amount) });
+      } else {
+        let balance = Balance { coin, amount: Amount(amount) };
+        // TODO: error for no instruction?
```
kayabaNerve: I'd move this `if` statement into the caller so `burn_internal` is InInstruction-agnostic and we don't have this potential error case (even if our static code never makes it reachable). Then we'd offer burn and burn_external, one not taking an instruction, one taking.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    #[pallet::call_index(1)]
+    #[pallet::weight((0, DispatchClass::Normal))] // TODO
+    pub fn burn(
+      o: OriginFor<T>,
+      balance: Balance,
+      instruction: Option<OutInstruction>,
+    ) -> DispatchResult {
+      let at = ensure_signed(o)?;
```
kayabaNerve: I'd still call this from.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      Ok(())
+    }
+  }
+
+  #[pallet::call]
+  impl<T: Config> Pallet<T> {
+    #[pallet::call_index(0)]
+    #[pallet::weight((0, DispatchClass::Normal))] // TODO
+    pub fn transfer(
+      o: OriginFor<T>,
```
kayabaNerve: This `o` gets compiled into an enum member (`Pallet::Call::transfer { o: OriginFor<T> }`). It'd be polite to call it `origin`-proper.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+      instruction: Option<OutInstruction>,
+    ) -> DispatchResult {
+      let at = ensure_signed(o)?;
+      Self::burn_internal(&at, balance.coin, balance.amount.0, instruction)?;
+      Ok(())
+    }
+  }
+
+  #[derive(Default)]
+  pub struct Imbalance {
```
kayabaNerve: Is there a reason not to use SubstrateAmount/Option<SubstrateAmount> directly?

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  impl<T: Config> OnChargeTransaction<T> for Pallet<T> {
+    type Balance = SubstrateAmount;
+    type LiquidityInfo = Option<Imbalance>;
+
+    /// Before the transaction is executed the payment of the transaction fees
+    /// need to be secured.
+    ///
+    /// Note: The `fee` already includes the `tip`.
```
kayabaNerve: These comments, directly copied from the trait docs, should not be here.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+    fn withdraw_fee(
+      who: &T::AccountId,
+      _call: &T::RuntimeCall,
+      _dispatch_info: &DispatchInfoOf<T::RuntimeCall>,
+      fee: Self::Balance,
+      _tip: Self::Balance,
+    ) -> Result<Self::LiquidityInfo, TransactionValidityError> {
+      // TODO: Could we have a 0 fee tx? Error in this case?
```
kayabaNerve: Yeah, I'd just return Ok as you do.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+        Err(_) => Err(InvalidTransaction::Payment.into()),
+        Ok(()) => Ok(Some(Imbalance { amount: fee })),
+      }
+    }
+
+    /// After the transaction was executed the actual fee can be calculated.
+    /// This function should refund any overpaid fees and optionally deposit
+    /// the corrected amount.
+    ///
+    /// Note: The `fee` already includes the `tip`.
```
kayabaNerve: Same comment as I gave to the prior comment.

substrate/coins/primitives/src/lib.rs
```diff
@@ -12,7 +12,7 @@ use scale_info::TypeInfo;
 
 use serai_primitives::{Balance, SeraiAddress, ExternalAddress, Data, pallet_address};
 
-pub const ADDRESS: SeraiAddress = pallet_address(b"Tokens");
+pub const FEE_ACCOUNT: SeraiAddress = pallet_address(b"FeeAccount");
```
kayabaNerve: Should we rename pallet_address to system_address?

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -73,10 +73,14 @@ pub mod pallet {
   impl<T: Config> Pallet<T> {
     fn execute(instruction: InInstructionWithBalance) -> Result<(), ()> {
       match instruction.instruction {
-        InInstruction::Transfer(address) => Tokens::<T>::mint(address, instruction.balance),
+        InInstruction::Transfer(address) => {
+          // TODO: pass balance directly here instead of coin and amount?
+          // but coins pallet deals only with SubstrateAmount and Balance type has Amount type.
+          Coins::<T>::mint(&address.into(), instruction.balance.coin, instruction.balance.amount.0)
```
kayabaNerve: Pass Balance, yeah

substrate/node/src/chain_spec.rs
```diff
@@ -34,24 +34,10 @@ fn testnet_genesis(
...
-        .iter()
-        .map(|coin| (*coin, TOKENS_ADDRESS.into(), true, 1))
-        .collect(),
-      metadata: vec![
-        (Coin::Bitcoin, b"Bitcoin".to_vec(), b"BTC".to_vec(), 8),
-        // Reduce to 8 decimals to feasibly fit within u64 (instead of its native u256)
-        (Coin::Ether, b"Ether".to_vec(), b"ETH".to_vec(), 8),
-        (Coin::Dai, b"Dai Stablecoin".to_vec(), b"DAI".to_vec(), 8),
-        (Coin::Monero, b"Monero".to_vec(), b"XMR".to_vec(), 12),
-      ],
```
kayabaNerve, line 53: Should we move these to a few new functions under the Coin enum?

substrate/runtime/src/lib.rs
```diff
@@ -217,7 +201,7 @@ impl system::Config for Runtime {
   type OnKilledAccount = ();
   type OnSetCode = ();
 
-  type AccountData = balances::AccountData<SubstrateAmount>;
+  type AccountData = SubstrateAmount;
```
kayabaNerve: This should be `()` now, I believe.

substrate/coins/pallet/src/lib.rs
```diff
@@ -0,0 +1,314 @@
...
+
+      if coin == Coin::Serai {
+        Self::deposit_event(Event::SriBurn { address: SeraiAddress(at.0), amount: Amount(amount) });
+      } else {
+        let balance = Balance { coin, amount: Amount(amount) };
+        // TODO: error for no instruction?
+        Self::deposit_event(Event::Burn {
+          address: SeraiAddress(at.0),
+          balance,
+          instruction: instruction.unwrap(),
```
kayabaNerve: This isn't valid. burn is user-facing. A user can call burn, calling burn_internal, with a non-Serai coin and None for the instruction, causing a panic.

kayabaNerve (commented):
Largely lints :+1:

substrate/coins/pallet/src/lib.rs
akildemir: yes. The only requirement for `OnChargeTransaction::LiquidityInfo` type is that should be something that implements `Default` trait and can't derive `Default` on plain u64. Hence why it is wrapped into a struct.

substrate/coins/primitives/src/lib.rs
akildemir: makes sense

substrate/coins/pallet/src/lib.rs
kayabaNerve: Ah, then Option<u64> should work.

kayabaNerve:
CI failure is unrelated and fixed in https://github.com/serai-dex/serai/commit/041ed4617112fd350e09744b380b8e89fb5c62e8

PR #404, opened 2023-10-20
fix key fn not being called in getter

econsta:
Fixed this line of code in the multisigs/db.rs file
```rust
pub fn resolved_plan<G: Get>(
    getter: &G,
    tx: <N::Transaction as Transaction<N>>::Id,
  ) -> Option<[u8; 32]> {
    getter.get(Self::resolved_key(tx.as_ref())).map(|id| id.try_into().unwrap())
}
```


kayabaNerve:
Thanks for noticing and fixing this :)

CI failures are unrelated and due to the most recent commit. I'll fix that now...

PR #407, opened 2023-10-26
Add Dex pallet

akildemir:


kayabaNerve:
Supersedes #363.

kayabaNerve:
Will merge as-is as this is:

1) Massive
2) Solid
3) A blocker

We need to follow up with:
1) Removing traits, as we're not testing against pallet-balances. Accordingly, we can just use coins-pallet directly.
2) Making liquidity tokens an instance of coins to reduce duplicated code and increase liquidity token functionality as needed for our future work.
3) Moving impl of DEX traits for coins-primitives from coins-primitives to dex-primitives. 
4) Simplifying dex-pallet to our usecase (no coin1, coin2, just Coin::Serai and coin, LPToken and PoolId set to Coin...)

PR #408, opened 2023-10-27
Database Macro

### LINKED ISSUE #364: Macro'd DBs
The Serai codebase is full of DB code, all with common patterns. Wrapping a generic `D: serai_db::Db`, using length-prefixed keys...

1) We should have a macro for keys, making our handling of them defined in a singular location, ideally with a DST-overlap check.

2) We should define static-set methods. Any item which is write-once, only has a single possible value barring invariants, and not an issue to be written earlier than expected should be written immediately *without a commit* with write re-attempts checking consistency. This helps prevent race conditions where it's written to a TXN, and something tries to read it, yet the TXN has yet to be committed.

It is a bit ridiculous to add this pattern (arguably itself an anti-pattern, as it may cause writes when the data isn't expected to be written yet) just to hedge against a security issue. Currently, we have several such DB entries though where we're unaffected by if it's written. The pattern is generally

```
if x {
  write y
  do z
}
```

where `x` is regardless of if `y` was written.

My main goal isn't necessarily to hedge against any anti-patterns, and it definitely isn't to introduce an anti-pattern. It's to formalize this type of data and define handling of it, given it's frequency.

3) Gets which don't internally panic.

We currently panic in several getters as their callers should only call them on existence. This means the method puts assumptions on the caller. The caller should take that responsibility by being returned an Option and calling expect itself.

4) Gets which use `<G: Get>`, not `self`, `db`, or `txn` explicitly.

---

Affects message-queue, processor, and coordinator.

davidjohnbell:
### Description

This pull request closes #364. The Serai code-base is full of repetitive database code used to namespace keys. The provided macro generates the database and item DSTs necessary for a new key. The macro supports zero or more `scale` encodable arguments in the generation of the key and will likewise automatically `bincode` serialize/de-serialize the database value resulting in better DX and hopefully less bugs. 

### Considerations

- The macro outputs an empty struct so that methods can be added using an `impl` block
- `bincode` was chosen over scale because it supports a wider range of types. It's still possible to store `scale` bytes in the form of a `Vec<u8>` and transform later as is the case now
- Many of the existing databases are generic (e.g., `Network`) which can't be auto generated by the macro. Moving forward, helper methods should receive those generic arguments instead.

### Example
```rust
create_db!(
  KeyGenDb {
    ParamsDb: (vs: &ValidatorSet) -> ThresholdParams,
    CommitmentsDb: (key: &KeyGenId) -> HashMap<Participant, Vec<u8>>,
    GeneratedKeysDb: (set: &ValidatorSet, key_one: &[u8; 32], key_two: &[u8]) -> Vec<u8>,
    KeysDb: (key: &Vec<u8>) -> Vec<u8>
  }
);
```
In this case 3 databases are created, but let's focus on `ParamsDB`. `ParamsDB` gets a "KeyGenDB" database DST and a "ParamsDB" item DST, whose `get` and `set` become:
```rust
impl ParamsDB {
  pub fn get(getter: &impl Get, vs: &ValidatorSet) -> Option<ThresholdParams> {
    // brevity
  }
  pub fn set(txn: &mut DbTxn, vs: &ValidatorSet, value: ThresholdParams) {
    //brevity
  }
}
```

### Future Work
The pull request is being opened now so that future work builds off of this macro while we convert old databases and pass tests.

### DIFF
--- a/common/db/src/create_db.rs
+++ b/common/db/src/create_db.rs
@@ -1,4 +1,9 @@
-pub fn db_key(db_dst: &'static [u8], item_dst: &'static [u8], key: impl AsRef<[u8]>) -> Vec<u8> {
+#[doc(hidden)]
+pub fn serai_db_key(
+  db_dst: &'static [u8],
+  item_dst: &'static [u8],
+  key: impl AsRef<[u8]>,
+) -> Vec<u8> {
   let db_len = u8::try_from(db_dst.len()).unwrap();
   let dst_len = u8::try_from(item_dst.len()).unwrap();
   [[db_len].as_ref(), db_dst, [dst_len].as_ref(), item_dst, key.as_ref()].concat()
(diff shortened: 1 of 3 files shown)


common/db/src/create_db.rs
```diff
@@ -22,7 +22,7 @@ pub fn db_key(db_dst: &'static [u8], item_dst: &'static [u8], key: impl AsRef<[u
 ///
 /// # Example
 ///
-/// ```
+/// ```no_run
```
kayabaNerve: Why not simply correct the missing import?

davidjohnbell: I believe the doc test is compiled in the current scope, but those types I use aren't going to be available, as is the case with the `bincode` dependency - so correcting the missing import would just lead to more errors :sweat_smile:

davidjohnbell: sorry if I was spamming CI :blush:

kayabaNerve: Ha, don't worry. We're not on any paid plan.

PR #409, opened 2023-10-29
Implement epoch trigger for Babe pallet

akildemir:


akildemir:
Closed since it won't be needed after https://github.com/serai-dex/serai/pull/440 .

PR #417, opened 2023-11-05
Adopt nextest

kayabaNerve:
Applied where it makes sense to do so.

### DIFF
--- a/.github/actions/build-dependencies/action.yml
+++ b/.github/actions/build-dependencies/action.yml
@@ -54,5 +54,9 @@ runs:
         components: ${{ inputs.rust-components }}
         targets: wasm32-unknown-unknown, riscv32imac-unknown-none-elf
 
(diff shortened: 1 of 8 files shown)


kayabaNerve:
Will confirm the CI runs and get some loose benchmarks on if there's benefit to this.

kayabaNerve:
I was stupid enough to let this run without manually triggering the relevant runs (which this didn't default to running). Will try to do so now, though I'm not immediately sure I can run a PR's workflow on a PR.

kayabaNerve:
Confirmed the newly queued runs have nextest installation in their logs.

kayabaNerve:
This had multiple test failures on the tests which spawn anything heavier than a message-queue Docker container. On the CIs which weren't so heavy, this wasn't impactful. Closing for now.

PR #421, opened 2023-11-08
Use lld on x86-64 Linux

kayabaNerve:
There does not appear to be blockers for this per
https://github.com/rust-lang/rust/issues/71515.


kayabaNerve:
Neither CI nor my local computer observed any notable performance benefit to the point I want to muck with this at this time.

PR #423, opened 2023-11-10
Convert coordinator/tributary/nonce_decider to use create_db macro

davidjohnbell:
### Description

This is a continuation of pr #408. `NonceDecider` was split into `NextNonceDb` and `ItemNonceDb`, some helper methods to convert to/from the expected type were removed as they're covered by the macro.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7578,6 +7578,7 @@ dependencies = [
  "serai-env",
  "serai-message-queue",
  "serai-processor-messages",
+ "serde",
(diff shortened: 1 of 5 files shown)


kayabaNerve:
Rebased and restored "NonceDecider" so we don't have a "Db" (presumed internal to some area) floating around with higher-level logic (really just a stylistic preference of mine). Will merge once CI passes :) Thanks for working on this. Definitely much cleaner now :D

Issue #358, opened 2023-08-24
Make which processor is built by Docker configurable


kayabaNerve:
Right now, we have a single `processor` Docker image which is built with `--all-features`. We should only build with the feature for the network it'll be deployed with. This way, if a dependency exclusive to the `bitcoin` library is compromised, we only have the Bitcoin processor compromised, not all of them.

Related #317.

kayabaNerve:
#375 would be a great solution to this.

PR #428, opened 2023-11-12
Dockerfile Parts

### LINKED ISSUE #375: Use a Makefile to generate Dockerfiles
They share a lot of common code, yet unfortunately, Docker doesn't offer a preprocessor, or even a way to refer to local images defined outside a file. Accordingly, we shouldn't write Dockerfiles, solely scripts which generate Dockerfiles.

A Makefile or bash script should be more than fine here.

kayabaNerve:
Breaks Dockerfiles into several commonly shared parts and uses a bash script to create the actual Dockerfiles.

Makes `sh` a dev dependency for working with the Dockerfiles.

Resolves #375.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -66,3 +66,11 @@ jobs:
 
       - name: Run rustfmt
         run: cargo +${{ steps.nightly.outputs.version }} fmt -- --check
+
+  dockerfiles:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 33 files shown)


kayabaNerve:
Will merge once/if CI passes. I'm still locally going through building all of the Docker images so this may not actually work quite yet.

kayabaNerve:
Added fix for #358

PR #433, opened 2023-11-15
Add a cosigning protocol to ensure finalizations are unique

### LINKED ISSUE #339: Theft of external coins via equivocation
Under sharded validator sets, the Serai validator set has been prior been defined as not having/needing a bond requirement as their role is liveliness and ordering, not necessarily security. Since they can equivocate however, finalizing two distinct chains, they can double spend Serai letting them swap to BTC and to ETH without restriction.

We could add a bond requirement to Serai itself, or halt if an equivocating commit (not just precommit) is detected, with some latency window to allow detection (relevant #281). The clean solution is on finalized block, each validator set produces a threshold signature for it. A validator set would only act on it once a supermajority of validator sets signs off on it.

This would require an equivocating Serai validator set to also have an equivocating coin validator set (with 34% of the coin-bonded Serai, to tip a split 33%/33% to 67%/67%).  This... may be fine? Should require 67%, not 34%?

We can additionally say we halt if any other validator set fails to produce a signature, or produces a signature for a different block, yet this enables various DoS angles. The former more reachable than the latter. We'd also still have some latency assumption.

At the least, we want to require all validator sets co-sign finalized blocks. Then we likely still want a halt system upon equivocation detection, with a dynamically latent buffer based on volume.

kayabaNerve:
Resolves https://github.com/serai-dex/serai/issues/339.

Now, if the Serai validators equivocate, they also need to perfectly split the validator sets for external networks and additionally have 34% of them equivocate.

Implemented as a completely off-chain protocol with evaluations linear to the amount of sets (and cosigning linear to the size of the set signing).

### DIFF
--- a/coordinator/Cargo.toml
+++ b/coordinator/Cargo.toml
@@ -37,6 +37,7 @@ processor-messages = { package = "serai-processor-messages", path = "../processo
 message-queue = { package = "serai-message-queue", path = "../message-queue" }
 tributary = { package = "tributary-chain", path = "./tributary" }
 
+sp-application-crypto = { git = "https://github.com/serai-dex/substrate", default-features = false, features = ["std"] }
 serai-client = { path = "../substrate/client", default-features = false, features = ["serai"] }
 
 hex = { version = "0.4", default-features = false, features = ["std"] }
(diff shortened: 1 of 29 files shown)


kayabaNerve:
This branch also resolves a potential (under very exact timing conditions) deadlock around how we handle Batch verification/publication, as we had to replace the old system which *always* deadlocked in a cosigning environment.

kayabaNerve:
I noticed a potential panic on rotation when reviewing the PR. If the above CI passes, I'll push the two line edit and merge without waiting another 3h for the CI again.

PR #436, opened 2023-11-16
Convert coordinator/substrate/db to use create_db macro

davidjohnbell:
### Description

This is a continuation of pr #408. `SubstrateDb` was split into `BlockDb`,  `EventDb`, `SessionDb`, and `BatchDb`, some helper methods to convert to/from the expected type were removed as they're covered by the macro. 

### Considerations
- Helper function `handled` was renamed to `is_unhandled`. Calling code was largely in the form `if !SubstrateDb::handled(...) { ... }`, removing the negation and changing the name helps align the code with the intent.
- `handle_event` _could_ be refactored into a simple set method, I noticed that the calling code often performs the necessary checks *however* I didn't feel comfortable making this change due to some `async` statements that could break the previous assertions.

### DIFF
--- a/coordinator/src/substrate/db.rs
+++ b/coordinator/src/substrate/db.rs
@@ -8,14 +8,14 @@ use serai_client::{
 pub use serai_db::*;
 
 create_db!(
-  CoordinatorSubstrateDb {
+  SubstrateDb {
     CosignTriggered: () -> (),
     IntendedCosign: () -> (u64, Option<u64>),
     BlockHasEvents: (block: u64) -> u8,
     LatestCosignedBlock: () -> u64,
-    NextBlock: () -> u64,
+    BlockDb: () -> u64,
     EventDb: (id: &[u8], index: u32) -> (),
-    SessionDb: (key: &[u8]) -> Session,
+    NextBlock: () -> u64,
     BatchDb: (network: NetworkId, id: u32) -> [u8; 32]
   }
 );
@@ -44,12 +44,12 @@ impl EventDb {
 
   pub fn handle_event(txn: &mut impl DbTxn, id: &[u8], index: u32) {
     assert!(Self::is_unhandled(txn, id, index));
-    Self::set(txn, id, index, &[0u8; 0]);
+    Self::set(txn, id, index, &());
   }
 }
 
 db_channel! {
-  SubstrateDb {
+  SubstrateDbChannels {
     CosignTransactions: (network: NetworkId) -> (Session, u64, [u8; 32]),
   }
 }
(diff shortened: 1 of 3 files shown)


coordinator/src/substrate/db.rs
```diff
@@ -9,12 +9,38 @@ use serai_client::{
...
+create_db!(
+  CoordinatorSubstrateDb {
     CosignTriggered: () -> (),
     IntendedCosign: () -> (u64, Option<u64>),
     BlockHasEvents: (block: u64) -> u8,
-    CosignTransactions: (network: NetworkId) -> Vec<(Session, u64, [u8; 32])>
+    LatestCosignedBlock: () -> u64,
+    CosignTransactions: (network: NetworkId) -> Vec<(Session, u64, [u8; 32])>,
+    BlockDb: () -> u64,
+    EventDb: (id: &[u8], index: u32) -> [u8; 0],
```
kayabaNerve: `-> ()`

coordinator/src/substrate/db.rs
```diff
@@ -9,12 +9,38 @@ use serai_client::{
...
+
+create_db!(
+  CoordinatorSubstrateDb {
     CosignTriggered: () -> (),
     IntendedCosign: () -> (u64, Option<u64>),
     BlockHasEvents: (block: u64) -> u8,
-    CosignTransactions: (network: NetworkId) -> Vec<(Session, u64, [u8; 32])>
+    LatestCosignedBlock: () -> u64,
+    CosignTransactions: (network: NetworkId) -> Vec<(Session, u64, [u8; 32])>,
+    BlockDb: () -> u64,
```
kayabaNerve: Mind if we call this NextBlock?

coordinator/src/substrate/mod.rs
```diff
@@ -568,25 +568,25 @@ async fn handle_new_blocks<D: Db, Pro: Processors>(
     )
     .await?;
     *next_block += 1;
-    db.set_next_block(*next_block);
+    let mut txn = db.txn(); // TODO: this line belongs to a removed method, remove SubstrateDb
```
kayabaNerve: This TODO doesn't seem still needed.

kayabaNerve (commented):
All LGTM, minus a few minor nits :)

coordinator/src/substrate/db.rs
davidjohnbell: Looks good to me, I’m not that opinionated 😄

coordinator/src/substrate/mod.rs
davidjohnbell: I was thinking about making it a helper but decided against it. I’ll investigate tonight. But I think you’re correct, the comment is old and can be removed if you have a moment.

kayabaNerve:
:+1: Will merge once CI passes. Thanks :)

PR #437, opened 2023-11-16
processor/db.rs macro implimentation

econsta:
implements the db macro for processor/src/db.rs, it's pretty simple, the tests pass locally, should be easy to merge


processor/src/db.rs
```diff
@@ -8,44 +7,18 @@ pub use serai_db::*;
...
 use crate::networks::{Block, Network};
 
-#[derive(Debug)]
-pub struct MainDb<N: Network, D: Db>(D, PhantomData<N>);
-impl<N: Network, D: Db> MainDb<N, D> {
-  pub fn new(db: D) -> Self {
-    Self(db, PhantomData)
+create_db!(
+  MainDb {
+    HandledMessageDb: (id: u64) -> Vec<u8>,
```
kayabaNerve: I think this can be ().

processor/src/main.rs
```diff
@@ -352,7 +354,9 @@ async fn handle_coordinator_msg<D: Db, N: Network, Co: Coordinator>(
             // We can't set these keys for activation until we know their queue block, which we
             // won't until the next Batch is confirmed
             // Set this variable so when we get the next Batch event, we can handle it
-            MainDb::<N, D>::set_pending_activation(txn, block_before_queue_block, set, key_pair);
+            let mut buf = (set, key_pair).encode();
+            buf.extend(block_before_queue_block.as_ref());
+            PendingActivationsDb::set(txn, &buf);
```
kayabaNerve: Can we preserve this in a set_pending_activation? Right now, this raw schema is present in db.rs and main.rs.

processor/src/main.rs
```diff
@@ -387,8 +391,8 @@ async fn handle_coordinator_msg<D: Db, N: Network, Co: Coordinator>(
                 activation_number,
               )
               .await;
-
-              MainDb::<N, D>::clear_pending_activation(txn);
+              //clear pending activation
+              PendingActivationsDb::set(txn, &vec![] as &Vec<u8>);
```
kayabaNerve: There is a delete function available.

https://github.com/serai-dex/serai/blob/develop/common/db/src/lib.rs#L21

kayabaNerve:
Definitely simple enough. Thanks for submitting :)

processor/src/main.rs
econsta: Is this good to use? The original clear_bending_activation did simply add an empty array

processor/src/main.rs
econsta: Sure

processor/src/main.rs
kayabaNerve: ... huh. I have no idea why I did that. Sorry for the confusion there.

AFAICT, `[]` is interpreted as `None`, so there's no functional difference between the two. Just a comment we *should* delete it, but yeah, not a fault of this PR. Feel free to either handle it or not handle it accordingly :)

kayabaNerve:
Looks good! Thanks :)

PR #439, opened 2023-11-16
implement db macro for processor/substrate_signer

econsta:
implements create_db macro for processor/src/substrate_signer.rs. It passes tests locally.


kayabaNerve (commented):
LGTM, just needs to use `()` :) Thanks

PR #440, opened 2023-11-17
Integrate session pallet into validator-sets pallet

akildemir:
- Integrates `pallet-session` functionality from  validator sets pallet.
- Removes `pallet-session` and `pallet-authority-discovery` pallets completely from runtime.


substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -53,23 +64,17 @@ pub mod pallet {
 
   /// The current session for a network.
   ///
-  /// This does not store the current session for Serai. pallet_session handles that.
+  /// This does not store the current session for Serai.
```
kayabaNerve: `CurrentSession` does now contain the current session for Serai, AFAICT.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -53,23 +64,17 @@ pub mod pallet {
...
-  impl<T: Config> Pallet<T> {
-    pub fn session(network: NetworkId) -> Option<Session> {
-      if network == NetworkId::Serai {
-        Some(Session(pallet_session::Pallet::<T>::current_index()))
-      } else {
-        CurrentSession::<T>::get(network)
-      }
-    }
 
-    pub fn latest_decided_session(network: NetworkId) -> Option<Session> {
```
kayabaNerve, line 69: This function was removed, yet that removes the view into how the latest decided session for Serai is current + 1. I'm unsure if that view is still needed (it may have implications for the signals pallet).

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -53,23 +64,17 @@ pub mod pallet {
...
-      CurrentSession::<T>::get(network)
-    }
-  }
+  /// The current set of serai validators.
+  #[pallet::storage]
+  #[pallet::getter(fn serai_validators)]
+  pub type SeraiValidators<T: Config> =
+    StorageValue<_, BoundedVec<(Public, u64), ConstU32<{ MAX_KEY_SHARES_PER_SET }>>, ValueQuery>;
```
kayabaNerve: `Babe::authorities` is valid for this.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -106,7 +111,7 @@ pub mod pallet {
     #[inline]
     fn in_active_serai_set(account: Public) -> bool {
       // TODO: This is bounded O(n). Can we get O(1) via a storage lookup, like we do with InSet?
-      for validator in pallet_session::Pallet::<T>::validators() {
+      for (validator, _) in Self::serai_validators() {
```
kayabaNerve: Hm. Babe::is_member exists, yet it's also O(n) and not O(1). Possibly still best to outsource to it *with* a TODO of "TODO: is_member is internally O(n). Update Babe to use an O(1) storage lookup?"?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -326,20 +329,26 @@ pub mod pallet {
       while key_shares < u64::from(MAX_KEY_SHARES_PER_SET) {
         let Some((key, amount)) = iter.next() else { break };
 
-        InSet::<T>::set(Self::in_set_key(network, key), Some(()));
-        participants.push(key);
+        let these_key_shares = amount.0 / allocation_per_key_share;
+        InSet::<T>::set(Self::in_set_key(network, key), Some(these_key_shares));
+        participants.push((key, these_key_shares));
```
kayabaNerve, line 332: Why was InSet/participants so modified?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -372,6 +381,18 @@ pub mod pallet {
     NonExistentValidator,
   }
 
+  #[pallet::hooks]
+  impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    fn on_initialize(n: BlockNumberFor<T>) -> Weight {
+      if T::ShouldEndSession::should_end_session(n) {
+        Self::rotate_session();
+        T::BlockWeights::get().max_block
```
kayabaNerve, line 388: Why is this max_block? Was pallet-session really that cursed?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -649,6 +675,40 @@ pub mod pallet {
...
+          None,
+        ),
+        WeakBoundedVec::force_from(
+          next_validators.into_iter().map(|(id, w)| (BabeAuthorityId::from(id), w)).collect(),
+          None,
+        ),
+        Some(session),
+      );
+      Grandpa::<T>::new_session(
+        true,
```
kayabaNerve, line 701: How do we feel about hard-coding `changed` to true?

kayabaNerve (commented):
CI failure is related. Many of my comments may just require an explanation on my end. Seems overall straightforward though :)

CI logs:
```
2023-11-20T10:29:33.1158090Z thread 'tokio-runtime-worker' panicked at coordinator/src/substrate/mod.rs:81:12:
2023-11-20T10:29:33.1159310Z validator selected for set yet didn't have an allocation
2023-11-20T10:29:33.1160256Z note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2023-11-20T10:29:33.1161558Z [2023-11-20T10:19:25Z ERROR serai_coordinator] exiting the process due to a task panicking
```

substrate/validator-sets/pallet/src/lib.rs
akildemir: yes it does, This comment line should be removed.

substrate/validator-sets/pallet/src/lib.rs
akildemir: The latest decided session for serai is current + 1, however function was still returning the current for serai as well IIRC(hence why it was removed), so if it should to return the current + 1 for serai, yes it should be re-added.

substrate/validator-sets/pallet/src/lib.rs
akildemir: because we needed that info to pass the grandpa, babe pallets.

substrate/validator-sets/pallet/src/lib.rs
akildemir: yea lol. Open to suggestions..

substrate/validator-sets/pallet/src/lib.rs
akildemir: I mean it is there for "optimization", grandpa does nothing with the new set if `changed` is `false` .  But `changed` should be `true` even if the set keys are the same but only the allocations are changed.  So should be `true` most of the time anyway. 

We either spent time figuring out whether there was change within the set or not, or spent the time for actually changing the set every time. Worst case scenario if we check for a change is; yes there was a change + change the session(maximum time spent). Best case scenario is; there was not a change so don't do anything(half-time spent).
if we set to constant `true`, All scenarios take same amount which is half-time spent to change the set.

Moreover looks like changing the set is O(1) AFAICT but  checking for a change would be O(n). So not sure why this was even an option.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Let's add a TODO to set a proper weight later.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: It was returning the value in CurrentSession which wasn't actually the current session in the case of Serai (as documented). It was the latest decided session. The current session was the one pallet-session had.

... it was pretty poorly named.

Reading the signals pallet, it appears it's not fundamentally broken. It's just mis-labeling sets, as it'll label the latest Serai set as having the current session.

It'd be best to restore this function and revert the changes to the signals pallet.

substrate/runtime/src/lib.rs
```diff
@@ -540,9 +541,13 @@ sp_api::impl_runtime_apis! {
 
   impl sp_authority_discovery::AuthorityDiscoveryApi<Block> for Runtime {
     fn authorities() -> Vec<AuthorityDiscoveryId> {
-      ValidatorSets::serai_validators()
+      Babe::authorities()
         .into_iter()
-        .map(|(id, _)| AuthorityDiscoveryId::from(id))
+        .map(|(id, _)| {
+          // TODO: any better way to do this?
```
kayabaNerve: id.into_inner().into()

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -64,18 +65,11 @@ pub mod pallet {
 
   /// The current session for a network.
   ///
```
kayabaNerve: This not being `//` may add a new line to the doc comments which shouldn't be there. I'm unsure if it'll trim...

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -110,13 +104,8 @@ pub mod pallet {
...
     #[inline]
     fn in_active_serai_set(account: Public) -> bool {
-      // TODO: This is bounded O(n). Can we get O(1) via a storage lookup, like we do with InSet?
-      for (validator, _) in Self::serai_validators() {
-        if validator == account {
-          return true;
-        }
-      }
-      false
+      // TODO: TODO: is_member is internally O(n). Update Babe to use an O(1) storage lookup?
```
kayabaNerve: Double TODO :P

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -84,15 +83,15 @@ pub mod pallet {
     _,
     Identity,
     NetworkId,
-    BoundedVec<Public, ConstU32<{ MAX_KEY_SHARES_PER_SET }>>,
+    BoundedVec<(Public, u64), ConstU32<{ MAX_KEY_SHARES_PER_SET }>>,
```
kayabaNerve, line 95: CI failure is due to serai-client expecting this to be u16 while this is a u64.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Except we never read this info from InSet AFAICT, solely from Participants. That should remove the reason for InSet being modified.

PR #447, opened 2023-11-21
implements the db macro for processor/src/multisigs/scanner.rs

econsta:


kayabaNerve:
Sorry this slipped through the cracks. Deprecated by #613, which further extends and extensively uses `create_db`. It's truly great to have and very appreciated :)

PR #448, opened 2023-11-21
Convert tributary to use create_db macro

davidjohnbell:
### Description

This is a continuation of pr #408. `TributaryDb` was split into several DBs, some helper methods to convert to/from the expected type were removed as they're covered by the macro. 

### Considerations
- At the moment we do not have a naming and plurality convention for databases (e.g, EventDb vs Events). There are issues concerning matching existing structs and potential conflicts with other modules. I wonder if we could use the first namespace in the token to create a module and then drop the suffix. Usage or implementation would be `impl substrate_db::Events { ... }`? I am missing a convention in this implementation opting for what felt right and avoiding conflicts.


kayabaNerve:
Looks good! Thanks :)

PR #455, opened 2023-11-26
Use parity-db in current Dockerfiles

kayabaNerve:
The motivation for redb was to remove the multiple rocksdb compile times from CI.


kayabaNerve:
This has massively faster CI times, yet I believe some services create TXNs while a TXN already exists, causing deadlocks with the redb backend :/

subxt also consistently fails for 4 specific tests with a websocket connectivity error? I have no idea why that's happening (a debug assertion, some really specific performance commentary...).

This PR *likely* should be broken into a few smaller PRs that let is triage individual changes.

kayabaNerve:
Rebased on top of #462.

kayabaNerve:
While I want to see if this PR's CI runs work (except coordinator and full stack), coordinator will always fundamentally fail. The tributary manages its own TXNs so we will always have nested TXNs.

sled? It hasn't had a release in a while, so we'd have to go to marble or it as a git dependency.

We could use the MemDB? AgateDB, by tikv? ParityDB, which is a patricia merkle trie yet won't incur C compile times?

kayabaNerve:
The prior full stack tests failed with the following.

```
2023-11-30T06:49:27.5628124Z [2023-11-30T06:41:01Z TRACE serai_coordinator] processor message effected transaction 95eb7054a83861413a3c66105e3847ba8c7613888998178449ffff33e4016eb7 Transaction::DkgCommitments { attempt: 0, signer: "e2f2ae0a6abc4e71a884a961c500515f58e30b6aa582dd8db6a65945e08d2d76", .. }
2023-11-30T06:49:27.5629363Z [2023-11-30T06:41:01Z TRACE serai_coordinator] getting next nonce for Tributary TX in response to processor message
2023-11-30T06:49:27.5630370Z [2023-11-30T06:41:01Z DEBUG serai_coordinator] publishing transaction e37b1ae87eb737b75af907d603a5d60c59298cbae7f5c8f036bcca1c1d190782
2023-11-30T06:49:27.5630879Z thread 'tokio-runtime-worker' panicked at coordinator/src/main.rs:167:17:
2023-11-30T06:49:27.5631858Z created an invalid transaction: InvalidSignature
```

I have absolutely no clue how this happened. The DB being corrupt should not have this effect. My only idea is DB Transaction::get being broken led to a genesis hash mismatch? That's the only piece of contextual data in the signing process.

kayabaNerve:
This saves 20-25 minutes in the CI.

PR #456, opened 2023-11-26
Resolve #360

### LINKED ISSUE #360: ProcessorMessages have extraneous/suboptimized fields
As the processor and coordinator are e2e tested, certain fields have gone unused/been found suboptimal and worthy of revisiting. While we need to revisit all of them, I'd immediately like to flag `key`.

`key` isn't needed for batches since only a single key will be used for batches at any given time (per network).

It is needed for sign, if we sign TXs from multiple multisigs at once (such as while transitioning). Instead of key, we should use the much more minimal `Session`. That'd allow reverting https://github.com/serai-dex/serai/commit/f249e20028e8dca2218ea240cc20ed40791014a5 and reduce disk space usage of messages a good percent.

kayabaNerve:
Uses session instead of key and performs other misc optimizations.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7866,7 +7866,6 @@ dependencies = [
  "serai-in-instructions-primitives",
  "serai-primitives",
  "serai-validator-sets-primitives",
- "zeroize",
 ]
 
 [[package]]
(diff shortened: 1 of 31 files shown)


kayabaNerve:
Resolves #360.

kayabaNerve:
Should save ~20% off message-queue storage costs, reduces code complexity a bit in exchange for the processor tracking sessions, and removes ~150 lines in total.

PR #457, opened 2023-11-26
Replace rustls with boring-ssl

kayabaNerve:
This removes all re-attempts present in monero-serai's RPC and is an attempt to narrow down the sporadic failures.

Inspired by https://github.com/hyperium/hyper/issues/3427

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -172,6 +172,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "antidote"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "34fde25430d87a9388dadbe6e34d7f72a462c8b43ac8d309b42b0a8505d7e2a5"
+
+[[package]]
 name = "anyhow"
 version = "1.0.75"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


kayabaNerve:
Without re-attempts, the CI consistently failed due to the amount of requests we made during a CI run. Recently, a run failed even with re-attempts, making the issue incredibly rare yet still present and reachable.

I'm just curious what this CI reports.

kayabaNerve:
```
2023-11-27T05:36:57.9226749Z In CI, so clearing cache to prevent hitting the storage limits.
2023-11-27T05:36:57.9227681Z thread 'tests::mint_and_burn::mint_and_burn_test' panicked at tests/full-stack/src/tests/mint_and_burn.rs:147:78:
2023-11-27T05:36:57.9228672Z called `Result::unwrap()` on an `Err` value: ConnectionError("Hyper(hyper::Error(IncompleteMessage))")
```

This still got IncompleteMessage, albeit, potentially with a very very reduced rate? I think if the edits I applied to coins/monero were applied to develop, while remaining on rustls, all of these CI runs would've failed.

Since there's no definity here, and this adds multiple depends, I'm fine closing this.

Remaining candidates are our async driving code, hyper itself, and epee (the server).

kayabaNerve:
I reopened this when I attempted moving to hyper 1.0, as possible thanks to an open PR to hyper-boring.

hyper 1.0 + boring-ssl still has the incomplete message issue, so this PR should be dropped.

Its work on moving to 1.0 should be yoinked to a new PR once we can move over *with rustls*.

Issue #402, opened 2023-10-17
Reduction of allocated stake requirements via light clients


kayabaNerve:
With the context of the problem statement in https://github.com/serai-dex/serai/issues/400, a question is raised:

> Can we ensure the SRI in a pool cannot be drained, removing the need to allocate stake to secure it?

Most other networks solve this by writing the value of the SRI-equivalent off to 0 upon multisig compromise, not allowing it to be used for profit statements. Serai is distinct for being robust against multisig compromise and would need a distinct mechanism.

Light clients are the immediately accessible idea. If we are able to independently validate the `InInstruction`s received from external networks, we're able to prevent the multisig from printing fake sriXYZ and in the process, draining the XYZ:SRI liquidity pool.

We aren't however, as stolen XYZ may be looped back into the multisig to give the appearance of the multisig having legitimately received fresh XYZ in order to execute swaps with. We'd need to prove, with each new XYZ received, no prior misbehavior has occurred.

Naively, this would require moving all scheduling logic currently in the processor on-chain and on-chain executing it with full views of the external chain state also included on-chain.

Practically, we actually could discuss compiling the scheduling logic from the processor into a no-std variant *which we then execute with [RISC Zero](https://www.risczero.com/)*. Even if the Serai chain only has the block hash from the external network, we'd be able to prove all TXs are as expected for the cost of a STARK. We would not have to hand-write any ZK circuits ourselves either.

There is still an exploit window under this proposal. A malicious multisig could send itself legitimate external XYZ, dumping them into the XYZ:SRI pool. After draining the SRI side that way, they wouldn't provide backing to arbitragers swaps in the other direction. They'd simply walk away with the XYZ given that any `Burn`s from SRI aren't expected to appear on-chain for several blocks.

In order to prevent this, under this proposal *and in general*, the stake allocation requirement cannot use the spot price of XYZ:SRI, yet instead must use some TWAP to ensure any attempt to so dump XYZ hits the capacity limit and errors. That way, while XYZ can still be sold to maximize profit, the XYZ taken *and the SRI profited* would be less than the stake allocated by the malicious actors.

This tail commentary is already noted in #394, leaving this issue with the idea of on-chain STARK proofs of proper behavior to prevent the ability to drain the SRI side of the pools when turning malicious. This would halve stake requirements.

[...]

Issue #394, opened 2023-10-12
Deallocation scheduling for validators


kayabaNerve:
For a small network secure by just 4 validators, the following risk exists:

1) Validators flood the pool with XYZ which doesn't actually exist, making inf XYZ valued at 1 SRI.
2) Validators deallocate due to the now reduced stake requirement.
3) Validators walk away with both the XYZ and the SRI.

Serai's continued security of having enough stake in the form of SRI w.r.t. lost XYZ is premised on the handover protocol. If the prior validators are so malicious, regardless of if its detected, the new validators will either refuse the handover or will accept the handover and maintain the SRI needed.

Under the above flow however, the new validators accept the handover *with an artificially low SRI requirement*, breaking security.

The naive solution is to look at the highest valuation for the token ever created, and always use that in stake requirement valuations. That way draining the pools doesn't decrease the SRI requirement.

The next iteration is to use the highest valuation over a single session, which does offer some reflexivity. For a malicious validator set at session `n`, and a cooperating malicious validator set at session `n+1`, this would prevent `n+1` from removing their SRI until `n+2`. Please note this scheme does expand from a single session to an `n` session write-up.

This would require either:

1) Other nodes to step in (such as the Serai validators censoring their TXs, preventing the handover, until a network upgrade)
2) Other actors with economic stake to sign up to be in `n+2`, preventing the handover, trapping the SRI stake of `n+1`

What's likely best is for a supermajority of nodes to be able to pause handovers for any external network. This gives time to perform a network upgrade. The only question is how much time would be needed for a supermajority of nodes to decide to pause handovers?

A single session, or one week, should be fine. 

Then there's the commentary a validator can decide to remove themselves from the validator set a block before the next validator set is decided, and in doing so, make a large amount of SRI suddenly available. This is compounded by the edge case for the Serai validator set itself where `validator-sets` doesn't track the current set in `InSet` yet *the upcoming set*, due to when `pallet-session` wants validators decided. This leads to a currently present issue within `validator-sets` where the current Serai validators are not considered active.

If we only let the prior validator set (`n-1`) actually reclaim their SRI once validator set `n+1` has been handed over to, this would extend the time period for the above discussion to two sessions (two weeks) *and* partially correct the current issue within the `validator-sets` pallet. It would also ensure notice before large portions of SRI become unlocked.

[...]

PR #459, opened 2023-11-27
Ensure economic security on validator sets

akildemir:


substrate/coins/pallet/src/lib.rs
```diff
@@ -43,6 +57,7 @@ pub mod pallet {
     AmountOverflowed,
     NotEnoughCoins,
     BurnWithInstructionNotAllowed,
+    NotEnoughStake,
```
kayabaNerve: MintNotAllowed?

substrate/coins/pallet/src/lib.rs
```diff
@@ -142,15 +157,21 @@ pub mod pallet {
...
-      // update the balance
-      Self::increase_balance_internal(to, balance)?;
-
-      // update the supply
       let new_supply = Self::supply(balance.coin)
         .checked_add(balance.amount.0)
         .ok_or(Error::<T, I>::AmountOverflowed)?;
+
+      // skip the economics check for lp tokens.
+      if TypeId::of::<I>() != TypeId::of::<LiquidityTokensInstance>() &&
```
kayabaNerve: Shouldn't we remove this entirely and just have the LiqTokensInstance use () for AllowMint?

substrate/dex/pallet/src/lib.rs
```diff
@@ -154,11 +154,34 @@ pub mod pallet {
...
 
+  #[pallet::storage]
+  #[pallet::getter(fn last_quote_for_block)]
+  pub type LastQuoteForBlock<T: Config> =
+    StorageDoubleMap<_, Identity, BlockNumberFor<T>, Identity, Coin, [u8; 8], ValueQuery>;
+
+  /// Moving window of oracle prices.
+  #[pallet::storage]
+  #[pallet::getter(fn oracle_prices)]
+  pub type OraclePrices<T: Config> = StorageMap<_, Identity, [u8; 8], u16, OptionQuery>;
```
kayabaNerve: I'd document the [u8; 8] is the amount's big endian bytes, and u16 is the amount of inclusions in this multi-set.

We also should have an on-genesis assert the window size wasn't configured to be >= 2**16 (which would risk an overflow here).

substrate/dex/pallet/src/lib.rs
```diff
@@ -154,11 +154,34 @@ pub mod pallet {
...
+  impl<T: Config> Pallet<T> {
+    /// Get the highest sustained(lowest) value of the window
```
kayabaNerve: """
Get the highest sustained value for this window.

This is actually the lowest price observed during the windows, as it's the price all prices are greater than or equal to.
"""

substrate/dex/pallet/src/lib.rs
```diff
@@ -154,11 +154,34 @@ pub mod pallet {
...
+    pub fn highest_sustained_price() -> Amount {
+      let mut iter = OraclePrices::<T>::iter_keys();
+      // we can unwrap because we will always have at least 1 key in this map
+      // before this function is called.
```
kayabaNerve: I really don't like this reasoning. It sets bounds on callers which aren't documented. Mind having this return an Option and having the caller handle it

OR

saying it's fine because the genesis enters a price of 0 for block 0?

Either make the assumptions endogenous, not exogenous.

substrate/dex/pallet/src/lib.rs
```diff
@@ -154,11 +154,34 @@ pub mod pallet {
...
+      Amount(u64::from_be_bytes(iter.next().unwrap()))
```
kayabaNerve: I'd add a note this was the lowest price due to the keys being lexicographically ordered.

substrate/dex/pallet/src/lib.rs
kayabaNerve: You're also missing the Coin. You need a DoubleMap with Coin, [u8; 8] or a Map with (Coin, [u8;8]).

substrate/dex/pallet/src/lib.rs
```diff
@@ -311,6 +334,38 @@ pub mod pallet {
 
   #[pallet::hooks]
   impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    fn on_finalize(n: BlockNumberFor<T>) {
+      for coin in Pools::<T>::iter_keys() {
+        // insert the new price to our oracle window
+        let last = Self::quote_price_exact_tokens_for_tokens(coin, Coin::native(), 1, false)
```
kayabaNerve: 1 isn't a valid quantity here. It'll lose all precision. Any coin worth less than SRI will always return 0.

I'd recommend using as many decimals as the external coin has (we used to have this metadata back with pallet-assets, but I believe it was removed/moved to some primitives?).

kayabaNerve: I'd also document the security considerations behind using the end of block price re: preventing mid-block spikes a malicious block proposer can exploit without risk of being arbitraged against.

substrate/dex/pallet/src/lib.rs
kayabaNerve: This just has greater clarity on highest sustained == lowest.

substrate/dex/pallet/src/lib.rs
```diff
@@ -154,11 +154,34 @@ pub mod pallet {
...
+  pub type LastQuoteForBlock<T: Config> =
+    StorageDoubleMap<_, Identity, BlockNumberFor<T>, Identity, Coin, [u8; 8], ValueQuery>;
+
+  /// Moving window of oracle prices.
+  #[pallet::storage]
+  #[pallet::getter(fn oracle_prices)]
+  pub type OraclePrices<T: Config> = StorageMap<_, Identity, [u8; 8], u16, OptionQuery>;
+  impl<T: Config> Pallet<T> {
+    /// Get the highest sustained(lowest) value of the window
+    pub fn highest_sustained_price() -> Amount {
```
kayabaNerve: Should we have a TODO here to consider an algorithm which removes outliers? This algorithm might work a good bit better if we remove the bottom `n` values (so some value sustained over 90% of blocks instead of all blocks in the window), 

I don't want to force that economic discussion into this PR though.

substrate/dex/pallet/src/types.rs
```diff
@@ -23,6 +23,9 @@ use super::*;
 use codec::{Decode, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
 
+/// 1hr worth of blocks.
+pub const ORACLE_WINDOW_SIZE: u32 = 600;
```
kayabaNerve: We should document how this needs to be long enough for arbitrage to occur and make holding any fake price up sufficiently unprofitable.

Also, this isn't long enough for that AFAICT. It'd take longer than this just to add BTC to Serai to swap it.

substrate/runtime/src/lib.rs
```diff
@@ -224,10 +224,12 @@ impl transaction_payment::Config for Runtime {
 
 impl coins::Config for Runtime {
   type RuntimeEvent = RuntimeEvent;
+  type AllowMint = ValidatorSets;
 }
 
 impl coins::Config<coins::Instance1> for Runtime {
   type RuntimeEvent = RuntimeEvent;
+  type AllowMint = ValidatorSets;
```
kayabaNerve: Shouldn't this be ()?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -377,6 +381,8 @@ pub mod pallet {
     BadSignature,
     /// Validator wasn't registered or active.
     NonExistentValidator,
+    /// Deallocation would take the stake below what is required.
+    NotEnoughStake,
```
kayabaNerve: "DeallocationWouldRemoveEconomicSecurity"? Keep it line with the similar ones above?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -703,6 +720,32 @@ pub mod pallet {
         validators.into_iter().map(|(id, w)| (GrandpaAuthorityId::from(id), w)).collect(),
       );
     }
+
+    /// Returns the required stake in terms SRI for a given `Balance`.
+    pub fn required_stake(balance: &Balance) -> SubstrateAmount {
+      // TODO: unwrap_or? We will always have an oracle value except the first block
+      // of each session since we have to wait for that block to end to update the window and
+      // the oracle value accordingly.
```
kayabaNerve: unwrap_or seems fine here.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -703,6 +720,32 @@ pub mod pallet {
...
+      let price = Dex::<T>::oracle_value(balance.coin).unwrap_or(Amount(0));
+      let mut total_coin_value = balance.amount.0.saturating_mul(price.0);
```
kayabaNerve: If we update the quote to have higher precision, here needs to be updated.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -703,6 +720,32 @@ pub mod pallet {
...
+
+      // required stake formula (COIN_VALUE * 1.5) + margin(20%)
+      total_coin_value = total_coin_value.saturating_mul(3).saturating_div(2);
+      total_coin_value.saturating_add(total_coin_value.saturating_div(5))
+    }
+
+    /// Returns the current total required stake for a given `network`.
+    pub fn required_stake_for_network(network: NetworkId) -> SubstrateAmount {
+      let mut total_required: SubstrateAmount = 0;
+      // TODO: have network.coins() instead of this?
```
kayabaNerve: `fn coins(&self) -> &'static [Coin]` sounds good to me :)

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -835,6 +879,18 @@ pub mod pallet {
     }
   }
 
+  impl<T: Config> AllowMint for Pallet<T> {
+    fn allow(balance: &Balance) -> bool {
```
kayabaNerve: is_allowed instead of allow? I'd expect allow to take in a struct's &mut self and whitelist something, not check if it's allowed.

kayabaNerve:
Resolves #394.

Closes #402.

PR #462, opened 2023-11-28
Use debug builds in our Dockerfiles to reduce CI times

kayabaNerve:
Also enables only spawning the mdns service when debug in the coordinator.

Broken out from https://github.com/serai-dex/serai/pull/455.


kayabaNerve:
The processor test is a legitimate failure.

```
2023-11-28T06:14:14.2781401Z thread 'main' panicked at coins/monero/src/wallet/send/mod.rs:533:5:
2023-11-28T06:14:14.2782187Z assertion `left == right` failed: Outgoing amount + fee != incoming amount
2023-11-28T06:14:14.2782654Z   left: 2260500000
2023-11-28T06:14:14.2782892Z  right: 1000000000000
```

PR #465, opened 2023-12-01
make Monero HTTP RPC timeout configurable.

Boog900:


kayabaNerve:
Thanks :) Will merge once CI passes. Will also try to set a saner default later.

PR #468, opened 2023-12-04
Slash bad validators

akildemir:


substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1037,6 +1246,16 @@ pub mod pallet {
     }
   }
 
+  impl<T: Config> FindAuthor<Public> for Pallet<T> {
```
kayabaNerve, line 1125: FindAuthor is implemented directly for pallet_babe. Is there a reason to re-impl?

substrate/runtime/src/lib.rs
```diff
@@ -134,8 +141,7 @@ parameter_types! {
       Weight::from_parts(2u64 * WEIGHT_REF_TIME_PER_SECOND, u64::MAX),
       NORMAL_DISPATCH_RATIO,
     );
-
-  pub const MaxAuthorities: u32 = validator_sets::primitives::MAX_KEY_SHARES_PER_SET;
+  pub ReportLongevity: u64 = 24 * 600; // TODO: get epoch duration from babe.
```
kayabaNerve: Should we resolve this TODO here? It's babe::Config::EpochDuration.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -632,7 +696,7 @@ pub mod pallet {
       } else {
         to_unlock_on.0 += 1;
       }
-      // Increase the session by one, creating a cooldown period
+      // Increase the session by one, creating a cool down period
```
kayabaNerve: cooldown is a single word or a hyphenated term. It is not valid with a space.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1025,6 +1145,95 @@ pub mod pallet {
...
+      GrandpaOffences::<T>::set(key, Some(()));
+      Ok(())
+    }
+
+    fn is_known_offence(
+      offenders: &[Public],
+      slot: &<GrandpaEquivocationOffence<Public> as Offence<Public>>::TimeSlot,
+    ) -> bool {
+      let key = (offenders[0], slot.set_id, slot.round);
+      Self::grandpa_offences(key).is_some()
```
kayabaNerve: If we slash validators to 0, can't we simply say it's a known offence if they have yet to be slashed and still have stake in set?

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -715,40 +779,96 @@ pub mod pallet {
...
+      // TODO: we don't know in which block the it might be deallocating, so we go through all
+      // but this is ugly.
+      for i in 0 .. 4 {
+        to_unlock_on.0 += i;
+        let pending = Self::take_deallocatable_amount(network, to_unlock_on, account);
+        if pending.is_some() {
+          allocation = allocation.add(pending.unwrap());
+          PendingDeallocations::<T>::set((network, to_unlock_on, account), None);
+        }
+      }
```
kayabaNerve: We can rewrite PendingDeallocations to be (network, account), to_unlock_on as a DoubleMap and use clear_prefix.

Also, take already removes it, so you don't need to set None.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -715,40 +779,96 @@ pub mod pallet {
...
+      // remove from the current & future sets.
+      // TODO: how to remove from the current?
+      let mut participants =
+        Participants::<T>::get(network).expect("no Serai participants on slash");
+      let index = participants.iter().position(|(id, _)| *id == account);
+      if let Some(i) = index {
+        participants.remove(i);
+        Participants::<T>::set(network, Some(participants));
+        InSet::<T>::remove(Self::in_set_key(network, account));
+      }
```
kayabaNerve, line 812: Removal from Participants + InSet, as done here, then the goal is disabling. Disabling is a mess. A recent comment on the polkadot-sdk repo sums it up best.

https://github.com/paritytech/polkadot-sdk/issues/784#issuecomment-1830345189

Basically, for us, the important part is:

> Disabling affects Block Authoring. (Both ways: block authoring equivocation disables and disabling stops block authoring)

I'm unsure we currently can use pallet-babe to error on blocks created by a slashed author. I'm also unsure we want to write our own shim for that.

kayabaNerve: Also, there's probably some unhandled side effects for the signals pallet here.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: Because you want to track the removed values, I believe there's a PrefixIterator which takes as it iterates. I'd guess it'd be available via `drain_prefix`, yet I haven't checked.

substrate/runtime/src/lib.rs
```diff
@@ -268,20 +274,34 @@ impl in_instructions::Config for Runtime {
...
+impl<C> frame_system::offchain::SendTransactionTypes<C> for Runtime
+where
+  RuntimeCall: From<C>,
+{
+  type Extrinsic = UncheckedExtrinsic;
+  type OverarchingCall = RuntimeCall;
+}
+
+// for validating equivocation evidences.
+impl pallet_authorship::Config for Runtime {
```
kayabaNerve, line 293: It's ridiculous to have pallet_authorship so Babe can call it for FindAuthor which just calls it for the author.

I'm not saying this shouldn't be here, and I get pallet_authorship is tracking historical values. I'm just noting this should be better in theory. It's fine for this PR.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1,5 +1,33 @@
...
+impl GetSessionNumber for MembershipProof {
+  fn session(&self) -> u32 {
+    self.session
+  }
+}
+impl GetValidatorCount for MembershipProof {
+  fn validator_count(&self) -> u32 {
+    self.validator_count
+  }
+}
```
kayabaNerve, line 58: Can we define:

1) Define MembershipProof  as the key alone
2) Have `GetSessionNumber` just read from the current state, returning the current session if in set, else the prior set
3) I don't have an immediate comment on GetValidatorCount as I can't claim I know why that exists tbh

Then prove would just return the key, and verify would check if they're in-set (if so true), and if not, if they have a pending deallocation (if so, true). Otherwise, false as it's out of date.

The session number will be valid so long as the proof is valid.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1025,6 +1145,95 @@ pub mod pallet {
...
+    type IdentificationTuple = Public;
+
+    fn prove(key: (KeyTypeId, D)) -> Option<Self::Proof> {
+      let (_, data) = key;
+      let validator = Public::try_from(data.as_ref()).ok()?;
+
+      // TODO: do we want this check here? this is node that wants to send the proof anyway.
+      if !Self::can_slash_serai_validator(&validator) {
+        return None;
+      }
```
kayabaNerve: I don't believe this call is needed here.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1025,6 +1145,95 @@ pub mod pallet {
...
+
+      Some(MembershipProof {
+        session: Self::session(NetworkId::Serai).unwrap().0,
+        // TODO: since this isn't even used, should we just 0 or 1ed out?
+        validator_count: Participants::<T>::get(NetworkId::Serai)
```
kayabaNerve: This is used by GetValidatorCount. If GetValidatorCount isn't used, that's a distinct note.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -1025,6 +1145,95 @@ pub mod pallet {
...
+        key: validator,
+      })
+    }
+
+    fn check_proof(key: (KeyTypeId, D), proof: Self::Proof) -> Option<Self::IdentificationTuple> {
+      let (_, data) = key;
+      let validator = Public::try_from(data.as_ref()).ok()?;
+
+      // check the offender and the proof offender are the same.
+      if validator != proof.key {
```
kayabaNerve: If the validator is passed here, why have it in the proof at all? I don't see a trait which returns it we must satisfy.

kayabaNerve (changes requested):
Definitely a series of complex changes. I do get some of these will be inevitable.

substrate/validator-sets/pallet/src/lib.rs
akildemir: The one on the babe takes the index as the type parameter but `pallet_authorship` requires implementation to take `AccountId`, hence we implement for the vs-set  but alternatively we can modify the one on the babe but not sure if we want to do that.

substrate/validator-sets/pallet/src/lib.rs
akildemir: well my dictionary was screaming that "cooldown" is not a word. Seems it wasn't correct.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: That's a very weird indirection. Thanks for clarifying.

substrate/validator-sets/pallet/src/lib.rs
akildemir: `pallet_babe` exposes `DisabledValidators` type that we implement [here](https://github.com/serai-dex/serai/blob/c511a54d18453814719155b30777c99bcf355f7f/substrate/validator-sets/pallet/src/lib.rs#L1004). The thing is yes, all it does is to "panic" if one of those disabled ones produces a block. So surely somewhere they prevent picking these validators from producing a block? Otherwise whole network would halt when one of them does. Because I don't see that code in the runtime-babe and so I didn't use that feature atm and put a TODO.

substrate/validator-sets/pallet/src/lib.rs
akildemir: This is related to https://github.com/serai-dex/serai/pull/468#discussion_r1418974452. If the offender still in the set and can make blocks or decide on the blocks(in grandpa) then we can have the same guy be the offender twice although since we slash everything the first time, we can't do anything(so why care?) the second time.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: No, panicking in the runtime won't cause the node to panic. It'll only cause the block to panic and be rejected.

You shouldn't panic in the runtime. Panicking in the runtime if a disabled validator is a block author is valid.

substrate/validator-sets/pallet/src/lib.rs
akildemir: since our proof is the validator id itlself, yes we don't need it normally. But if we want to check whether it is in the current session or no within `GetSessionNumber` as stated here https://github.com/serai-dex/serai/pull/468#discussion_r1417123763 we then need it.

kayabaNerve:
Rebased onto develop. Working on reviewing.

PR #472, opened 2023-12-07
monero-serai: make it clear that not providing a change address is fingerprintable

j-berman:
When a signable transaction is constructed without a change address specified, all change is added to the tx fee (this helps Serai avoid accumulating dust outputs that spam the chain). This PR makes it clearer to the caller that it is fingerprintable when the caller does not provide a change address, since the caller must explicitly initialize the change param with `Change::fingerprintable(None)` (as is done in the added test). A tx created this way is fingerprintable because the fee is distinguishable from a fee created with wallet2.

### DIFF
--- a/coins/monero/src/wallet/send/mod.rs
+++ b/coins/monero/src/wallet/send/mod.rs
@@ -324,7 +324,7 @@ pub struct SignableTransaction {
 /// Specification for a change output.
 #[derive(Clone, PartialEq, Eq, Zeroize)]
 pub struct Change {
-  address: MoneroAddress,
+  address: Option<MoneroAddress>,
   view: Option<Zeroizing<Scalar>>,
 }
 
@@ -338,21 +338,21 @@ impl Change {
   /// Create a change output specification from a ViewPair, as needed to maintain privacy.
   pub fn new(view: &ViewPair, guaranteed: bool) -> Change {
     Change {
-      address: view.address(
+      address: Some(view.address(
         Network::Mainnet,
         if !guaranteed {
           AddressSpec::Standard
         } else {
           AddressSpec::Featured { subaddress: None, payment_id: None, guaranteed: true }
         },
-      ),
+      )),
       view: Some(view.view.clone()),
     }
   }
 
   /// Create a fingerprintable change output specification which will harm privacy. Only use this
   /// if you know what you're doing.
-  pub fn fingerprintable(address: MoneroAddress) -> Change {
+  pub fn fingerprintable(address: Option<MoneroAddress>) -> Change {
     Change { address, view: None }
   }
 }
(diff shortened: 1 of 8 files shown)


coins/monero/src/wallet/send/mod.rs
```diff
@@ -364,13 +364,17 @@ fn need_additional(payments: &[InternalPayment]) -> (bool, bool) {
...
+            has_change_view = true;
+            // It should not be possible to construct a change specification to a subaddress with a
+            // view key
+            debug_assert!(!change_address.is_subaddress());
+          }
+          change_address.is_subaddress()
+        } else {
+          false
         }
-        change.address.is_subaddress()
```
kayabaNerve, line 373: Will an InternalPayment::Change exist if change.address is none?

coins/monero/src/wallet/send/mod.rs
```diff
@@ -624,7 +627,7 @@ impl SignableTransaction {
       // regarding it's view key
       payment = if !modified_change_ecdh {
         if let InternalPayment::Change(change, amount) = &payment {
-          InternalPayment::Payment((change.address, *amount))
+          InternalPayment::Payment((change.address.unwrap(), *amount))
```
kayabaNerve: Prior comment asked about this, here unwraps.

We could also make InternalPayment::Change not internally use the Change struct, avoiding the Option entirely and offering compile-time safety.

coins/monero/tests/runner.rs
```diff
@@ -298,7 +298,10 @@ macro_rules! test {
             rpc.publish_transaction(&signed).await.unwrap();
             mine_until_unlocked(&rpc, &random_address().2.to_string(), signed.hash()).await;
             let tx = rpc.get_transaction(signed.hash()).await.unwrap();
-            check_weight_and_fee(&tx, fee_rate);
+            if stringify!($name) != "spend_one_input_to_two_outputs_no_change" {
+              // Skip weight and fee check for the above test
```
kayabaNerve: Mind having the comment explain why?

" as if there's no change, any would-be change will be shunted to the fee" or so?

processor/src/networks/monero.rs
```diff
@@ -376,6 +376,12 @@ impl Monero {
       })
       .collect::<Vec<_>>();
 
+    let change = Change::fingerprintable(if change.is_some() {
+      Some(change.clone().unwrap().into())
```
kayabaNerve: `if let Some(change) = change {}`, to avoid the unwrap

kayabaNerve (changes requested):
LGTM, thanks for resolving this fingerprint :)

coins/monero/src/wallet/send/mod.rs
j-berman: It shouldn't, I went with your suggestion in the other comment to not use the Option for `InternalPayment::Change`

coins/monero/src/wallet/send/mod.rs
j-berman: > We could also make InternalPayment::Change not internally use the Change struct, avoiding the Option entirely and offering compile-time safety.

I thought of doing this when working on it too and decided not to. I like it though. Done in the latest.

coins/monero/tests/runner.rs
j-berman: Done

processor/src/networks/monero.rs
j-berman: Did that suggestion, then the compiler warned it's equivalent to `Option::map` and  suggested `change.as_ref().map(|change| change.clone().into())`. See the latest. I was wondering what the cleanest way to write this was, cool.

kayabaNerve:
Thanks :)

Issue #258, opened 2023-03-15
Support input-encoded instructions


kayabaNerve:
Ordinals encode data on Bitcoin by creating a P2TR output, then in the revealed scripts, including `PUSH FALSE, IF, PUSH *data*, ELSE, CHECK SIG` (roughly). This has no limit (OP_RETURN has an 80-byte limit) and takes advantage of the discount applied to witness data. Serai can support a similar method for its instructions.

I will note ordinals are contentious. SegWit data is prunable, so it's undeniably healthier than OP_RETURN (in my opinion). The main concern is BTC advocates may ban ordinals. In order to not get banned, which would likely be done by banning scripts with `IF FALSE`s, we'd likely need to have a script where the data segment is a potential spend path (if a hash preimage is provided). Serai clients would then make up a hash with no known preimage (and till requiring a `CHECK SIG`).

While this does require creating a leading transaction, to then spend with the instruction encoded in the input, Bitcoin supports TX chaining so there shouldn't be a time delay on this. As long as the second TX is the one sending to Serai, there's also no risk of failing the data availability problem.

kayabaNerve:
This will be necessary in order to perform BTC -> XMR swaps post-Seraphis, as JAMTIS addresses exceed 80 bytes.

PR #486, opened 2023-12-12
support input encoded data for bitcoin network

akildemir:
addresses  https://github.com/serai-dex/serai/issues/258

### DIFF
--- a/processor/src/networks/bitcoin.rs
+++ b/processor/src/networks/bitcoin.rs
@@ -503,9 +503,9 @@ impl Bitcoin {
       for input in &tx.input {
         if self.spent_output_of(input).await.script_pubkey.is_p2wsh() {
           let witness = input.witness.to_vec();
-          let redeem_script = ScriptBuf::from_bytes(witness[0].clone());
+          let redeem_script = ScriptBuf::from_bytes(witness[1].clone());
           if let Some(true) = self.expected_script_pattern(&redeem_script) {
-            data = witness[1].clone();
+            data = witness[0].clone();
             break;
           }
         }
--- a/processor/src/networks/bitcoin.rs
+++ b/processor/src/networks/bitcoin.rs
@@ -21,7 +21,9 @@ use bitcoin_serai::{
     consensus::{Encodable, Decodable},
     script::Instruction,
     address::{NetworkChecked, Address as BAddress},
-    Transaction, Block, Network as BNetwork,
+    Transaction, Block, Network as BNetwork, ScriptBuf,
+    opcodes::all::{OP_SHA256, OP_EQUAL},
+    blockdata::transaction::{TxIn, TxOut},
   },
   wallet::{
     tweak_keys, address_payload, ReceivedOutput, Scanner, TransactionError,
@@ -37,7 +39,7 @@ use bitcoin_serai::bitcoin::{
   sighash::{EcdsaSighashType, SighashCache},
   script::{PushBytesBuf, Builder},
   absolute::LockTime,
-  Amount as BAmount, Sequence, Script, Witness, OutPoint, TxOut, TxIn,
+  Amount as BAmount, Sequence, Script, Witness, OutPoint,
   transaction::Version,
 };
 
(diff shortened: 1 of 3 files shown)


processor/src/tests/literal/mod.rs
```diff
@@ -19,6 +51,141 @@ mod bitcoin {
...
+      let message = Shorthand::transfer(None, serai_address.into()).encode();
+      let mut data = Sha256::engine();
+      data.input(&message);
+
+      // make the output script => `OP_SHA256 PUSH MSG_HASH OP_EQUAL`
+      let script = ScriptBuf::builder()
+        .push_opcode(OP_SHA256)
+        .push_slice(Sha256::from_engine(data).as_byte_array())
+        .push_opcode(OP_EQUAL)
+        .into_script();
```
kayabaNerve, line 98: The output script also needs to be *concatenated* with a traditional spend script. Ideally, the `scriptPubKey` is written `msg_script || spend_script` and the sig `msg || script_sig` (where, of course, `script_sig` is a `vec` of witness items, so `msg` would simply be the first).

Else there's a DoS.

processor/src/tests/literal/mod.rs
```diff
@@ -19,6 +51,141 @@ mod bitcoin {
...
+      let Ok(shorthand) = Shorthand::decode(&mut data) else {
+        panic!("can't decode msg data into Shorthand")
+      };
+      let Ok(instruction) = RefundableInInstruction::try_from(shorthand) else {
+        panic!("can't convert Shorthand to instruction")
+      };
+      match instruction.instruction {
+        InInstruction::Transfer(address) => assert_eq!(address, serai_address.into()),
+        _ => panic!("wrong instruction type"),
+      }
```
kayabaNerve: Can't you replace almost all of this with `assert_eq!(shorthand, message);`?

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
       }
     }
   }
+
+  // expected script has to start with SHA256 PUSH MSG_HASH OP_EQ ..
+  fn expected_script_pattern(&self, script: &ScriptBuf) -> Option<bool> {
```
kayabaNerve: `segwit_data_pattern`? This is only expected if not using OP_RETURN.

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
...
+
+    // next should be a equality check
+    if ins.next()?.ok()?.opcode()? != OP_EQUAL {
+      return Some(false);
+    }
+
+    Some(true)
+  }
+
+  async fn spent_output_of(&self, input: &TxIn) -> TxOut {
```
kayabaNerve: `_of` doesn't necessarily make grammatical sense here. This returns the spent output and that alone make sense.

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
...
+          }
+        }
+      }
+      vec![]
+    })();
+
+    // check inputs
+    if data.is_empty() {
+      for input in &tx.input {
+        if self.spent_output_of(input).await.script_pubkey.is_p2wsh() {
```
kayabaNerve: I don't believe this is a necessary check. Checking alone the witness is as expected *should* be sufficient. If this isn't *some* SegWit TX, then witness will be empty.

It does mean we may interpret a non-P2WSH witness as a P2WSH witness. I'm unsure that's an issue. The person who sent this TX still gave us something we can interpret *and they included it and therefore authorized it*.

Considering how expensive the spent_output check is (it's pulling an old TX which is a full RPC call and an extra user of the txindex functionality of the underlying node), I don't see the benefit unless there's an explicit issue at risk.

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
...
+          let witness = input.witness.to_vec();
+          let redeem_script = ScriptBuf::from_bytes(witness[1].clone());
```
kayabaNerve: [1] will cause a panic. This needs to be last() with a continue if it doesn't exist (somehow).

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
...
+          if let Some(true) = self.expected_script_pattern(&redeem_script) {
```
kayabaNerve: `== Some(true)`?

processor/src/networks/bitcoin.rs
```diff
@@ -450,6 +477,85 @@ impl Bitcoin {
...
+          }
+        }
+      }
+    }
+
+    data.truncate(MAX_DATA_LEN.try_into().unwrap());
+    data
+  }
+
+  async fn extract_origin(&self, tx: &Transaction) -> Option<Address> {
```
kayabaNerve: Is this worth moving to a dedicated function when there's a single caller?

processor/src/networks/bitcoin.rs
kayabaNerve: To be clear, we use txindex which adds GB of data to the Bitcoin node because it's currently impractical to track everything on our end. Ideally, yes, we don't need txindex to minimize data usage and improve performance.

If usage of it makes sense, I don't mind using it. If usage of it is extraneous (pushing us further away from being able to remove it), I can't claim to like it.

processor/src/networks/bitcoin.rs
kayabaNerve: spent_output also resolves to having a single caller depending on the above note.

processor/src/networks/bitcoin.rs
```diff
@@ -573,55 +679,24 @@ impl Network for Bitcoin {
...
+
+      // populate the rest of the outputs
+      let presumed_origin = self.extract_origin(tx).await;
+      let data = self.extract_serai_data(tx).await;
+      for output in &mut outputs {
+        if output.kind == OutputType::External {
+          output.data = data.clone();
+        }
+        output.presumed_origin = presumed_origin.clone();
+      }
```
kayabaNerve, line 703: Since this is Bitcoin, should we collapse all of these outputs?

I'm unsure why someone would want to do `n` identical actions in a single transaction. If we don't allow multiple segments of data (such as data defined as `vec![return_data, ...segwit_inputs]`) ,let's continue your cleanup and simplify this :)

kayabaNerve (changes requested):

Issue #478, opened 2023-12-08
Monero seed test failed in CI


kayabaNerve:
https://github.com/serai-dex/serai/actions/runs/7139662256/job/19443422741

```
failures:

---- tests::seed::test_classic_seed stdout ----
thread 'tests::seed::test_classic_seed' panicked at coins/monero/src/tests/seed.rs:180:88:
called `Result::unwrap()` on an `Err` value: InvalidChecksum
stack backtrace:
   0: rust_begin_unwind
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/std/src/panicking.rs:597:5
   1: core::panicking::panic_fmt
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/result.rs:1652:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/result.rs:1077:23
   4: monero_serai::tests::seed::test_classic_seed
             at ./src/tests/seed.rs:180:24
   5: monero_serai::tests::seed::test_classic_seed::{{closure}}
             at ./src/tests/seed.rs:17:24
   6: core::ops::function::FnOnce::call_once
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5
   7: core::ops::function::FnOnce::call_once
             at /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    tests::seed::test_classic_seed

test result: FAILED. 17 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.48s

error: test failed, to rerun pass `-p monero-serai --lib`
```

j-berman:
Where I'm at with this...

It's panicking in `seed.entropy()` in the second print statement from #487, so the failing seed doesn't actually print:

https://github.com/serai-dex/serai/blob/74a68c6f68e5fe0cacd69981b8081e49ab5f50b4/coins/monero/src/tests/seed.rs#L188-L193

Here are some failing seeds:

[...]

PR #487, opened 2023-12-12
seed: print seed info in tests

hinto-janai:
Prints the input seed info, which shows after panic.

Should help debug https://github.com/serai-dex/serai/issues/478.


kayabaNerve:
Thanks.

That issue will likely be fixed whenever someone sits down with a CPU core and grinds seeds out.

PR #497, opened 2023-12-21
Use TCP keep-alive in Monero RPC

kayabaNerve:
May resolve the prior observed spontaneous errors, with the retry code moved to just the authentication-stale case to check.

If this doesn't work, this should still be merged with the retry code as originally placed.

### DIFF
--- a/coins/monero/src/rpc/http.rs
+++ b/coins/monero/src/rpc/http.rs
@@ -97,7 +97,7 @@ impl HttpRpc {
         Err(RpcError::ConnectionError("invalid amount of passwords".to_string()))?;
       }
 
-      let client = Client::without_connection_pool(&url)
+      let client = Client::without_connection_pool(&url, true)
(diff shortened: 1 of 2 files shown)


kayabaNerve:
This PR messed up and did not set keep-alive as expected. It just calls set keep-alive, yet does in a way turning it off (as it already should be).

Despite this, the CI passes for the above two runs when it was prior too erratic to survive one?? Did we update some patch making this no longer needed? Is keep-alive not already off and that was the issue?

It can't be only retrying on stale auth is enough. The prior code which retried on any error still occasionally caused failures (just with acceptably low probability).

We also now wrap a connector instead of having hyper-rustls provide the connector. The provided connector has `enforce_scheme` set to `false`, which we don't touch. That could be it? Except the experiment with boring-ssl reproduced this error (suggesting its agnostic to the TLS layer used).

Or keep-alive is on-by-default yet breaks things here, or the above two runs are a statistical fluke. I have no idea.

Issue #473, opened 2023-12-07
Malicious validator set can spike the price right before a session rotation, enabling the new session to arbitrarily mint


kayabaNerve:
Posit:

1) The old session economically secures enough sriXYZ to spike the price down in its last block
2) The price oracle resets to the highest sustained price, AKA the lowest price
3) The new session can not mint without economic security constraints
4) Arbitragers acquire the 'cheap' sriXYZ, letting further, effectively unsecured, sriXYZ mints to capture the SRI used for the arbitrage

While all the sriXYZ created by the first session will be economically secured, effectively none of the additional sriXYZ by the second session will be. If the SRI captured in arbitrage exceeds 2/3rds of the margin (2/3rds of 20% of the sriXYZ value, which in this case I believe is the margin of the old session), this would be profitable.

Instead of using the highest sustained price, if we double the size of the window *and use the median price for the window*, this attack is infeasible. The price would have to be spiked down, and held down, under economic security, for the length of the window which presumably is enough time to commit arbitrage and cause unprofitability.

This also wouldn't raise the responsiveness to price increases, as if the price goes up for half of the doubled window and any increase is sustained, the sustained increased will still become the oracle value (since all prior values will be lower, forming the bottom half of the median).

The main annoyance would be in the median calculation. The naive impl is to create an iter, skipping the first `n` entries to reach the median. This would take `n` iterations to perform and is not sublinear.

This would also resolve https://github.com/serai-dex/serai/blob/d1d5ee6b3d1b1ce1d470673367462ec369970379/substrate/dex/pallet/src/lib.rs#L171-L173.

kayabaNerve:
Averages can be efficiently accumulated as followed:

`((existing_value * amount_of_items_accumulated) + accumulating_value) / (amount_of_items_accumulated + 1)`

With popping + pushing as:

`((existing_value * amount_of_items_accumulated) - popping_value + pushing_value) / amount_of_items_accumulated`

While we want a median, I don't immediately realize such an algorithm existing for medians. There may be a log n algorithm? Easy enough to look up prior published algorithms.

kayabaNerve:
Instead of decreasing upon new set, we can decrease upon new set after new set. While we accept there may be a malicious set after a set which acted malicious, we don't accept there may be a malicious set after a malicious set after a set which acted malicious, due to voting to halt handovers presumed to occur within that time period.

[...]

PR #498, opened 2023-12-24
use median price instead of the highest sustained

akildemir:
Addresses https://github.com/serai-dex/serai/issues/473


kayabaNerve:
CI failures were https://github.com/serai-dex/serai/commit/ec8dfd46394d8428f2dfdd1d978bdf7fa742c63e and latency causing a DKG removal (addressed by https://github.com/serai-dex/serai/commit/02776c54a804c4b3d8f91bbc94585434d37e0627).

substrate/primitives/src/lib.rs
```diff
@@ -143,5 +143,14 @@ impl AsRef<[u8]> for Data {
   }
 }
 
+/// Lexicographically reverses a given byte array.
+pub fn reverse_lexicographic_order<const N: usize>(bytes: [u8; N]) -> [u8; N] {
```
kayabaNerve, line 152: Let's take a moment and add a test this works for all 2-byte arrays?

Re-reviewing this, I'm not sure it's wrong, yet I'm also not sure I wrote the right algorithm. Substrate uses a nibble-based solution and this is a bit/byte-based. It may be simply as nibble order may be allowed to be consistent? Yet easiest to exhaustively test and call it a day.

Just populate two StorageMaps (value, reverse(value)) -> value and check they're as expected.

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
   /// Moving window of oracle prices.
   ///
   /// The second [u8; 8] key is the amount's big endian bytes, and u16 is the amount of inclusions
-  /// in this multi-set.
+  /// in this multi-set. Since the underlying map lexiogracpilly sorted, this map stores amounts
```
kayabaNerve: "is lexicographically" :P

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
...
-    /// all prices are greater than or equal to.
-    pub fn highest_sustained_price(coin: &Coin) -> Option<Amount> {
-      let mut iter = OraclePrices::<T>::iter_key_prefix(coin);
-      // the first key will be the lowest price due to the keys being lexicographically ordered.
-      iter.next().map(|amount| Amount(u64::from_be_bytes(amount)))
-    }
-  }
+
+  /// Same as `OraclePrices` but in reverse order, since the keys we save to this map are
+  /// lexiogracpilly reversed. Hence this map stores the amounts from hight to low.
```
kayabaNerve: "lexicographically"
"high to low"

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
...
+          if inserted >= current {
+            OraclePrices::<T>::iter_key_prefix_from(coin, current.to_be_bytes().to_vec()).next()
+          } else {
+            let reversed = reverse_lexicographic_order(current.to_be_bytes());
+            OraclePricesReverse::<T>::iter_key_prefix_from(coin, reversed.to_vec())
+              .next()
+              .map(reverse_lexicographic_order)
+          }
+        } else {
+          Some(current.to_be_bytes())
```
kayabaNerve, line 228: Let's document that if there's an even amount of items, there's two candidates and the current one remains a valid candidate.

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
...
+            .next()
+            .map(reverse_lexicographic_order)
+        } else {
+          // no change if inserted and removed both in same side(both smaller or bigger).
+          Some(current.to_be_bytes())
+        }
+      };
+
+      // TODO: use current or 0 for unwrap?
+      let next = next.unwrap_or(current.to_be_bytes());
```
kayabaNerve, line 253: Is there a way this can fail? AFAICT, it'd require current be the edge and we try to further move. That'd only fail if there's a single item? We can short-circuit this algorithm while length < 3 for such cases.

The failure point of this median algorithm is how any drift is never corrected and always accumulates. We definitely shouldn't flatline to 0 accordingly, yet remaining on current sounds like it'd introduce a drift.

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
...
+      inserted: u64,
+      removed: Option<u64>,
+      n: BlockNumberFor<T>,
+    ) {
+      let current = Self::median_price(coin).unwrap_or(Amount(0)).0;
+
+      // if we didn't remove  an amount from the window, just means we still didn't hit
+      // the full size of the window.
+      let next = if removed.is_none() {
+        if n % 2u32.into() == 1u32.into() {
```
kayabaNerve: This can't use `n` yet needs to use the length of the map as if we introduce new coins in the future, they won't open on 0 (which this assumes).

substrate/dex/pallet/src/lib.rs
```diff
@@ -362,35 +408,57 @@ pub mod pallet {
...
             if *v == Some(0) {
               *v = None;
             }
           });
-        }
+
+          // Remove this price from the high-to-low multiset
+          let reverse = reverse_lexicographic_order(start_spot_price);
+          OraclePricesReverse::<T>::mutate_exists(coin, reverse, |v| {
+            *v = Some(v.unwrap_or(1) - 1);
```
kayabaNerve: Shouldn't this just be an unwrap? There's no way for it to not be here if it's being removed.

substrate/dex/pallet/src/lib.rs
```diff
@@ -161,29 +161,75 @@ pub mod pallet {
...
+    StorageDoubleMap<_, Identity, Coin, Identity, [u8; 8], u16, OptionQuery>;
 
   #[pallet::storage]
   #[pallet::getter(fn oracle_value)]
   pub type OracleValue<T: Config> = StorageMap<_, Identity, Coin, Amount, OptionQuery>;
 
+  /// Current median price of the prices in the oracle prices maps at any given time.
+  #[pallet::storage]
+  #[pallet::getter(fn median_price)]
+  pub type MedianPrice<T: Config> = StorageMap<_, Identity, Coin, Amount, OptionQuery>;
```
kayabaNerve, line 192: Let's also add a test for this :+1:

kayabaNerve (commented):
Should have a pair of tests, yet looks like the proper ideas and a sane stab at this :)

PR #503, opened 2024-01-03
monero: only mask user features on new polyseed, not on decode

j-berman:
- This commit ensures a polyseed string that has unsupported features correctly errors on decode (without this PR, the test in this PR would trigger a debug assert, and if instead triggered in a non-debug build then `Polyseed::from_string` could return an incorrect successful response)
- Also avoids panicking in debug builds when checksum calculation is unexpectedly wrong

Polyseed reference impl for feature masking:
- polyseed_create: https://github.com/tevador/polyseed/blob/b7c35bb3c6b91e481ecb04fc235eaff69c507fa1/src/polyseed.c#L61
- polyseed_decode: https://github.com/tevador/polyseed/blob/b7c35bb3c6b91e481ecb04fc235eaff69c507fa1/src/polyseed.c#L212

### DIFF
--- a/coins/monero/src/tests/seed.rs
+++ b/coins/monero/src/tests/seed.rs
@@ -7,7 +7,7 @@ use curve25519_dalek::scalar::Scalar;
 use crate::{
   hash,
   wallet::seed::{
-    Seed, SeedType,
+    Seed, SeedType, SeedError,
     classic::{self, trim_by_lang},
     polyseed,
   },
@@ -404,4 +404,13 @@ fn test_polyseed() {
       );
     }
   }
+}
+
+#[test]
+fn test_invalid_polyseed() {
+  let seed = "include domain claim resemble urban hire lunch bird \
+    crucial fire best wife ring warm ignore model"
+    .into();
+  let res = Seed::from_string(Zeroizing::new(seed));
+  assert_eq!(res, Err(SeedError::UnsupportedFeatures));
 }
--- a/coins/monero/src/tests/seed.rs
+++ b/coins/monero/src/tests/seed.rs
@@ -408,6 +408,7 @@ fn test_polyseed() {
 
 #[test]
 fn test_invalid_polyseed() {
+  // This seed includes unsupported features bits and should error on decode
   let seed = "include domain claim resemble urban hire lunch bird \
     crucial fire best wife ring warm ignore model"
     .into();
--- a/coins/monero/src/wallet/seed/polyseed.rs
+++ b/coins/monero/src/wallet/seed/polyseed.rs
@@ -179,6 +179,31 @@ fn valid_entropy(entropy: &Zeroizing<[u8; 32]>) -> bool {
   res.into()
 }
 
+fn from_internal(
+  language: Language,
+  masked_features: u8,
+  encoded_birthday: u16,
+  entropy: Zeroizing<[u8; 32]>,
+) -> Result<Polyseed, SeedError> {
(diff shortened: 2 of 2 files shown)


kayabaNerve:
> Also avoids panicking in debug builds when checksum calculation is unexpectedly wrong

This is a feature, not a bug.

coins/monero/src/tests/seed.rs
```diff
@@ -405,3 +405,12 @@ fn test_polyseed() {
     }
   }
 }
+
+#[test]
+fn test_invalid_polyseed() {
```
kayabaNerve, line 410: I'd clarify how it's invalid.

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -179,6 +179,31 @@ fn valid_entropy(entropy: &Zeroizing<[u8; 32]>) -> bool {
   res.into()
 }
 
+fn from_internal(
```
kayabaNerve: Is there a reason to have this outside the Polyseed impl?

coins/monero/src/wallet/seed/polyseed.rs
```diff
@@ -375,9 +387,12 @@ impl Polyseed {
     let features =
       u8::try_from(extra >> DATE_BITS).expect("couldn't convert extra >> DATE_BITS to u8");
 
-    let res = Polyseed::from(lang, features, birthday_decode(birthday), entropy);
+    let res = from_internal(lang, features, birthday, entropy);
     if let Ok(res) = res.as_ref() {
-      debug_assert_eq!(res.checksum, checksum);
+      if res.checksum != checksum {
+        // This should never trigger
```
kayabaNerve: If this should never trigger, the debug_assert is valid.

coins/monero/src/wallet/seed/polyseed.rs
j-berman: `from_string` will first decode the string and yield the already masked features before passing it over to `from_internal`. The features shouldn't be masked again (the primary bug this PR is fixing).

`from` expects an *unmasked* `features`, which it then masks for the caller. I matched `polyseed_create` from the polyseed spec for this behavior, which also takes in unmasked `features` and masks it for the caller. Source: https://github.com/tevador/polyseed/blob/b7c35bb3c6b91e481ecb04fc235eaff69c507fa1/src/polyseed.c#L61 . I figured it would make sense that monero-serai's `from` would match the polyseed spec for this behavior since it's the only exposed function that a caller can pass `features` into.

Because of the above 2 distinct call paths, I pulled out the `from_internal` and had it take in the *already masked* features.

Sort of similar line of thinking for the birthday btw (`from_string` yields the encoded birthday so it's unnecessary to decode it, whereas `from` expects the decoded birthday which needs to be encoded).

coins/monero/src/wallet/seed/polyseed.rs
j-berman: Ok, why should it be a debug assert and not a normal assert? I see no reason the function should return a normal response in a non-debug build if this statement is true. The seed would be invalid.

j-berman: I guess debug assert in one line, then return the invalid seed error in the next could make sense.

kayabaNerve: Because it should be impossible, and is accordingly redundant, so it should be optimized out on release.

PR #506, opened 2024-01-05
Check miner tx has a miner input when deserializing.

Boog900:


kayabaNerve:
This doesn't resolve the panic noted in `fn number` as it's still possible to manually construct a block without a miner TX. While I'm fine merging this, do you want to fix that in this PR, in a new PR, or shall I follow up there? (any is fine with me)

Boog900:
I didn't want to change the API of the `number` function and IMO if someone is creating invalid blocks *manually* panicking is probably ok (should be documented though). I had a look and I think it would probably be best for you to decide how to handle invalid miner txs in `number` as its used in places where i'm unsure what should be done.

Issue #102, opened 2022-08-21
Monero Library Tracking Issue


kayabaNerve:
This is solely discussing the Monero library, and explicitly NOT the FROST aspects which are pending the IETF draft finalization and so on.

For all open issues: https://github.com/serai-dex/serai/issues?q=is%3Aopen+is%3Aissue+label%3Amonero

The following table is meant to go over all of them, along with other impactful issues.

Issue | Required | Status          | Notes                                     |
------|----------|-----------------|-------------------------------------------|
#104  | No       | Not Started     | Potentially would never be used by anyone |
#207  | Yes      | Not Started     |                                           |
#210  | No       | Not Started     |                                           |
#265  | Yes      | Not Started     |                                           |
#266  | Yes      | Awaiting Review |                                           |
#273  | No       | Not Started     |                                           |
#415  | Yes      | Not Started     |                                           |
#478  | Yes      | Not Started     |                                           |

If any issue is closed, it'll be removed from the table. If a new issue is created, it should be added.

Additional tasks:

- Deser/ser all TXs/blocks on mainnet/testnet to ensure we can.

Additional requirements:

- Remaining TODOs.
- Extensive unit testing.
- External review.

Additional notes:

kayabaNerve:
~~This library has TODOs which also must be revisited.~~ Moved into a new "Additional Tasks" section.

kayabaNerve:
https://github.com/serai-dex/serai/commit/8a24fc39a61633fb774f806cd3a194e5d5a7788b fixed an issue where scanned funds may be unspendable via monero-serai. This was possible to reach on the Monero mainnet due to some very specific protocol complexities, which we prior accommodated yet I missed the scanner didn't also accommodate.

Yanking all current releases accordingly.

kayabaNerve:
The newly added reserialize_chain binary ensures we can ser/deser all blocks/TXs. I'm running it now.

kayabaNerve:
We have `.decompress` calls when we should only use `read_point` as `.decompress` is less strict than `read_point` which matches Monero.

kayabaNerve:
We should test read_point for such cases...

kayabaNerve:
Closing. Only #104 stands and the library is effectively at the 1.0 rc, being queued for audits. I don't believe the dedicated tracking issue here is better than the tag.

PR #515, opened 2024-01-11
monero: match monero's stricter check when decompressing points

j-berman:
https://github.com/serai-dex/serai/issues/102#issuecomment-1627474637
https://github.com/serai-dex/serai/issues/102#issuecomment-1627475063


coins/monero/src/wallet/send/mod.rs
```diff
@@ -787,7 +787,7 @@ impl SignableTransaction {
       fee -= output.commitment.amount;
       tx_outputs.push(Output {
         amount: None,
-        key: output.dest.compress(),
+        key: output.dest.compress().to_bytes(),
```
kayabaNerve: Why did you remove the typing from CompressedEdwardsY to [u8; 32]?

j-berman: Reverted

kayabaNerve:
LGTM on being consistent :)

PR #519, opened 2024-01-16
hyper-rustls 0.26

kayabaNerve:
Updates to hyper 1.0, which is unfortunately currently duplicated in our tree due to other parties which have yet to update.


kayabaNerve:
For record keeping, it's primarily bollard (dockertest) and reqwest (ethers-providers) which cause the duplicate tree.

PR #520, opened 2024-01-21
complete various todos in tributary

akildemir:


coordinator/tributary/src/tendermint/mod.rs
```diff
@@ -275,9 +275,14 @@ pub struct TendermintNetwork<D: Db, T: TransactionTrait, P: P2p> {
 
 pub const BLOCK_PROCESSING_TIME: u32 = 999;
 pub const LATENCY_TIME: u32 = 1667;
-// TODO: Add test asserting this
 pub const TARGET_BLOCK_TIME: u32 = BLOCK_PROCESSING_TIME + (3 * LATENCY_TIME);
 
+#[test]
+fn assert_target_block_time_no_overflow() {
```
kayabaNerve: This isn't about checking it doesn't overflow. It's about checking that the actual target block time *from the `tendermint-machine` crate* is equal to this.

coordinator/tributary/src/tendermint/tx.rs
```diff
@@ -83,18 +83,15 @@ fn decode_and_verify_signed_message<N: Network>(
 }
 
 // TODO: Move this into tendermint-machine
-// TODO: Strongly type Evidence, instead of having two messages and no idea what's supposedly
-// wrong with them. Doing so will massively simplify the auditability of this (as this
-// re-implements an entire foreign library's checks for malicious behavior).
+// This function takes a TendermintTx, which can't be imported to tendermint create.
```
kayabaNerve: 1) This comment is contradictory, assuming it's part of the PR.
2) The underlying evidence checking can be moved.

coordinator/tributary/src/tendermint/mod.rs
```diff
@@ -343,6 +348,7 @@ impl<D: Db, T: TransactionTrait, P: P2p> Network for TendermintNetwork<D, T, P>
 
     // add tx to blockchain and broadcast to peers
     // TODO: Make a function out of this following block
+    // Why? Seems like a single caller.
```
kayabaNerve: If it's still a single caller and doesn't appear to have common value, then likely an old comment we can remove :+1:

coordinator/tributary/src/tendermint/tx.rs
akildemir: oh yes the evidence checking can be moved I thought TODO meant the whole function as is. 👍

coordinator/tributary/src/tendermint/mod.rs
akildemir: yea was not sure either:D 👍

PR #530, opened 2024-02-04
Redo Dockerfile generation

kayabaNerve:
PR opened as this should be equivalent to develop.

Further work is justified before merging.


orchestration/dev/coins/monero/run.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
```
rlking, line 2: the alpine image doesn't have a bash by default

kayabaNerve: Fun. Guess I'll consolidate to /sh :sweat_smile:  Thanks.

I pushed these changes as I went to bed, will review the accumulated CI errors now.

PR #531, opened 2024-02-06
Remove unused brew packages on macOS

kayabaNerve:
Attempts to resolve the random SIGABRTs we're getting in CI.


kayabaNerve:
This also added a m1 clippy which is ~2x as fast.

This had the Intel macOS and ARM macOS pass, so I'm assuming it was a storage limit (as GH runners for macOS have a fraction of the storage as Linux/Windows) now resolved.

We can be more aggressive in the future by writing `rm -rf ~/Library/Android` as well.

PR #532, opened 2024-02-12
Add validator set rotation test for the node side

akildemir:


substrate/client/src/serai/mod.rs
```diff
@@ -195,6 +195,19 @@ impl Serai {
...
 
+  pub async fn active_network_validators(
+    &self,
+    network: NetworkId,
+  ) -> Result<Vec<Public>, SeraiError> {
+    let hash: String = self
+      .call("state_call", ["SeraiRuntimeApi_validators".to_string(), hex::encode(network.encode())])
+      .await?;
+    let bytes = Self::hex_decode(hash)?;
+    let r = decode_from_bytes::<Vec<Public>>(bytes.into())
```
kayabaNerve: Why use decode_from_bytes instead of `Vec::<Public>::decode`?

substrate/client/src/serai/mod.rs
```diff
@@ -195,6 +195,19 @@ impl Serai {
     Ok(())
   }
 
+  pub async fn active_network_validators(
```
kayabaNerve: Do we want to add this fn over the existing ValidatorSets calls?

substrate/client/src/serai/mod.rs
```diff
@@ -195,6 +195,19 @@ impl Serai {
...
+    &self,
+    network: NetworkId,
+  ) -> Result<Vec<Public>, SeraiError> {
+    let hash: String = self
+      .call("state_call", ["SeraiRuntimeApi_validators".to_string(), hex::encode(network.encode())])
```
kayabaNerve, line 200: Do we want to expose this when it's primarily there to satisfy BABE/GRANDPA/authority-discovery (not our own API)? We'd be committing to its usage.

substrate/client/tests/validator_sets.rs
```diff
@@ -8,11 +8,13 @@ use serai_client::{
...
   },
-  Serai,
+  Amount, Serai,
 };
 
 mod common;
-use common::validator_sets::set_keys;
+use common::validator_sets::{set_keys, allocate_stake, deallocate_stake};
+
+const EPOCH_INTERVAL: u64 = 5;
```
kayabaNerve, line 17: We can likely modify our Substrate to take the epoch length as an argument over the CLI. It'd be part of the replace_cmd below (as config is on the node-side, not the runtime-side).

substrate/client/tests/validator_sets.rs
```diff
@@ -73,3 +75,198 @@ serai_test!(
...
+    let mut finalized_block = serai.latest_finalized_block().await.unwrap().header.number;
+    while finalized_block <= epoch_block + 2 {
+      tokio::time::sleep(tokio::time::Duration::from_secs(6)).await;
+      finalized_block = serai.latest_finalized_block().await.unwrap().header.number;
+    }
+  })
+  .await
+  .unwrap();
+
+  // TODO: verfiy key shares as well?
```
kayabaNerve: Two instances of "verfiy" in comments and the function name.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -688,16 +688,20 @@ pub mod pallet {
     }
 
     pub fn retire_set(set: ValidatorSet) {
-      let keys = Keys::<T>::take(set).unwrap();
-      // If the prior prior set didn't report, emit they're retired now
+      // If the prior set didn't report, emit they're retired now
```
kayabaNerve: No, this is prior prior.

The prior set is the one now eligible to report.

kayabaNerve (commented):
Very minor comments, except for discussion over the decision to use a feature instead of a config argument on the node-side with a CLI arg.

substrate/client/src/serai/mod.rs
akildemir: is it best if we do this private and add a public one to ValidatorSets calls?  I get that it makes sense for it to be called through ValidatorSets calls but the implementation is on the same level as base serai client since it makes raw calls and needs helper functions from there.

substrate/client/src/serai/mod.rs
akildemir: We will aways need an api to get the active network validators in the future, this is as good as any? Alternative is we implement it on the vs pallet and call that fn from here(to prevent code duplication) which I think achieves the same thing with extra steps. And I don't think we want to fetch the raw participants from storage and play with the session to figure out the active ones per network rule whenever we need the set participants.

PR #535, opened 2024-02-26
Add coordinator rotation test

akildemir:


tests/coordinator/src/tests/key_gen.rs
```diff
@@ -23,37 +23,48 @@ use crate::tests::*;
...
+      match &msg {
+        CoordinatorMessage::KeyGen(messages::key_gen::CoordinatorMessage::GenerateKey {
+          params,
+          ..
+        }) => {
+          participant_is.push(params.i());
+          found = true;
+        }
+        CoordinatorMessage::Substrate(
+          messages::substrate::CoordinatorMessage::ConfirmKeyPair { .. },
```
kayabaNerve: Why is this block here? I'm unclear why we're not deterministically handling ConfirmKeyPair and instead are ad-hoc dropping them.

PR #534, opened 2024-02-24
Creation of new Debian 12 guide for installation of validator 

bennyb0y:
Created a new document to describe the process of installing and operating a dev node validator. This guide is geared to Debian 12.5.0 but may work with others. Not sure where this guide might best live, so I will leave it here for now. 

Removed some redundant commands from the original Getting started.md

PR #542, opened 2024-03-12
Create Debian 12 validator install.md

bennyb0y:
Installation of a validator instance, from a clean Debian 12.5.

Not sure where this belongs, it seems like the docs folder is getting more organized. Happy to help with other content.


kayabaNerve:
Supersedes #534.

PR #545, opened 2024-03-27
Implement genesis liquidity protocol

akildemir:


substrate/abi/src/genesis_liquidity.rs
```diff
@@ -0,0 +1,21 @@
+pub use serai_genesis_liquidity_primitives as primitives;
+
+use serai_primitives::*;
+use primitives::*;
+
+#[derive(Clone, PartialEq, Eq, Debug, scale::Encode, scale::Decode, scale_info::TypeInfo)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum Call {
+  remove_coin_liquidity { balance: Balance },
+  set_initial_price { prices: Prices, signature: Signature },
```
kayabaNerve: We aren't setting the initial price. We're oraclizing the value of external coins in relation to Bitcoin.

substrate/abi/src/genesis_liquidity.rs
```diff
@@ -0,0 +1,21 @@
...
+}
+
+#[derive(Clone, PartialEq, Eq, Debug, scale::Encode, scale::Decode, scale_info::TypeInfo)]
+#[cfg_attr(feature = "borsh", derive(borsh::BorshSerialize, borsh::BorshDeserialize))]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum Event {
+  GenesisLiquidityAdded { by: SeraiAddress, balance: Balance },
+  GenesisLiquidityRemoved { by: SeraiAddress, balance: Balance },
+  GenesisLiquidityAddedToPool { coin1: Balance, coin2: Balance },
```
kayabaNerve, line 19: Isn't this always `coin, sri`?

substrate/abi/src/liquidity_tokens.rs
```diff
@@ -0,0 +1,18 @@
+use serai_primitives::{Balance, SeraiAddress};
```
kayabaNerve, line 1: This file was added yet I don't see it used at this time?

substrate/abi/src/genesis_liquidity.rs
```diff
@@ -0,0 +1,21 @@
...
+  EconomicSecurityReached { network: NetworkId },
```
kayabaNerve, line 20: Genesis Liquidity pallet is also handling EcSec?

substrate/client/src/serai/dex.rs
```diff
@@ -57,4 +59,21 @@ impl<'a> SeraiDex<'a> {
       send_to: address,
     })
   }
+
+  pub async fn get_reserves(
+    &self,
+    coin1: Coin,
+    coin2: Coin,
```
kayabaNerve: One will always be SRI, right?

substrate/client/src/serai/dex.rs
```diff
@@ -57,4 +59,21 @@ impl<'a> SeraiDex<'a> {
...
+  ) -> Result<Option<(Amount, Amount)>, SeraiError> {
+    let hash = self
+      .0
+      .serai
+      .call("state_call", ["DexApi_get_reserves".to_string(), hex::encode((coin1, coin2).encode())])
+      .await?;
+    let bytes = hex_decode(hash)
```
kayabaNerve: "hash" when we're not fetching a hash yet the "reserves"

substrate/client/src/serai/dex.rs
```diff
@@ -57,4 +59,21 @@ impl<'a> SeraiDex<'a> {
...
+      .map_err(|_| SeraiError::InvalidNode("expected hex from node wasn't hex".to_string()))?;
+    let resut = decode_from_bytes::<Option<(u64, u64)>>(bytes.into())
+      .map_err(|e| SeraiError::ErrorInResponse(e.to_string()))?;
+    Ok(resut.map(|amounts| (Amount(amounts.0), Amount(amounts.1))))
```
kayabaNerve: "resut"

substrate/runtime/src/lib.rs
```diff
@@ -290,7 +295,7 @@ pub type ReportLongevity = <Runtime as pallet_babe::Config>::EpochDuration;
 
 impl babe::Config for Runtime {
   #[cfg(feature = "fast-epoch")]
-  type EpochDuration = ConstU64<{ MINUTES / 2 }>; // 30 seconds
+  type EpochDuration = ConstU64<{ HOURS / 2 }>; // 30 minutes
```
kayabaNerve: Is this what it's supposed to be? Double checking after recent bugs from this :p

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+use sp_std::vec::Vec;
+
+use scale::{Encode, Decode, MaxEncodedLen};
+use scale_info::TypeInfo;
+
+use serai_primitives::*;
+use validator_sets_primitives::ValidatorSet;
+
+// amount of blocks in 30 days for 6s per block.
+pub const BLOCKS_PER_MONTH: u32 = 10 * 60 * 24 * 30;
```
kayabaNerve: The runtime defines `TARGET_BLOCK_TIME, DAYS`. It's probably best to move those into serai-primitives and set this to `30 * DAYS`? Same with the other constants.

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// 180 days of blocks
+pub const GENESIS_SRI_TRICKLE_FEED: u64 = 10 * 60 * 24 * 180;
+
+// 100 Million SRI
+pub const GENESIS_SRI: u64 = 100_000_000 * 10_u64.pow(8);
```
kayabaNerve: This is 8 decimals which is traditional, yet would we prefer 6 to align with SI units? The economics document does say `.000000`, this I believe this was a prior discussion.

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,47 @@
...
+
+// This is the account to hold and manage the genesis liquidity.
+pub const GENESIS_LIQUIDITY_ACCOUNT: SeraiAddress = system_address(b"Genesis-liquidity-account");
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Zeroize))]
+#[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Prices {
+  pub bitcoin: u64,
```
kayabaNerve: If in terms of Bitcoin, won't this field simply be 1e8?

substrate/client/src/serai/mod.rs
```diff
@@ -82,6 +87,14 @@ impl<'a> Clone for TemporalSerai<'a> {
   }
 }
 
+pub fn hex_decode(str: String) -> Result<Vec<u8>, FromHexError> {
```
kayabaNerve: Is this a merge artifact? I'm not sure why it was moved to an independent function.

It is no longer under `Self` (and that's fine, as it doesn't use `self`), but the return type makes calling requiring an explicit error map. I believe I moved it under Self when I bundled the error map (which would explain this if it was a merge artifact).

substrate/client/tests/common/mod.rs
```diff
@@ -66,3 +66,67 @@ macro_rules! serai_test {
     )*
   }
 }
+
+#[macro_export]
+macro_rules! serai_test_fast_epoch {
```
kayabaNerve, line 71: Why is this on this PR? I'd guess some merge artifact, apologies there, I know it's due to my delays. If that's the sole reason, I can remove this on my end.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+  use dex_pallet::{Pallet as Dex, Config as DexConfig};
+  use coins_pallet::{Config as CoinsConfig, Pallet as Coins, AllowMint};
+
+  use serai_primitives::*;
+  use validator_sets_primitives::{ValidatorSet, Session, musig_key};
+  pub use genesis_liquidity_primitives as primitives;
+  use primitives::*;
+
+  /// LiquidityTokens Pallet as an instance of coins pallet.
+  pub type LiquidityTokens<T> = coins_pallet::Pallet<T, coins_pallet::Instance1>;
```
kayabaNerve, line 25: Should this be a config argument somehow?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+
+  #[pallet::validate_unsigned]
+  impl<T: Config> ValidateUnsigned for Pallet<T> {
+    type Call = Call<T>;
+
+    fn validate_unsigned(_: TransactionSource, call: &Self::Call) -> TransactionValidity {
+      match call {
+        Call::set_initial_price { ref prices, ref signature } => {
+          let set = ValidatorSet { network: NetworkId::Serai, session: Session(0) };
+          let signers = Participants::<T>::get(NetworkId::Serai);
```
kayabaNerve: Can't we fetch this from the validator_sets pallet?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+    #[pallet::call_index(1)]
+    #[pallet::weight((0, DispatchClass::Operational))] // TODO
+    pub fn set_initial_price(
+      origin: OriginFor<T>,
+      prices: Prices,
+      _signature: Signature,
+    ) -> DispatchResult {
+      ensure_none(origin)?;
+
+      // set the prices
```
kayabaNerve: Lack of a check this hasn't been called before.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+          account_values.insert(coin, vec![]);
+          let mut pool_amount: u128 = 0;
+          for (account, amount) in Liquidity::<T>::iter_prefix(coin) {
+            pool_amount = pool_amount.saturating_add(amount.into());
+            let value_this_addr =
+              u128::from(amount).saturating_mul(value.into()) / 10u128.pow(coin.decimals());
+            account_values.get_mut(&coin).unwrap().push((account, value_this_addr))
+          }
+          // sort, so that everyone has a consistent accounts vector per coin
+          account_values.get_mut(&coin).unwrap().sort();
```
kayabaNerve: 1) The runtime is fully deterministic so even if there was something traditionally with randomness, it shouldn't be possible here.

2) Iterating COINS is ordered, as iteration of the DB, and then the BTreeMap is a sorted data structure as well.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+            continue;
+          }
+
+          // initial coin value in terms of btc
+          let value = Oracle::<T>::get(coin);
+
+          // get the pool & individual address values
+          account_values.insert(coin, vec![]);
+          let mut pool_amount: u128 = 0;
+          for (account, amount) in Liquidity::<T>::iter_prefix(coin) {
```
kayabaNerve: We should be able to do this without an O(n) iteration of all accounts (which could be thousands) if we track account shares of genesis liquidity as we track liquidity tokens with relation to a pool.

All coins would be added to the pool, with all users having some share of the resulting initial liquidity.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// Add genesis liquidity for the given account. All accounts that provide liquidity
+    /// will receive the genesis SRI according to their liquidity ratio.
+    pub fn add_coin_liquidity(account: PublicKey, balance: Balance) -> DispatchResult {
+      // check we are still in genesis period
+      if Self::genesis_ended() {
+        Err(Error::<T>::GenesisPeriodEnded)?;
+      }
+
+      // mint the coins
+      Coins::<T>::mint(GENESIS_LIQUIDITY_ACCOUNT.into(), balance)?;
```
kayabaNerve: Prior mints were done in the II pallet. We may want to keep that pattern?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+      Liquidity::<T>::set(balance.coin, account, Some(new));
+
+      Self::deposit_event(Event::GenesisLiquidityAdded { by: account.into(), balance });
+      Ok(())
+    }
+
+    /// Returns the number of blocks since the coin's network reached economic security first time.
+    /// If the network is yet to be reached that threshold, 0 is returned. And maximum of
+    /// `GENESIS_SRI_TRICKLE_FEED` returned.
+    fn blocks_since_ec_security(coin: &Coin) -> u64 {
```
kayabaNerve: `-> Option<u64>`

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+      let ec_security_block =
+        EconomicSecurityReached::<T>::get(coin.network()).saturated_into::<u64>();
+      let current = <frame_system::Pallet<T>>::block_number().saturated_into::<u64>();
+      if ec_security_block > 0 {
+        let diff = current - ec_security_block;
+        if diff > GENESIS_SRI_TRICKLE_FEED {
```
kayabaNerve: There are `min`/`max` methods on numbers.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+          1,
+          GENESIS_LIQUIDITY_ACCOUNT.into(),
+        )?;
+        let current_sri = Coins::<T>::balance(GENESIS_LIQUIDITY_ACCOUNT.into(), Coin::Serai);
+        let current_coin = Coins::<T>::balance(GENESIS_LIQUIDITY_ACCOUNT.into(), balance.coin);
+
+        // burn the SRI if necessary
+        let mut sri = current_sri.0.saturating_sub(prev_sri.0);
+        let distance_to_full_pay =
+          GENESIS_SRI_TRICKLE_FEED - Self::blocks_since_ec_security(&balance.coin);
```
kayabaNerve: It's blocks since all coins reached economic security, not since any.

Also, instead of capping `blocks_since_ec_security` to `GENESIS_SRI_TRICKLE_FEED`, we can use a saturating sub here so the blocks fn returns the actual value yet this won't panic.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        return diff;
+      }
+
+      0
+    }
+
+    fn genesis_ended() -> bool {
+      <frame_system::Pallet<T>>::block_number() >= BLOCKS_PER_MONTH.into()
```
kayabaNerve: Is genesis over if we have yet to oraclize values w.r.t. Bitcoin?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+    fn validate_unsigned(_: TransactionSource, call: &Self::Call) -> TransactionValidity {
+      match call {
+        Call::set_initial_price { ref prices, ref signature } => {
+          let set = ValidatorSet { network: NetworkId::Serai, session: Session(0) };
+          let signers = Participants::<T>::get(NetworkId::Serai);
+
+          if !musig_key(set, &signers).verify(&set_initial_price_message(&set, prices), signature) {
+            Err(InvalidTransaction::BadProof)?;
+          }
+
```
kayabaNerve, line 456: This should check this hasn't been prior called and that the month is up.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+        // burn the SRI if necessary
+        let mut sri = current_sri.0.saturating_sub(prev_sri.0);
+        let distance_to_full_pay =
+          GENESIS_SRI_TRICKLE_FEED - Self::blocks_since_ec_security(&balance.coin);
+        let burn_sri_amount = sri.saturating_mul(distance_to_full_pay) / GENESIS_SRI_TRICKLE_FEED;
+        Coins::<T>::burn(
+          origin.clone().into(),
+          Balance { coin: Coin::Serai, amount: Amount(burn_sri_amount) },
+        )?;
+        sri -= burn_sri_amount;
```
kayabaNerve: This section doesn't cover movement between pools as necessary. Let's at least add a TODO for that?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,394 @@
...
+        let existing = Liquidity::<T>::get(balance.coin, account).unwrap_or(0);
+        if balance.amount.0 > existing || balance.amount.0 == 0 {
+          Err(Error::<T>::NotEnoughLiquidity)?;
+        }
+        if balance.amount.0 < existing {
+          Err(Error::<T>::CanOnlyRemoveFullAmount)?;
+        }
+
+        // TODO: do external transfer instead for making it easier for the user?
+        // or do we even want to make it easier?
```
kayabaNerve: We do want to make it easier. If they're only removing the coin, an ext transfer probably makes sense? Yet if they're removing both, it doesn't, and we have to cover that case.

I think transferring to a SRI account is fine. If they can call this fn, they have SRI and can call the next fn. The larger issue is needing SRI to call this fn in the first place.

kayabaNerve (commented):
Initial review for now. I have yet to read over the test. Mainly concerned about algorithmic complexity for the genesis distribution.

substrate/abi/src/genesis_liquidity.rs
akildemir: I think this is the what the meaning of "set" is, just doesn't also explains what it is setting to, which can be coin value in terms of btc or anything else. So what would be the correct naming? oraclize_value_in_terms_of_btc?

substrate/abi/src/genesis_liquidity.rs
akildemir: yeap but sri amounts are different

substrate/abi/src/genesis_liquidity.rs
akildemir: There is no handling of EcSec. It is just a milestone that genesis liquidty pallet checks for. These events are kind of arbitrary, so can be removed if wanted.

substrate/abi/src/liquidity_tokens.rs
akildemir: Ah yes, because `abi/lib.rs` and `client/src/serai/liquidity_tokens.rs` was still using the coins-pallet types. Will be fixed.

substrate/client/src/serai/dex.rs
akildemir: ah yes

substrate/genesis-liquidity/pallet/src/lib.rs
akildemir: I don't have a strong opinion on this but I would rather keep it here? The prior check is here. Also the reason previous mints were there was because mint account belongs there. Also prob because we didn't want to further modify(adding extra functions etc.) Dex pallet? Again can move it to there with the check if that makes more sense.

substrate/genesis-liquidity/primitives/src/lib.rs
akildemir: yea, should we just hardcode it like that?

substrate/genesis-liquidity/pallet/src/lib.rs
akildemir: iirc we tried doing that and it hadn't worked

akildemir: iirc we tried doing that and it hadn't worked

substrate/runtime/src/lib.rs
akildemir: lol. It was set to be 30 min because I knew it was like this in the previous pr. So I wanted to make sure the test passes with the same epoch time. But apparently it works just as good with 30 sec so we don't need this change 👍

substrate/abi/src/genesis_liquidity.rs
kayabaNerve: `oraclize_values` sounds fine to me. Method documentation can detail it's the value of one unit relative to Bitcoin.

substrate/abi/src/genesis_liquidity.rs
kayabaNerve: ... But why `coin1, coin2` and not `sri: Amount, coin: Balance`?

substrate/abi/src/genesis_liquidity.rs
kayabaNerve: Prob best to distinguish genesis from pre-econ and leave determining econ sec to the latter. That may require some trait/need to be followed up in a later PR though.

substrate/genesis-liquidity/primitives/src/lib.rs
kayabaNerve: Probably. That helps clarify we're oraclizing in terms of bitcoin.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+  #[pallet::storage]
+  pub(crate) type Oracle<T: Config> = StorageMap<_, Identity, Coin, u64, OptionQuery>;
+
+  #[pallet::storage]
+  pub(crate) type GenesisComplete<T: Config> = StorageValue<_, bool, OptionQuery>;
+
+  #[pallet::hooks]
+  impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    fn on_finalize(n: BlockNumberFor<T>) {
+      #[cfg(feature = "fast-epoch")]
```
kayabaNerve, line 80: 1) Do we want to overload fast-epoch?
2) Would it make more sense to do a constant in the config for this?

The latter benefits us if we ever want to do a distinct runtime entirely (not just a feature flagged one) for tests (as we'd just specify the constants in the definitions of `test-runtime/src/lib.rs`). I don't love the idea of another runtime, but as complexity increases, we should further discuss how to version/feature this.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+  #[pallet::storage]
+  pub(crate) type Oracle<T: Config> = StorageMap<_, Identity, Coin, u64, OptionQuery>;
+
+  #[pallet::storage]
+  pub(crate) type GenesisComplete<T: Config> = StorageValue<_, bool, OptionQuery>;
+
+  #[pallet::hooks]
+  impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+    fn on_finalize(n: BlockNumberFor<T>) {
```
kayabaNerve: on_initialize is preferred as it allows properly estimating gas (since the gas cost is known prior to executing TXs, you can filter the included TXs). I don't think anything here needs to run at end of block?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+          }
+
+          // initial coin value in terms of btc
+          let Some(value) = Oracle::<T>::get(coin) else {
+            continue;
+          };
+
+          let pool_amount = u128::from(Supply::<T>::get(coin).unwrap_or((0, 0)).1);
+          let pool_value = pool_amount.saturating_mul(value.into()) / 10u128.pow(coin.decimals());
+          total_value = total_value.saturating_add(pool_value);
```
kayabaNerve: I do appreciate the effort to use saturating here to prevent panics, yet also, these should panic. pool amount and value are u64 so they can't overflow a u128. If total value exceeds u128, the calculations will be fucked (as the sum of the pool value now exceeds total value). That should be an invariant panicked upon.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+
+        // add the liquidity per pool
+        let mut total_sri_distributed = 0;
+        let pool_values_len = pool_values.len();
+        for (i, (coin, pool_amount, pool_value)) in pool_values.into_iter().enumerate() {
+          // whatever sri left for the last coin should be ~= it's ratio
+          let sri_amount = if i == (pool_values_len - 1) {
+            GENESIS_SRI - total_sri_distributed
+          } else {
+            u64::try_from(u128::from(GENESIS_SRI).saturating_mul(pool_value) / total_value).unwrap()
```
kayabaNerve: With reference to my prior comment, if sum of pool values exceeded total value this will cause minting more SRI than defined in GENESIS_SRI (until L130 triggers a panic by the underflow, assuming range checks are on/we instead use checked_sub there).

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+          };
+          total_sri_distributed += sri_amount;
+
+          // we can't add 0 liquidity
```
kayabaNerve: Isn't there actually a minimum in the DEX pallet for the initial liq? 10 or 50k atomic units?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+        // All transferred to the pools.
+        for coin in COINS {
+          assert_eq!(Coins::<T>::balance(GENESIS_LIQUIDITY_ACCOUNT.into(), coin), Amount(0));
+        }
+
+        GenesisComplete::<T>::set(Some(true));
+      }
+
+      // we accept we reached economic security once we can mint smallest amount of a network's coin
+      for coin in COINS {
```
kayabaNerve, line 174: This part may make sense to have at end of block, yet also shouldn't be in this pallet to be honest. It should be derived from the pre-econ/post-econ pallet. We can mark this TODO for now to replace in a later PR?

substrate/genesis-liquidity/pallet/src/lib.rs
kayabaNerve: The argument for it is the coins were flagged as incoming with the II, and the II is just executing upon the incoming coins.

The risk is we mint in II _and_ mint in some module.

I do think II should be the minter. If it mints the coins, and then we fail to execute the II (add genesis liq), it'll get scheduled to get sent back out and not be an issue.

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+        Err(Error::<T>::GenesisPeriodEnded)?;
+      }
+
+      // mint the coins
+      Coins::<T>::mint(GENESIS_LIQUIDITY_ACCOUNT.into(), balance)?;
+
+      // calculate new shares & supply
+      let (new_shares, new_supply) = if let Some(supply) = Supply::<T>::get(balance.coin) {
+        // calculate amount of shares for this amount
+        let shares = Self::mul_div(supply.0, balance.amount.0, supply.1)?;
```
kayabaNerve: Let's move this to a struct so we're not at risk of conflating 0/1? Premise LGTM though :)

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,444 @@
...
+  #[pallet::validate_unsigned]
+  impl<T: Config> ValidateUnsigned for Pallet<T> {
+    type Call = Call<T>;
+
+    fn validate_unsigned(_: TransactionSource, call: &Self::Call) -> TransactionValidity {
+      match call {
+        Call::set_initial_price { ref prices, ref signature } => {
+          // TODO: if this is supposed to be called after a month, serai set won't still be
+          // in the session 0? Ideally this should pull the session from Vs pallet?
+          let set = ValidatorSet { network: NetworkId::Serai, session: Session(0) };
```
kayabaNerve: We definitely should be able to read the latest session. We won't have distinct validators since there's no staking/unstaking.

substrate/abi/src/genesis_liquidity.rs
akildemir: separate pr after this is merged sounds good 👍

substrate/genesis-liquidity/pallet/src/lib.rs
akildemir: yea  but it is for liquidity tokens that the pallet handles itself. Only 0 check is being done on the actual coins being added.

substrate/abi/src/genesis_liquidity.rs
akildemir: method doc already says that value is in terms of btc 👍

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+
+use sp_std::vec::Vec;
+
+use scale::{Encode, Decode, MaxEncodedLen};
+use scale_info::TypeInfo;
+
+use serai_primitives::*;
+use validator_sets_primitives::ValidatorSet;
+
+pub const GENESIS_LP_SHARES: u64 = 10_000;
```
kayabaNerve: This seems inaccurately described.

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+// This is the account to hold and manage the genesis liquidity.
+pub const GENESIS_LIQUIDITY_ACCOUNT: SeraiAddress = system_address(b"Genesis-liquidity-account");
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Zeroize))]
+#[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Prices {
+  pub monero: u64,
+  pub ethereum: u64,
```
kayabaNerve: ether != ethereum

substrate/genesis-liquidity/primitives/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+pub const GENESIS_LP_SHARES: u64 = 10_000;
+
+// This is the account to hold and manage the genesis liquidity.
+pub const GENESIS_LIQUIDITY_ACCOUNT: SeraiAddress = system_address(b"Genesis-liquidity-account");
+
+#[derive(Clone, Copy, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[cfg_attr(feature = "std", derive(Zeroize))]
+#[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]
+#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+pub struct Prices {
```
kayabaNerve: Values

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+async fn test_genesis_liquidity(serai: Serai) {
+  // amounts
+  let amounts = vec![
+    Amount(5_53246991),
+    Amount(3_14562819),
+    Amount(9_33648912),
+    Amount(150_873639000000),
+    Amount(248_665228000000),
+    Amount(451_765529000000),
+  ];
```
kayabaNerve: Should these be hard-coded or RNG with re-derived derivatives?

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+  OsRng.fill_bytes(&mut block_hash.0);
+  let xmr_ins = xmr_addresses
+    .iter()
+    .map(|(addr, amount)| InInstructionWithBalance {
+      instruction: InInstruction::GenesisLiquidity(*addr),
+      balance: Balance { coin: Coin::Monero, amount: *amount },
+    })
+    .collect::<Vec<_>>();
+  let batch = Batch { network: NetworkId::Monero, id: 0, block: block_hash, instructions: xmr_ins };
+  provide_batch(&serai, batch).await;
```
kayabaNerve: We can iterate over `[Coin::Bitcoin, Coin::Monero]` here. We should also be to test with Ether/Dai?

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+
+  // wait until genesis ends..
+  tokio::time::timeout(tokio::time::Duration::from_secs(300), async {
```
kayabaNerve: We should import a const for this value.

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+    while serai.latest_finalized_block().await.unwrap().number() < 10 {
+      tokio::time::sleep(Duration::from_secs(6)).await;
+    }
+  })
+  .await
+  .unwrap();
+
+  // set prices
+  let prices = Prices { monero: 184100, ethereum: 4785000, dai: 1500 };
```
kayabaNerve: values

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+  set_prices(&serai, &prices).await;
```
kayabaNerve: set_values

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+  // check each xmr liq provider got liq tokens proportional to their value
+  let xmr_liq_supply = serai.genesis_liquidity().supply(Coin::Monero).await.unwrap();
+  for (acc, amount) in xmr_addresses {
+    let acc_liq_shares =
+      serai.genesis_liquidity().liquidity(&acc, Coin::Monero).await.unwrap().0 .0;
+
+    let shares_ratio = (GENESIS_LP_SHARES * acc_liq_shares) / xmr_liq_supply.0 .0;
+    let amounts_ratio = (GENESIS_LP_SHARES * amount.0) / u64::try_from(pool_xmr).unwrap();
+    assert_eq!(shares_ratio, amounts_ratio);
+  }
```
kayabaNerve, line 180: Same commentary on being able to use an iter over an array here.

substrate/client/tests/genesis_liquidity.rs
```diff
@@ -0,0 +1,206 @@
...
+    &mut OsRng,
+    frost::tests::algorithm_machines(
+      &mut OsRng,
+      &Schnorrkel::new(b"substrate"),
+      &HashMap::from([(threshold_keys.params().i(), threshold_keys.into())]),
+    ),
+    &oraclize_values_message(&set, prices),
+  );
+
+  // set initial prices
```
kayabaNerve: oraclize values

substrate/genesis-liquidity/pallet/src/lib.rs
kayabaNerve: Can we at least add a TODO to revisit this then please? It's not fundamentally unacceptable yet we need to be working towards robust solutions.

substrate/genesis-liquidity/pallet/src/lib.rs
kayabaNerve: https://github.com/serai-dex/serai/blob/develop/substrate/dex/pallet/src/lib.rs#L608-L623

https://github.com/serai-dex/serai/blob/develop/substrate/runtime/src/lib.rs#L231

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -0,0 +1,471 @@
...
+        }
+        assert_eq!(total_sri_distributed, GENESIS_SRI);
+
+        // we shouldn't have left any coin in genesis account at this moment, including SRI.
+        // All transferred to the pools.
+        for coin in COINS {
+          assert_eq!(Coins::<T>::balance(GENESIS_LIQUIDITY_ACCOUNT.into(), coin), Amount(0));
+        }
+
+        GenesisComplete::<T>::set(Some(true));
```
kayabaNerve: We can have the value be `()` instead of a `bool`, where if a value was set is if it's complete.

kayabaNerve:
Processor CI hung, unrelated.

PR #551, opened 2024-04-09
Implement block emissions

akildemir:


substrate/abi/src/lib.rs
```diff
@@ -32,8 +33,9 @@ pub enum Call {
   Coins(coins::Call),
   LiquidityTokens(liquidity_tokens::Call),
   Dex(dex::Call),
-  GenesisLiquidity(genesis_liquidity::Call),
   ValidatorSets(validator_sets::Call),
+  GenesisLiquidity(genesis_liquidity::Call),
+  Emissions(emissions::Call),
```
kayabaNerve: If the call/empty is empty, why not make the inner-call here `()`? That wouldn't require empty (invalid) definitions.

substrate/client/src/serai/dex.rs
```diff
@@ -61,18 +61,11 @@ impl<'a> SeraiDex<'a> {
...
-      .call(
-        "state_call",
-        ["DexApi_get_reserves".to_string(), hex::encode((coin, Coin::Serai).encode())],
-      )
-      .await?;
-    let bytes = Serai::hex_decode(reserves)?;
-    let result = decode_from_bytes::<Option<(u64, u64)>>(bytes.into())
-      .map_err(|e| SeraiError::ErrorInResponse(e.to_string()))?;
-    Ok(result.map(|amounts| (Amount(amounts.0), Amount(amounts.1))))
+  pub async fn get_reserves(&self, coin: Coin) -> Result<Option<(u64, u64)>, SeraiError> {
```
kayabaNerve: Should this return u64 or Amount? They share an encoding so we can do either.

substrate/client/src/serai/genesis_liquidity.rs
```diff
@@ -62,4 +62,8 @@ impl<'a> SeraiGenesisLiquidity<'a> {
   pub async fn supply(&self, coin: Coin) -> Result<LiquidityAmount, SeraiError> {
     Ok(self.0.storage(PALLET, "Supply", coin).await?.unwrap_or(LiquidityAmount::zero()))
   }
+
+  pub async fn genesis_complete(&self) -> Result<Option<()>, SeraiError> {
+    self.0.storage(PALLET, "GenesisComplete", ()).await
```
kayabaNerve: `.map(Option::is_some)` would enable us to return a bool here.

substrate/client/tests/common/genesis_liquidity.rs
```diff
@@ -0,0 +1,210 @@
...
+    Amount, NetworkId, Coin, Balance, BlockHash, SeraiAddress, insecure_pair_from_name, GENESIS_SRI,
+  },
+  in_instructions::primitives::{InInstruction, InInstructionWithBalance, Batch},
+  Serai,
+};
+
+use crate::common::{in_instructions::provide_batch, tx::publish_tx};
+
+#[allow(dead_code)]
+pub async fn test_genesis_liquidity(serai: Serai) -> HashMap<NetworkId, u32> {
```
kayabaNerve: I really don't like the idea of this `test` being common code. The pieces of it desirable in other tests should be broken out into common fns, yet this should be left as a test.

substrate/client/tests/validator_sets.rs
```diff
@@ -326,8 +326,8 @@ async fn verify_session_and_active_validators(
   session: u32,
   participants: &[Public],
 ) {
-  // wait until the active session. This wait should be max 30 secs since the epoch time.
-  let block = tokio::time::timeout(core::time::Duration::from_secs(2 * 60), async move {
+  // wait until the active session. This wait should be max 2 mins since the epoch time.
+  let block = tokio::time::timeout(core::time::Duration::from_secs(5 * 60), async move {
```
kayabaNerve: *something something if we read this constant from primitives, we wouldn't have to manually update tests*

substrate/client/src/serai/mod.rs
```diff
@@ -378,6 +367,28 @@ impl<'a> TemporalSerai<'a> {
...
+      .call(
+        "state_call",
+        [method.to_string(), hex::encode(params.encode()), hex::encode(self.block)],
+      )
+      .await?;
+
+    let bytes = Serai::hex_decode(result.clone())?;
+    R::decode(&mut bytes.as_slice()).map_err(|_| {
+      SeraiError::InvalidRuntime(format!(
+        "different type than what is expected returned, raw value: {}",
```
kayabaNerve: "expected to be returned"

substrate/runtime/src/abi.rs
```diff
@@ -138,6 +127,18 @@ impl From<Call> for RuntimeCall {
...
+          RuntimeCall::GenesisLiquidity(genesis_liquidity::Call::remove_coin_liquidity { balance })
+        }
+        serai_abi::genesis_liquidity::Call::oraclize_values { values, signature } => {
+          RuntimeCall::GenesisLiquidity(genesis_liquidity::Call::oraclize_values {
+            values,
+            signature,
+          })
+        }
+      },
+      Call::Emissions(_) => todo!(), // TODO
```
kayabaNerve: Hm. We do need a solution for this...

substrate/primitives/src/networks.rs
```diff
@@ -15,7 +15,9 @@ use sp_core::{ConstU32, bounded::BoundedVec};
 use crate::{borsh_serialize_bounded_vec, borsh_deserialize_bounded_vec};
 
 /// The type used to identify networks.
-#[derive(Clone, Copy, PartialEq, Eq, Hash, Debug, Encode, Decode, MaxEncodedLen, TypeInfo)]
+#[derive(
+  Clone, Copy, PartialEq, Eq, Hash, Debug, Encode, Decode, PartialOrd, Ord, MaxEncodedLen, TypeInfo,
```
kayabaNerve, line 19: Why are we adding PartialOrd/Ord?

substrate/genesis-liquidity/pallet/src/lib.rs
```diff
@@ -72,6 +72,7 @@ pub mod pallet {
   pub(crate) type Oracle<T: Config> = StorageMap<_, Identity, Coin, u64, OptionQuery>;
 
   #[pallet::storage]
+  #[pallet::getter(fn genesis_complete)]
```
kayabaNerve, line 75: FYI, `getter` was deprecated in modern versions of Substrate and I'd agree it's not great. I do get it's used here to make this value accessible from other sections of the codebase though.

substrate/primitives/src/networks.rs
kayabaNerve, line 19: Why are we adding PartialOrd/Ord?

substrate/emissions/pallet/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
+
+[package.metadata.cargo-machete]
+ignored = ["scale", "scale-info"]
+
+[lints]
+workspace = true
+
+
```
kayabaNerve: Random new line sneaked itself in

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+      // TODO: may panic? It might be best for this math ops to return the result as is instead of
+      // doing an unwrap so that it can be properly dealt with.
+      let sri_amount = balance.amount.mul(value);
+
+      // Mint & stake the SRI for the network.
+      Coins::<T>::mint(to, Balance { coin: Coin::Serai, amount: sri_amount })?;
+      // TODO: deposit_stake lets staking less than per key share. Should we allow that here?
+      ValidatorSets::<T>::deposit_stake(network, to, sri_amount)?;
+      Ok(())
+    }
```
kayabaNerve, line 444: Still needs to form the PoL.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+      // calculate how much SRI the balance makes
+      let value =
+        Dex::<T>::security_oracle_value(balance.coin).ok_or(Error::<T>::NoValueForCoin)?;
+      // TODO: may panic? It might be best for this math ops to return the result as is instead of
+      // doing an unwrap so that it can be properly dealt with.
+      let sri_amount = balance.amount.mul(value);
+
+      // Mint & stake the SRI for the network.
+      Coins::<T>::mint(to, Balance { coin: Coin::Serai, amount: sri_amount })?;
+      // TODO: deposit_stake lets staking less than per key share. Should we allow that here?
```
kayabaNerve: Good catch! No, this should require the would-be stake after the value exceeds 1 node (so partial stakes only if already staked to 1 validator).

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+      balance: Balance,
+    ) -> DispatchResult {
+      // check the network didn't reach the economic security yet
+      if Self::economic_security_reached(network) {
+        Err(Error::<T>::NetworkHasEconomicSecurity)?;
+      }
+
+      // calculate how much SRI the balance makes
+      let value =
+        Dex::<T>::security_oracle_value(balance.coin).ok_or(Error::<T>::NoValueForCoin)?;
```
kayabaNerve: Uhhh we don't want to use the security oracle value, which is the highest-observed median price. We want to use the median price OR the spot price. Neither is *great* as the median price has a price delay and sounds close to what happened with Haven (which you'd have more knowledge/experience on than myself). We could use a second, smaller median?

deny.toml
```diff
@@ -52,6 +52,9 @@ exceptions = [
 
   { allow = ["AGPL-3.0"], name = "serai-coins-pallet" },
   { allow = ["AGPL-3.0"], name = "serai-dex-pallet" },
+  
+  { allow = ["AGPL-3.0"], name = "serai-genesis-liquidity-pallet" },
+  { allow = ["AGPL-3.0"], name = "serai-emissions-pallet" },
 
   { allow = ["AGPL-3.0"], name = "serai-genesis-liquidity-pallet" },
```
kayabaNerve, line 56: genersis-liquidity slipped in twice.

substrate/dex/pallet/src/lib.rs
```diff
@@ -194,6 +194,11 @@ pub mod pallet {
   #[pallet::getter(fn security_oracle_value)]
   pub type SecurityOracleValue<T: Config> = StorageMap<_, Identity, Coin, Amount, OptionQuery>;
 
+  /// Current length of the `SpotPrices` map.
+  #[pallet::storage]
+  #[pallet::getter(fn swap_volume)]
+  pub type SwapVolume<T: Config> = StorageMap<_, Identity, Coin, u64, OptionQuery>;
```
kayabaNerve: The name of this map doesn't seem to align with the documentation at all.

substrate/dex/pallet/src/lib.rs
```diff
@@ -561,11 +537,14 @@ pub mod pallet {
...
 
-      let pool_id = Self::get_pool_id(coin, Coin::Serai).unwrap();
+      let pool_id = Self::get_pool_id(coin, Coin::Serai)?;
 
-      Pools::<T>::get(pool_id).as_ref().ok_or(Error::<T>::PoolNotFound)?;
+      // create the pool if it doesn't exist. We can just attempt to do that because our checks
+      // far enough to allow that.
+      if Pools::<T>::get(pool_id).is_none() {
+        Self::create_pool(coin)?;
+      }
```
kayabaNerve, line 547: This implicitly assumes the system is the first to call add_liquidity, which means we absolutely never mint 1 unlocked atomic unit of SRI to anyone, anywhere. That should be the case, yet we can also make this more robust if we also check genesis is complete. That'd be a bit tricky to implement though (as dex would rely on genesis liquidity would rely on dex).

substrate/dex/pallet/src/lib.rs
```diff
@@ -890,6 +869,20 @@ pub mod pallet {
           }
           i += 1;
         }
+
+        // update the volume, SRI amount is always at index 1 if the path len is 2 or 3.
+        // path len 1 is not allowed and 3 is already the maximum.
```
kayabaNerve: For a SRI to BTC swap, why is SRI at index 1? Some sort policy?

kayabaNerve: For a SRI to BTC swap, why is SRI at index 1? Some sort policy?

substrate/dex/pallet/src/lib.rs
```diff
@@ -890,6 +869,20 @@ pub mod pallet {
...
+        let swap_volume = amounts.get(1).ok_or(Error::<T>::CorrespondenceError)?;
+        let existing = SwapVolume::<T>::get(coin1).unwrap_or(0);
+        let new_volume = existing.saturating_add(*swap_volume);
+        SwapVolume::<T>::set(coin1, Some(new_volume));
+
+        // if we did 2 pools, update the volume for second coin as well.
```
kayabaNerve: This isn't the same value though. One exited the pool after fees, one entered into the pool *before that pool took fees*.

substrate/emissions/pallet/Cargo.toml
```diff
@@ -0,0 +1,61 @@
...
+  "coins-pallet/std",
+  "validator-sets-pallet/std",
+  "dex-pallet/std",
+  "genesis-liquidity-pallet/std",
+
+  "serai-primitives/std",
+  "emissions-primitives/std",
+]
+try-runtime = [] # TODO
+default = ["std"]
```
kayabaNerve: Missing newline at end of file.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -819,6 +832,22 @@ pub mod pallet {
       total_required
     }
 
+    // TODO: make the increase_allocation public instead?
```
kayabaNerve: I think you can call the existing allocate function so long as you craft a signed runtime origin containing the account ID (it should just be a struct, this doesn't imply forging some signature).

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -490,11 +502,12 @@ pub mod pallet {
...
       account: T::AccountId,
       amount: Amount,
+      block_reward: bool,
     ) -> DispatchResult {
       let old_allocation = Self::allocation((network, account)).unwrap_or(Amount(0)).0;
       let new_allocation = old_allocation + amount.0;
       let allocation_per_key_share = Self::allocation_per_key_share(network).unwrap().0;
-      if new_allocation < allocation_per_key_share {
+      if new_allocation < allocation_per_key_share && !block_reward {
         Err(Error::<T>::InsufficientAllocation)?;
```
kayabaNerve, line 527: This may break an invariant (present in map yet with less allocated than necessary for a key share)?

Why would they be getting a block reward if they don't have an allocation? Because they deallocated? If so, I think adding to their pending deallocation likely makes the most sense, if it isn't too complex? Or we can re-check our usages of this map and ensure it's documented and fine that accounts in it may not have enough for a key share.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -309,6 +310,12 @@ pub mod pallet {
   #[pallet::storage]
   pub type SeraiDisabledIndices<T: Config> = StorageMap<_, Identity, u32, Public, OptionQuery>;
 
+  /// Mapping from session to block number
+  #[pallet::storage]
+  #[pallet::getter(fn session_begin_block)]
+  pub type SessionBeginBlock<T: Config> =
```
kayabaNerve, line 331: Double checking there's no existing fn for this in BABE/GRANDPA.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+    fn initial_period(n: BlockNumberFor<T>) -> bool {
+      // TODO: we should wait for exactly 2 months according to paper. This waits double the time
+      // it took until genesis complete since we assume it will be done in a month. We know genesis
+      // period blocks is a month but there will be delay until oracilization is done and genesis
+      // completed and emissions start happening. If we wait exactly 2 months and the delay is big
+      // enough we might not be able to distribute all funds we want to in this period.
+      // In the current case we will distribute more than we want to. What to do?
+      let genesis_complete_block = GenesisCompleteBlock::<T>::get();
+      genesis_complete_block.is_some() &&
+        (n.saturated_into::<u64>() < (3 * genesis_complete_block.unwrap()))
```
kayabaNerve: Why not calculate this as `genesis_complete_block + (2 * BLOCKS_PER_MONTH)`?

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+
+    fn distribute_to_validators(n: NetworkId, reward: u64) -> DispatchResult {
+      // distribute among network's set based on
+      // -> (key shares * stake per share) + ((stake % stake per share) / 2)
+      let stake_per_share = ValidatorSets::<T>::allocation_per_key_share(n).unwrap().0;
+      let mut scores = vec![];
+      let mut total_score = 0u64;
+      for (p, amount) in Self::participants(n).unwrap() {
+        let remainder = amount % stake_per_share;
+        let score = (amount - remainder) + (remainder / 2);
```
kayabaNerve: `amount - (remainder / 2)` would also work if we find that clearer.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+
+      // stake the rewards
+      for (p, score) in scores {
+        let p_reward = u64::try_from(
+          u128::from(reward).saturating_mul(u128::from(score)) / u128::from(total_score),
+        )
+        .unwrap();
+
+        Coins::<T>::mint(p, Balance { coin: Coin::Serai, amount: Amount(p_reward) })?;
+        ValidatorSets::<T>::deposit_stake(n, p, Amount(p_reward))?;
```
kayabaNerve: If the mint passes, yet the deposit_stake fails (which we may want? I left a review comment on having less than the key share allocated, which the current impl does so this doesn't fail), this will leave the reward directly in the account. We can ensure if the deposit stake fails, that doesn't happen, by having this section create/use a new DB layer (so if any part of it errors, all storage changes within it error).

I don't want to call for this change yet, I just want to note the option/consideration.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+        GenesisCompleteBlock::<T>::set(Some(n.saturated_into::<u64>()));
+      }
+
+      // we wait 1 extra block after genesis ended to see the changes. We only need this extra
+      // block in dev&test networks where we start the chain with accounts that already has some
+      // staked SRI. So when we check for ec-security pre-genesis we look like we are economically
+      // secure. The reason for this although we only check for it once the genesis is complete(so
+      // if the genesis complete we shouldn't be economically secure because the funds are not
+      // enough) is because ValidatorSets pallet runs before the genesis pallet in runtime.
+      //  So ValidatorSets pallet sees the old state until next block.
```
kayabaNerve, line 122: Can we simply have the ValidatorsSets pallet run after? What updates does the ValidatorSets pallet do that's relevant here? Since we have to program in genesis validators not actually having stake at some point, so we should not there here and schedule this to be removed upon that `TODO` being completed?

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+      let session = ValidatorSets::<T>::session(NetworkId::Serai).unwrap_or(Session(0));
+      if session.0 > Self::session(NetworkId::Serai) {
+        session_changed = true;
+        CurrentSession::<T>::set(NetworkId::Serai, session.0);
+      }
+
+      // update participants per session before the genesis and after the genesis
+      // we update them after reward distribution.
+      if !genesis_ended && session_changed {
+        Self::update_participants();
```
kayabaNerve, line 147: Why can't we read the participants from validator-sets?

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+            let reward = u64::try_from(
+              u128::from(reward_this_epoch).saturating_mul(u128::from(distance)) /
+                u128::from(total_distance),
+            )
+            .unwrap();
+            (n, reward)
+          })
+          .collect::<BTreeMap<NetworkId, u64>>()
+      } else {
+        volume_per_network
```
kayabaNerve: If we only use volume_per_network post-EC (as sane), we should only calculate it in this branch.

One `if` with two large branches, not two `if`s.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+        // distribute validators rewards
+        if Self::distribute_to_validators(n, validators_reward).is_err() {
+          // TODO: log the failure
+          continue;
+        }
+
+        // send the rest to the pool
+        let coin_count = u64::try_from(n.coins().len()).unwrap();
+        for c in n.coins() {
+          // assumes reward is equally distributed between network coins.
```
kayabaNerve: We should calculate per-pool, and simply decide the validator split off the sum for the network.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -0,0 +1,413 @@
...
+          if Coins::<T>::mint(
+            Dex::<T>::get_pool_account(*c),
+            Balance { coin: Coin::Serai, amount: Amount(pool_reward / coin_count) },
+          )
+          .is_err()
+          {
+            // TODO: log the failure
```
kayabaNerve: Can minting SRI fail? I think we can panic here without too much concern (though sure, a panic at the start of the block isn't great).

substrate/emissions/pallet/src/lib.rs
kayabaNerve: Why not calculate this as `genesis_complete_block + (2 * BLOCKS_PER_MONTH)`?

substrate/primitives/src/networks.rs
akildemir: Because we use `Network` in a `BTreeMap` in the `emissions` pallet which requires these 2 trait to be derived as well.

substrate/genesis-liquidity/pallet/src/lib.rs
akildemir: yup, we can't directly access to storage map iirc atm. Would like to change it if it is possible to  access to the storage maps in this version like however they are doing in those modern versions.

substrate/emissions/pallet/src/lib.rs
akildemir: yea possible. Also just spot price is also seems fine?

substrate/dex/pallet/src/lib.rs
akildemir: We do call `add_liquidity` before we set genesis is complete. We do call it in process of completing the genesis. So that check would make that call not possible. No one already can't call `add_liquidity` since there is no SRI until genesis is complete.

substrate/emissions/pallet/Cargo.toml
akildemir: why didn't fmt give me an error for this?

substrate/validator-sets/pallet/src/lib.rs
akildemir: won't genesis validators have any existing allocations when they start to receive block rewards? Deallocations of block rewards will go through regular deallocation process except for the last session? Last session reward will immediately be deallocatable since the validator won't be in the set.

substrate/validator-sets/pallet/src/lib.rs
akildemir: BABE/GRANDPA only care about the epochs and not for sessions. So yes, they don't have such fn.

substrate/emissions/pallet/src/lib.rs
akildemir: Because then we won't be able to reach after the initial period for tests. Not that our current test cover that case but hey it is a TODO! 🤣

substrate/emissions/pallet/src/lib.rs
akildemir: there is `is_allowed` check being done on the SRI as well, so either that or overflow. I just think any panic in the hooks will stop the network since everyone will panic? or substrate just retries a new block?

substrate/emissions/pallet/src/lib.rs
akildemir: Well GenesisLiquidity pallet depends on ValidatorSets so I don't think making it run after doesn't makes sense. As the comment states, If we check ec-security in the same block as genesis-complete, `ValidatorSets::is_allowed` returns true tho it shouldn't since we just minted the coins. But emissions pallet sees it. `ValidatorsSets` runs before `GenesisLiquidity` and `Emissions` runs after. This discrepancy was actually I spent the most time on this pr lol. But it should be ok as is? Didn't understand your last sentence tho.

substrate/emissions/pallet/src/lib.rs
akildemir: We can and we already do within `update_participants`.  It is just that we keep the participants of the last session in this pallet since rewards are distributed to them in the current session. And we update them per session.

substrate/genesis-liquidity/pallet/src/lib.rs
kayabaNerve: Or manually add the fn, but yeah, we can resolve it on upgrading.

substrate/emissions/pallet/src/lib.rs
kayabaNerve: Eh. It lets someone spike the price, then mint SRI at the inflated value. We should at least use prior block's end of block price to be close to spot yet require multiple sequential validators to successfully perform such attacks. That's still not a *great* trust assumption.

substrate/emissions/pallet/Cargo.toml
kayabaNerve: It's not required, it's just a pleasantry.

Also, this is a toml and fmt doesn't touch toml.

substrate/validator-sets/pallet/src/lib.rs
kayabaNerve: So we do need block rewards -> allocations, and we just need to double check we don't assume present in allocations map == eligible. ACK.

substrate/emissions/pallet/src/lib.rs
kayabaNerve: I'm unsure. I'd assume it fail the block, and move to the next block, which would stall the network *if all blocks at that position have some error*. This is something we should check.

Heard on overflow though. I still don't like having this error condition. This should be an invariant and any attempted handling of this invariant will always be incomplete to some degree. Solely logging there's an invariant doesn't change the presence of the invariant.

substrate/emissions/pallet/src/lib.rs
kayabaNerve: I think validator-sets, in general, needs to have its database handling improved. There's a variety of specific behavior it's optimized around, yet as we've evolved it and extended it, it's drifted away from that behavior and lead to... wherever we are now.

validator-sets, upon its database handling cleanup/redo, should be able to service this functionality. emissions should not *need* to do this. If we need to accept that for now, fine, but we should consider and mark that.

substrate/emissions/pallet/src/lib.rs
kayabaNerve: In that case, we should make a dedicated constant for this, and on non-test `2 * BLOCKS_PER_MONTH`, and on test `x`.

substrate/dex/pallet/src/lib.rs
kayabaNerve: > since there is no SRI until genesis is complete.

That should be a safe invariant to rely on. I was just considering the damage possible if even a single SRI became free-floating somehow.

substrate/emissions/pallet/src/lib.rs
kayabaNerve: I'm fine with the re-ordering. I just am trying work towards minimize these weird edge cases which require paragraphs explaining them. Some will always be justified.

> Since we have to program in genesis validators not actually having stake at some point

The genesis validators for mainnet will not have stake. They will be considered to have stake for the purpose of validator selection. Their only stake will be the rewards they earn.

> so we should not there here and schedule this to be removed upon that TODO being completed?

> > so we should not there here

*so we should note this here

> > and schedule this to be removed upon that TODO being completed

Since eventually, genesis validators will not actually have stake per EconSec considerations, the necessity for this workaround should fall away.

substrate/emissions/pallet/src/lib.rs
akildemir: Yea but the idea is that this is a rewards distribution code and we know it shouldn't fail. If it still fails for a node, just that node won't get the rewards for that session and that is all there is. We can see it in the logs and decide whether it is the result of a bug or just something normal and if it turns out to be a bug we can fix it. Which is better than panicking IMO. Because if we panic for the first node in the array we won't send any rewards to any node, if we log the error and continue we at least have chance at rewarding other nodes.  If you check the latest version of the code that I pushed yesterday this should be more obvious.

Issue #549, opened 2024-04-02
Directly fuzz monero-serai against the C++ codebase


kayabaNerve:

kayabaNerve:
Manually migrated with https://github.com/monero-oxide/monero-oxide/issues/4.

PR #552, opened 2024-04-09
fix CLSAG verification.

Boog900:
We were not setting c1 to the last calculated c during verification, instead keeping it set to the one provided in the signature, allowing forgeries.

@kayabaNerve I did message you on matrix but it seems there are still issues so I decided just to open this PR, although this is bad AFAIK no one relies on this ATM.

### DIFF
--- a/coins/monero/src/ringct/clsag/mod.rs
+++ b/coins/monero/src/ringct/clsag/mod.rs
@@ -9,7 +9,7 @@ use std_shims::{
 use rand_core::{RngCore, CryptoRng};
 
 use zeroize::{Zeroize, ZeroizeOnDrop, Zeroizing};
-use subtle::{ConstantTimeEq, Choice, CtOption};
+use subtle::{ConstantTimeEq, ConditionallySelectable};
 
 use curve25519_dalek::{
   constants::ED25519_BASEPOINT_TABLE,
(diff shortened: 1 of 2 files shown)


kayabaNerve:
No. This is a critical in not only monero-serai yet in the multisig functionality specifically. This would cause invalid participants in the FROST process to not be flagged as it'd think the signature is still (wrongly) valid.

On the one hand, monero-serai wasn't audited (queued), I haven't done my final pass, and it's not a 1.0. On the other, I'm fucking pissed this made it through and horrified. At best, I can be proud monero-serai did enough right it caught enough other people this was itself caught.

Matrix is acting up and I don't have any messages from you. Please reach out on Discord (Serai Discord to find me) or Telegram (kayabaNerve), or let me know if I need to make a new Matrix account on monero.social so we can sync on this.

kayabaNerve:
All versions of monero-serai are currently yanked for an independent reason (several months ago). I never re-published after (I don't even think I've published the audited bitcoin-serai), so no yanking needed due to this. Still outrageous.

kayabaNerve:
Expansion of basic tests, as even those would've caught this (further emphasizing the fault here), fuzz tests, and prioritizing https://github.com/serai-dex/serai/issues/549? That's my immediate thoughts on the matter. Possibly addition of code coverage evaluators? That'd highlight such missing cases...

kayabaNerve:
Also, to be perfectly clear, only myself is to blame here. All my frustration is self-directed.

Boog900:
I don't use any of those other platforms. If the Matrix issues are causing problems in other chats, it might be best to make a monero.social account otherwise I could make a matrix.org account.

Boog900:
+1 for more tests

Boog900:
I ended up making a matrix.org account: boog900:matrix.org

PR #561, opened 2024-04-21
#560 take two, now that #560 has been reverted

kayabaNerve:
https://github.com/serai-dex/serai/commit/933b17aa911d1b27c8c76d5c4715b329d6cbfa5a for the comments on why this was reverted.

There's probably some typo/massive perf issue here needing resolving? Rebroadcasting frequency?

### DIFF
--- a/common/db/src/parity_db.rs
+++ b/common/db/src/parity_db.rs
@@ -4,6 +4,7 @@ pub use ::parity_db::{Options, Db as ParityDb};
 
 use crate::*;
 
+#[must_use]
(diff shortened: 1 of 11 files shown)


kayabaNerve:
The upons fix is critical for consensus, the round proposal part is an efficiency improvement (as off-hand, I'd assume this lib is slower than the prior lib, even if it's clearer and capable of being audited).

kayabaNerve:
:partying_face: 

@akildemir for the credit on this crossing the finish line.

PR #568, opened 2024-05-10
monero: documentation

hinto-janai:
Adds more crate documentation to `monero-serai`.

This PR is forked from and is being merged into `monero-tidy`.

### DIFF
--- a/coins/monero/README.md
+++ b/coins/monero/README.md
@@ -47,3 +47,15 @@ It also won't act as a wallet, just as a transaction library. wallet2 has
 several *non-transaction-level* policies, such as always attempting to use two
 inputs to create transactions. These are considered out of scope to
 monero-serai.
+
+### Feature flags
+monero-serai has certain functionality behind feature flags:
+
+- `std:` Enables usage of Rust's `std` and several other functionality. See `Cargo.toml` for the full list.
+- `http-rpc`: Enables an HTTP(S) transport type within the `rpc` module
+- `multisig`: Enables multi-signature features within the `wallet` module
+- `binaries`: TODO
+
+The features enabled by default are:
+- `std`
+- `http-rpc`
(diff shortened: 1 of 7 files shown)


hinto-janai:
@kayabaNerve Would it make more sense to open this after the crate/module move + FCMP changes to avoid conflicts?

kayabaNerve:
FCMPs aren't being done with this codebase. I personally won't merge any FCMP code into the Serai repo until Monero confirms it will be the next hard fork.

The one question would be if we smash the crates. If we smash the crates, I'd rather do so from this branch.

Feel free to comment your opinions on having monero-generators, monero-mlsag, monero-clsag, monero-bulletproofs, etc.

hinto-janai:
IMO meta-crates have slightly less user-friendly `cargo doc`s and usage compared to modules, other than that I think it's mostly the same to users either way.

- Meta-crate example: https://docs.rs/jsonrpsee
- Module example: https://docs.rs/tokio

From an auditing and maintenance perspective, meta-crates are probably easier to work with.

kayabaNerve:
I'd rather cite bitcoin or alloy as the meta-crate examples.

I'd also note re: tokio, Serai uses a variety of crates from the smol ecosystem (which is an modular attempt at an async ecosystem) because of wanting specific functionality without pulling in all of/binding to tokio.

hinto-janai:
Ah I see, I didn't know `#[doc(inline)]` existed (wonder why these [don't](https://docs.rs/symphonia) [use](https://docs.rs/jsonrpsee) [it](https://microsoft.github.io/windows-docs-rs)). As far as user-friendliness goes then, I guess it's mostly the same.

hinto-janai:
Closing for now, thanks for adding docs.

kayabaNerve:
FYI, I do still plan to reconcile this PR. Sorry for stepping over you. I decided to smash the crates per a few brief conversations, and that mandated clear API boundaries, and with it, I decided to get the documentation out of the way.

I should've reconciled this first and forgot to do so. I apologize there.

PR #577, opened 2024-07-04
Clean the Monero lib for auditing

kayabaNerve:
This still needs:

1) More tests in general
2) More documentation improvements from #568
3) Potentially uplifting the FCMP++ BP to replace the BP impl present here
4) Resolution of TODOs

Yet it is time to start running the CI again.


kayabaNerve:
I'm willing to handle the rest in commits.

PR #581, opened 2024-07-15
Fix tendermint chain sync

akildemir:


coordinator/src/p2p.rs
```diff
@@ -48,6 +48,9 @@ use crate::{Transaction, Block, Tributary, ActiveTributary, TributaryEvent};
 const MAX_LIBP2P_MESSAGE_SIZE: usize = tributary::BLOCK_SIZE_LIMIT + 1024;
```
kayabaNerve: This now needs to be `(BLOCKS_PER_BATCH * BLOCK_SIZE_LIMIT) + 1024`. We should probably different GOSSIP_MESSAGE_SIZE from REQRES_MESSAGE_SIZE (the former remaining the same, the latter being this new definition).

Also, BLOCKS_PER_BATCH constant.

kayabaNerve (commented):
LGTM with clippy fix and corrected constants.

Issue #576, opened 2024-07-04
1-round DKG without re-attempts


kayabaNerve:
We currently have a 2-round DKG, where we perform re-attempts as necessary.

https://eprint.iacr.org/2024/397 proposes a 1-round DKG. Since we don't vary our threshold upon re-attempt, we can do the 1-round DKG where the first `t` entities to participate decide the key. Then, we get at least `t` confirmations of a lack of malicious behavior and we have a usable key.

We actually want the full `n` to confirm the keys validity. Once we have at least `t` confirmations, we set a timer to expire in `(n-c)w + w` (where `c` is the amount of confirmations) time since the we received `t` confirmations. If we receive a new confirmation, we recreate the timer. Only once we have `n` confirmations or the timer expires would we move forward.

This has the benefit where if we have 100 validators, performing a DoS on 33 of them to attempt to exclude them from the DKG will incur a time delay of `33w` and adds resistance against such classes of attack.

Parties that don't participate on-time would be removed on-chain (as currently done).

If a broken key share is identified during the timer, we would reselect the participants to be the first `t` participants excluding the ones malicious.

kayabaNerve:
For the eVRF implementation, https://github.com/kayabaNerve/fcmp-plus-plus/ has all of the code necessary.

kayabaNerve:
The issue with the above protocol is that there's a risk of creating a `t`-of-`t` multisig. This is a risk as we only have `t` people confirm they received valid signature shares. The below is a proposal for a new method of transmitting the secret shares such that one can efficiently prove they transmitted correct shares.

---

To recap, we have an elliptic curve `T` with scalar field `p`. We then have elliptic curve `E` with field `p` and scalar field `l`.

The eVRF premised DKG is initialized with `n` public keys on the elliptic curve `E`, one per participant. Each participant performs ECDHs on `E` with their key and points sampled with a hash to curve by the group to produce deterministic, efficiently verifiable, secret points `C_i`. The `x` coordinates of these points (specifically, the sum of the `x` coordinates of two points) become the coefficients of the polynomial over the scalar field of `T`.

The eVRF paper simply says to send the secrets to each party. We assume a verifiable broadcast channel *which orders the messages broadcast over it*. We broadcast the commitments to the coefficients (with the relevant eVRF proof) and the encrypted secret shares here.

[...]

PR #589, opened 2024-08-06
One Round DKG

kayabaNerve:
Implement https://github.com/serai-dex/serai/issues/576.

Resolves lingering security issues with the DKG re-attempt protocol by removing the DKG re-attempt protocol, and moving to a DKG where there are no excluded parties upon completion (nor parties potentially with invalid shares).

This can't be merged. It adds ~10k lines of unaudited cryptography which is expected to be audited over the next 2-3 months. If it was merged now, it'd become in-scope automatically under our bug bounty program, and I'm unwilling to place such a bounty on such a degree of unaudited code. I do want this to the basis for further processor/coordinator clean ups though.


kayabaNerve:
There's some bug in set_keys which will cause client/coordinator/full-stack to fail at this time.

PR #590, opened 2024-08-08
add missing prevotes

akildemir:


kayabaNerve:
For context, this corresponds with lines 26/32 of the Tendermint specification. It prevents falsely logging we didn't receive a proposal (when we did) and slashing proposers who get rejected over temporal behavior (as we shouldn't).

kayabaNerve:
Classified bug due to how any nodes yet to receive a Provided transaction would slash nodes who proposed with such a Provided transaction.

PR #598, opened 2024-08-23
Monero: fix signature hash for V1 txs

Boog900:


kayabaNerve:
Getting the signature hash... with the signatures.

:facepalm:

PR #601, opened 2024-09-02
Move emissions tests to its pallet

akildemir:


substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -414,6 +416,10 @@ pub mod pallet {
       Pallet::<T>::deposit_event(Event::NewSet { set });
 
       Participants::<T>::set(network, Some(participants.try_into().unwrap()));
+      if network == NetworkId::Serai {
```
akildemir, line 421: This check also needs to include `session == Session(0)`  since we only need this for the first session. For all the next sessions `retire_set` is called for serai for network on `rotate_session`. 

Edit: This is fixed with https://github.com/serai-dex/serai/pull/614

PR #602, opened 2024-09-02
set correct swap fee, burn half

akildemir:


kayabaNerve:
This... burns 0.25% of the pool balance on every swap? Not half of the fee from each swap? Am I misreading this?

If I'm not, this allows draining the pool by swapping 1 atomic unit of SRI until the pool has effectively 0 SRI, then swapping 1 SRI (now worth most of the pool's value).

PR #603, opened 2024-09-02
prevent stake allocation during genesis period

akildemir:


kayabaNerve:
What's the context on this? Haven't we already defined an invariant that SRI isn't circulating at this point in time?

akildemir:
> What's the context on this? Haven't we already defined an invariant that SRI isn't circulating at this point in time?

Ah, sorry. This was in my TODO list before that discussion probably. I just went for it, forgot it was unnecessary. Will close this now.

PR #606, opened 2024-09-12
add coins pallet tests

akildemir:


substrate/coins/pallet/src/mock.rs
```diff
@@ -0,0 +1,70 @@
...
+  type SystemWeightInfo = ();
+  type SS58Prefix = ();
+  type OnSetCode = ();
+  type MaxConsumers = ConstU32<16>;
+}
+
+impl Config for Test {
+  type RuntimeEvent = RuntimeEvent;
+
+  type AllowMint = ();
```
akildemir, line 57: We don't set this to `ValidatorSets` pallet here mainly because it isn't even a dependency for the `Coins` pallet. Also we test the basic functionality for the coins pallet here so I doubt we need `AllowMint` check to be present that extra complicates the tests. It is present for other necessary pallets. Still not sure whether it should exist here or not.

PR #614, opened 2024-09-20
add validator sets pallet tests

akildemir:
These tests make the ones in `substrate/client/tests/validator_sets.rs` absolute. They would now only be valuable as testing the connection to pallet functions(txs) and also everything is still running as expected under a proper runtime and block production but not valuable as a test to actual pallet functionality. So there is a question to remove or still keep them. If we decide to keep them a better place might under `tests/` as an e2e tests.


kayabaNerve:
Speculative merge into next prior to substrate review.

PR #617, opened 2024-09-24
Constant-time divisors

kayabaNerve:
This is 9.6x slower yet ensures timing analysis of the divisor calculation doesn't leak any information about the secret key.


kayabaNerve:
Thanks to @Boog900 for reviewing my initial ct_* functions and helping improve then to what they are now.

Issue #573, opened 2024-06-22
Replace SCALE with borsh entirely


kayabaNerve:
I'm unsure this is feasible technically/feasible within our timeline.

It'd simplify the protocol though.

kayabaNerve:
This is completely infeasible.

If #647 occurs, we can define our own merklization scheme for events, and manually implement Encode/Decode on Event, Transaction, Block s.t. borsh is used internally. This should completely remove scale *from the blockchain itself*. Caveats apply as the RPC would be scale (#625 relevant) and consensus commits would presumably be. I honestly care less about consensus as we'll ideally replace it eventually, and at that point, we can checkpoint the last block produced under the old consensus after the transition.

kayabaNerve:
Prototyping this, solely for our own ABI, everything is fine except:
1) borsh is more inefficient due to the lack of VarInts. For `BoundedVec`, we can define it not as `Vec`, but as an enum however (where a max length of 512 is an outer enum of `00, 01, 02` and inner enums of `00 ..= ff, 00 ..= ff, 00`).  This commpresses our vector length prefixes while technically remaining within the borsh specification.
2) borsh treats writes as fallible due to io::Write being fallible. scale didn't, which is appreciated when dealing with `Vec` whose write doesn't fail, yet also did risk panics when internals technically could raise errors (which isn't appreciated).

kayabaNerve:
I'd consider this completed with `next-polkadot-sdk`'s work. We still use SCALE, but only inside Substrate, not within any of our own APIs 

Also, SCALE is pseudo-deprecated due to JAM/JAM codec, justifying this (even if retroactively).

PR #619, opened 2024-09-26
add specific network/coin/balance types

akildemir:
This pr adds new types such as `ExternalNetworkId`, `ExternalCoin` to distinctly identify them and help reduce the bugs. These changes helps code readability and prevent us from constantly checking for native network or coin.

### DIFF
--- a/coordinator/src/cosign_evaluator.rs
+++ b/coordinator/src/cosign_evaluator.rs
@@ -204,14 +204,12 @@ impl<D: Db> CosignEvaluator<D> {
 
       let mut total_stake = 0;
       let mut total_on_distinct_chain = 0;
-      // TODO: `network` isn't being used in the following loop. is this a bug?
-      // why are we going through the networks here?
-      for _network in EXTERNAL_NETWORKS {
+      for network in EXTERNAL_NETWORKS {
         // Get the current set for this network
         let set_with_keys = {
           let mut res;
           while {
-            res = set_with_keys_fn(&serai, cosign.network).await;
+            res = set_with_keys_fn(&serai, network).await;
             res.is_err()
           } {
             log::error!(
--- a/coordinator/src/cosign_evaluator.rs
+++ b/coordinator/src/cosign_evaluator.rs
@@ -12,9 +12,9 @@ use tokio::{
 use borsh::BorshSerialize;
 use sp_application_crypto::RuntimePublic;
 use serai_client::{
-  primitives::{NETWORKS, NetworkId, Signature},
-  validator_sets::primitives::{Session, ValidatorSet},
-  SeraiError, TemporalSerai, Serai,
+  primitives::{ExternalNetworkId, NetworkId, Signature, EXTERNAL_NETWORKS, NETWORKS},
+  validator_sets::primitives::{ExternalValidatorSet, Session},
+  Serai, SeraiError, TemporalSerai,
 };
 
 use serai_db::{Get, DbTxn, Db, create_db};
@@ -28,17 +28,17 @@ use crate::{
 
 create_db! {
   CosignDb {
-    ReceivedCosign: (set: ValidatorSet, block: [u8; 32]) -> CosignedBlock,
-    LatestCosign: (network: NetworkId) -> CosignedBlock,
-    DistinctChain: (set: ValidatorSet) -> (),
+    ReceivedCosign: (set: ExternalValidatorSet, block: [u8; 32]) -> CosignedBlock,
+    LatestCosign: (network: ExternalNetworkId) -> CosignedBlock,
+    DistinctChain: (set: ExternalValidatorSet) -> (),
   }
 }
 
 pub struct CosignEvaluator<D: Db> {
   db: Mutex<D>,
   serai: Arc<Serai>,
-  stakes: RwLock<Option<HashMap<NetworkId, u64>>>,
-  latest_cosigns: RwLock<HashMap<NetworkId, CosignedBlock>>,
+  stakes: RwLock<Option<HashMap<ExternalNetworkId, u64>>>,
+  latest_cosigns: RwLock<HashMap<ExternalNetworkId, CosignedBlock>>,
 }
 
 impl<D: Db> CosignEvaluator<D> {
@@ -79,7 +79,7 @@ impl<D: Db> CosignEvaluator<D> {
     let serai = self.serai.as_of_latest_finalized_block().await?;
 
     let mut stakes = HashMap::new();
-    for network in NETWORKS {
+    for network in EXTERNAL_NETWORKS {
       // Use if this network has published a Batch for a short-circuit of if they've ever set a key
       let set_key = serai.in_instructions().last_batch_for_network(network).await?.is_some();
       if set_key {
@@ -87,7 +87,7 @@ impl<D: Db> CosignEvaluator<D> {
           network,
           serai
             .validator_sets()
-            .total_allocated_stake(network)
+            .total_allocated_stake(network.into())
             .await?
             .expect("network which published a batch didn't have a stake set")
             .0,
@@ -126,9 +126,9 @@ impl<D: Db> CosignEvaluator<D> {
 
     async fn set_with_keys_fn(
       serai: &TemporalSerai<'_>,
-      network: NetworkId,
-    ) -> Result<Option<ValidatorSet>, SeraiError> {
-      let Some(latest_session) = serai.validator_sets().session(network).await? else {
+      network: ExternalNetworkId,
+    ) -> Result<Option<ExternalValidatorSet>, SeraiError> {
+      let Some(latest_session) = serai.validator_sets().session(network.into()).await? else {
         log::warn!("received cosign from {:?}, which doesn't yet have a session", network);
         return Ok(None);
       };
@@ -136,13 +136,13 @@ impl<D: Db> CosignEvaluator<D> {
       Ok(Some(
         if serai
           .validator_sets()
-          .keys(ValidatorSet { network, session: prior_session })
+          .keys(ExternalValidatorSet { network, session: prior_session })
           .await?
           .is_some()
         {
-          ValidatorSet { network, session: prior_session }
+          ExternalValidatorSet { network, session: prior_session }
         } else {
-          ValidatorSet { network, session: latest_session }
+          ExternalValidatorSet { network, session: latest_session }
         },
       ))
     }
@@ -231,7 +231,8 @@ impl<D: Db> CosignEvaluator<D> {
           let stake = {
             let mut res;
             while {
-              res = serai.validator_sets().total_allocated_stake(set_with_keys.network).await;
+              res =
+                serai.validator_sets().total_allocated_stake(set_with_keys.network.into()).await;
               res.is_err()
             } {
               log::error!(
@@ -271,7 +272,7 @@ impl<D: Db> CosignEvaluator<D> {
   #[allow(clippy::new_ret_no_self)]
   pub fn new<P: P2p>(db: D, p2p: P, serai: Arc<Serai>) -> mpsc::UnboundedSender<CosignedBlock> {
     let mut latest_cosigns = HashMap::new();
-    for network in NETWORKS {
+    for network in EXTERNAL_NETWORKS {
       if let Some(cosign) = LatestCosign::get(&db, network) {
         latest_cosigns.insert(network, cosign);
       }
--- a/coordinator/src/cosign_evaluator.rs
+++ b/coordinator/src/cosign_evaluator.rs
@@ -12,7 +12,7 @@ use tokio::{
 use borsh::BorshSerialize;
 use sp_application_crypto::RuntimePublic;
 use serai_client::{
-  primitives::{ExternalNetworkId, NetworkId, Signature, EXTERNAL_NETWORKS, NETWORKS},
+  primitives::{ExternalNetworkId, Signature, EXTERNAL_NETWORKS},
   validator_sets::primitives::{ExternalValidatorSet, Session},
   Serai, SeraiError, TemporalSerai,
 };
@@ -204,11 +204,9 @@ impl<D: Db> CosignEvaluator<D> {
 
       let mut total_stake = 0;
       let mut total_on_distinct_chain = 0;
-      for network in NETWORKS {
-        if network == NetworkId::Serai {
-          continue;
-        }
-
+      // TODO: `network` isn't being used in the following loop. is this a bug?
+      // why are we going through the networks here?
+      for _network in EXTERNAL_NETWORKS {
         // Get the current set for this network
         let set_with_keys = {
           let mut res;
--- a/coordinator/src/substrate/mod.rs
+++ b/coordinator/src/substrate/mod.rs
@@ -9,11 +9,14 @@ use zeroize::Zeroizing;
 use ciphersuite::{group::GroupEncoding, Ciphersuite, Ristretto};
 
 use serai_client::{
-  SeraiError, Block, Serai, TemporalSerai,
-  primitives::{BlockHash, NetworkId},
-  validator_sets::{primitives::ValidatorSet, ValidatorSetsEvent},
-  in_instructions::InInstructionsEvent,
   coins::CoinsEvent,
+  in_instructions::InInstructionsEvent,
+  primitives::{BlockHash, ExternalNetworkId, NetworkId},
+  validator_sets::{
+    primitives::{ExternalValidatorSet, ValidatorSet},
+    ValidatorSetsEvent,
+  },
+  Block, Serai, SeraiError, TemporalSerai,
 };
 
 use serai_db::DbTxn;
@@ -52,9 +55,9 @@ async fn handle_new_set<D: Db>(
   new_tributary_spec: &mpsc::UnboundedSender<TributarySpec>,
   serai: &Serai,
   block: &Block,
-  set: ValidatorSet,
+  set: ExternalValidatorSet,
 ) -> Result<(), SeraiError> {
-  if in_set(key, &serai.as_of(block.hash()), set)
+  if in_set(key, &serai.as_of(block.hash()), set.into())
     .await?
     .expect("NewSet for set which doesn't exist")
   {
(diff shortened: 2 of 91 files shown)


coordinator/src/cosign_evaluator.rs
```diff
@@ -204,11 +204,9 @@ impl<D: Db> CosignEvaluator<D> {
 
       let mut total_stake = 0;
       let mut total_on_distinct_chain = 0;
-      for network in NETWORKS {
-        if network == NetworkId::Serai {
-          continue;
-        }
-
+      // TODO: `network` isn't being used in the following loop. is this a bug?
+      // why are we going through the networks here?
```
kayabaNerve: Yes. This is meant to calculate the total stake across all networks. It's calculating the stake for the network which triggered this re-calc `n` times (where `n` is the amount of networks).

This alone seems to immediately validate this PR.

coordinator/src/substrate/mod.rs
```diff
@@ -232,6 +235,7 @@ async fn handle_block<D: Db, Pro: Processors>(
       continue;
     }
 
+    let set: ExternalValidatorSet = set.try_into().unwrap();
```
kayabaNerve: We can do `let Ok(set) = ExternalValidatorSet::try_from(set) else { continue };` to avoid the unwrap.

The unwrap is safe thanks to the above clause. We just can use the `try_into` result as the clause itself. This is a minor nit and doesn't need to be done to be merged.

coordinator/src/substrate/mod.rs
```diff
@@ -290,6 +294,7 @@ async fn handle_block<D: Db, Pro: Processors>(
...
```
kayabaNerve: Same note as above

coordinator/src/substrate/mod.rs
```diff
@@ -311,6 +316,7 @@ async fn handle_block<D: Db, Pro: Processors>(
...
```
kayabaNerve: Same note as above

substrate/coins/pallet/src/lib.rs
```diff
@@ -161,7 +161,12 @@ pub mod pallet {
     pub fn mint(to: Public, balance: Balance) -> Result<(), Error<T, I>> {
       // If the coin isn't Serai, which we're always allowed to mint, and the mint isn't explicitly
       // allowed, error
-      if (balance.coin != Coin::Serai) && (!T::AllowMint::is_allowed(&balance)) {
+      if !balance.coin.is_native() &&
+        (!T::AllowMint::is_allowed(&ExternalBalance {
+          coin: balance.coin.try_into().unwrap(),
+          amount: balance.amount,
+        }))
+      {
```
kayabaNerve, line 167: `if !ExternalCoin::try_from(balance.coin).map(|coin| T::AllowMint::is_allowed(&ExternalBalance { coin, amount: balance.amount })).unwrap_or(true)`

kayabaNerve: Single clause, doesn't have an unwrap (unreachable due to prior clause, of course, but adds an extra piece to the machine).

substrate/dex/pallet/src/lib.rs
```diff
@@ -509,8 +510,10 @@ pub mod pallet {
     /// A hook to be called whenever a network's session is rotated.
     pub fn on_new_session(network: NetworkId) {
       // reset the oracle value
-      for coin in network.coins() {
-        SecurityOracleValue::<T>::set(*coin, Self::median_price(coin));
+      if let NetworkId::External(n) = network {
```
kayabaNerve, line 513: Can we document why this bound is present, please? `// Only track the price for non-SRI coins as this is SRI denominated` or so?

substrate/dex/pallet/src/lib.rs
```diff
@@ -921,9 +937,9 @@ pub mod pallet {
       ensure!((coin1 == Coin::Serai) || (coin2 == Coin::Serai), Error::<T>::PoolNotFound);
       ensure!(coin1 != coin2, Error::<T>::EqualCoins);
       if coin1 == Coin::Serai {
-        Ok(coin2)
+        Ok(coin2.try_into().unwrap())
       } else {
-        Ok(coin1)
+        Ok(coin1.try_into().unwrap())
       }
```
kayabaNerve: I want to posit `ExternalCoin::try_from(coin1).or_else(|| ExternalCoin::try_from(coin2)).unwrap()`, yet that doesn't remove the unwrap.

Also, this PR introduces an edge case where we can never have a native coin which isn't Serai. That should be fine for now yet is worth noting.

substrate/in-instructions/primitives/src/lib.rs
```diff
@@ -2,6 +2,7 @@
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![cfg_attr(not(feature = "std"), no_std)]
 
+use serai_primitives::ExternalBalance;
```
kayabaNerve: Why do we have this stray import here instead of consolidated with the below imports from this crate?

substrate/primitives/src/amount.rs
```diff
@@ -29,6 +29,8 @@ pub type SubstrateAmount = u64;
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub struct Amount(pub SubstrateAmount);
 
+// TODO: these impl shouldn't panic and return error to be dealt with.
+// Otherwise we might have a panic that stops the network.
```
kayabaNerve, line 33: We can replace `Add` with `checked_add`. This `TODO` is fine to leave for now, but this should be replaced by `checked_add` (not modified to return an error).

substrate/in-instructions/pallet/src/lib.rs
```diff
@@ -210,40 +209,44 @@ pub mod pallet {
                     // TODO: Properly pass data. Replace address with an OutInstruction entirely?
                     data: None,
                   },
-                  balance: Balance { coin: out_balance.coin, amount: coin_balance },
+                  balance: ExternalBalance {
+                    coin: out_balance.coin.try_into().unwrap(),
```
kayabaNerve, line 213: This is a reachable panic if somehow someone specifies `Coin::Serai` with a non-native address (which I don't immediately see a prior check for). This should use `?`, not `.unwrap()`.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -362,16 +358,18 @@ pub mod pallet {
     pub fn swap_to_staked_sri(
       to: PublicKey,
       network: NetworkId,
-      balance: Balance,
+      balance: ExternalBalance,
     ) -> DispatchResult {
       // check the network didn't reach the economic security yet
-      if EconomicSecurity::<T>::economic_security_block(network).is_some() {
-        Err(Error::<T>::NetworkHasEconomicSecurity)?;
+      if let NetworkId::External(n) = network {
```
kayabaNerve, line 367: How do we feel about the fact people can swap to staked SRI for the Serai network even if the Serai network has the amount of stake it should? By the economics, we target 20% of the network's stake to be behind the Serai network itself. Maybe worth an additional clause in that regard?

substrate/dex/pallet/src/tests.rs
```diff
@@ -18,7 +18,10 @@
 // It has been forked into a crate distributed under the AGPL 3.0.
 // Please check the current distribution for up-to-date copyright and licensing information.
 
-use crate::{mock::*, *};
+use crate::{
+  mock::{*, MEDIAN_PRICE_WINDOW_LENGTH},
```
kayabaNerve, line 22: If this is importing everything, why is it also importing this constant?

substrate/emissions/pallet/src/lib.rs
```diff
@@ -250,7 +249,7 @@ pub mod pallet {
         } else {
           // calculate pool vs validator share
           let capacity = ValidatorSets::<T>::total_allocated_stake(n).unwrap_or(Amount(0)).0;
-          let required = ValidatorSets::<T>::required_stake_for_network(n);
+          let required = ValidatorSets::<T>::required_stake_for_network(n.try_into().unwrap());
```
kayabaNerve: This is `if n == NetworkId::Serai { .. } else { ... n.try_into().unwrap() }`. We can use a match to avoid this unwrap.

substrate/emissions/pallet/src/lib.rs
```diff
@@ -272,13 +271,14 @@ pub mod pallet {
           let vpc = volume_per_coin.as_ref().unwrap();
           for c in n.coins() {
             let pool_reward = u64::try_from(
-              u128::from(network_pool_reward).saturating_mul(u128::from(vpc[c])) /
+              u128::from(network_pool_reward)
+                .saturating_mul(u128::from(vpc[&c.try_into().unwrap()])) /
```
kayabaNerve: I'd write a single `c.try_into().unwrap()` (there's one following) with the clarifying comment `network_pool_reward` is 0 for Serai.

I'd also ask if we can merge this into the above if/else since it only happens for external networks.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -690,20 +695,23 @@ pub mod pallet {
         return false;
       }
 
-      // Handover is automatically complete for Serai as it doesn't have a handover protocol
-      if network == NetworkId::Serai {
-        return true;
-      }
+      if let NetworkId::External(n) = network {
```
kayabaNerve: Considering the `else` is just `true`, this could just be `let NetworkId::External(n) = network else { return true };`. Unnecessary nit, just preserves the prior level of indentation.

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -171,6 +171,7 @@ pub mod pallet {
     ///
     /// This will still include participants which were removed from the DKG.
     pub fn in_set(network: NetworkId, account: Public) -> bool {
+      // TODO: should InSet only work for `ExternalNetworkId`?
```
kayabaNerve: Why would it? Do we just never check if Serai validators are in-set?

substrate/primitives/src/networks.rs
```diff
@@ -10,38 +10,98 @@ use borsh::{BorshSerialize, BorshDeserialize};
...
-  Serai,
+pub enum ExternalNetworkId {
   Bitcoin,
   Ethereum,
   Monero,
 }
+
+/// The type used to identify networks.
+#[derive(
+  Clone, Copy, PartialEq, Eq, Hash, Debug, Encode, Decode, PartialOrd, Ord, MaxEncodedLen, TypeInfo,
```
kayabaNerve: We shouldn't derive Encode/Decode/MaxEncodedLen. This is now a one/two-byte encoding as there's one byte used for Serai/External, and one byte used for the ExternalNetworkId. We can manually implement it such that Serai == 0, ExternalNetworkId = id.encode()[0] + 1.

We can also manually implement it for ExternalNetworkId so that ENId starts at 1. This would preserve the existing encoding format and allow upgrading ENId instances to NId instances in the future, without changing the wire format/invalidating storage (though we would need a clause banning using the ::Serai variant prior to some block height if we ever introduced such an expansion). In that case, the NId encoding would simply be `if Serai [0] else enid.encode()`.

substrate/primitives/src/networks.rs
```diff
@@ -52,12 +112,81 @@ pub const COINS: [Coin; 5] = [Coin::Serai, Coin::Bitcoin, Coin::Ether, Coin::Dai
 #[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
 pub enum Coin {
   Serai,
+  External(ExternalCoin),
+}
+
+/// The type used to identify external coins.
+#[derive(
+  Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug, Encode, Decode, MaxEncodedLen, TypeInfo,
```
kayabaNerve: The same commentary on manually implementing Encode/Decode/MaxEncodedLen stands as it did for NetworkId. There is the slight caveat we may want to start not at `1` yet `4` or `8`. This allows using the lower three bits for native coins (if we ever add, say, a native USD analogue).

Also, slight nit, can we define ExternalCoin prior to Coin please?

substrate/primitives/src/networks.rs
```diff
@@ -52,12 +112,81 @@ pub const COINS: [Coin; 5] = [Coin::Serai, Coin::Bitcoin, Coin::Ether, Coin::Dai
...
+pub enum ExternalCoin {
   Bitcoin,
   Ether,
   Dai,
   Monero,
 }
 
+impl From<ExternalCoin> for Coin {
+  fn from(coin: ExternalCoin) -> Self {
+    match coin {
```
kayabaNerve: `Coin::External(coin)` can replace this entire match

substrate/primitives/src/networks.rs
```diff
@@ -10,38 +10,98 @@ use borsh::{BorshSerialize, BorshDeserialize};
...
+      Self::External(network) => {
+        network.coins().into_iter().map(core::convert::Into::into).collect()
+      }
+    }
+  }
+}
+
+impl From<ExternalNetworkId> for NetworkId {
+  fn from(network: ExternalNetworkId) -> Self {
+    match network {
```
kayabaNerve: `NetworkId::External(network)` can replace this entire `match`

substrate/primitives/src/networks.rs
```diff
@@ -10,38 +10,98 @@ use borsh::{BorshSerialize, BorshDeserialize};
...
   Ethereum,
   Monero,
 }
+
+/// The type used to identify networks.
+#[derive(
+  Clone, Copy, PartialEq, Eq, Hash, Debug, Encode, Decode, PartialOrd, Ord, MaxEncodedLen, TypeInfo,
+)]
+#[cfg_attr(feature = "std", derive(Zeroize))]
+#[cfg_attr(feature = "borsh", derive(BorshSerialize, BorshDeserialize))]
```
kayabaNerve: There's also some question of if we should similarly update Borsh. We don't need to, as Borsh is intended to be used over-the-wire as some canonical format (and is instead just used in our local DB), but we have the same commentary our wire-format flexibility.

Also, cc #573 as a relevant discussion here.

I'll also note manually defining `BorshSerailize/BorshDeserialize` to be the scale `Encode/Decode` (not replicating the same function bodies yet explicitly calling into scale) is more than fine. I'd not complain about that to de-duplicate the logic.

kayabaNerve (commented):
Really great catch on what we should do. Bunch of nits which honestly don't matter, yet the wire-format comments are worth discussing!

substrate/dex/pallet/src/lib.rs
akildemir: We don't need the `unwrap()` here. We can just do `.map_err()` at the end and return the result. Also, I think the assumption of we can't have a native coin other than Serai was already in the codebase but implicitly, this pr just makes it explicit.

substrate/dex/pallet/src/tests.rs
akildemir: Because we have 2 of these, One in the `primitives/src/constants.rs` and other one is in the `dex/pallet/src/mock.rs`, compiler doesn't know which to use if we do `*` on both files. So we import the `mock.rs` one explicityly.

substrate/in-instructions/pallet/src/lib.rs
akildemir: I think you mean this check? https://github.com/akildemir/serai/blob/c66650d259be95871de10b411ec5ea18c49cbb4f/substrate/in-instructions/pallet/src/lib.rs#L168

substrate/validator-sets/pallet/src/lib.rs
```diff
@@ -733,19 +741,26 @@ pub mod pallet {
...
-        Self::deposit_event(Event::SetRetired {
-          set: ValidatorSet { network: set.network, session: Session(set.session.0 - 1) },
-        });
-      }
-
       // Serai doesn't set keys and network slashes are handled by BABE/GRANDPA
-      if set.network != NetworkId::Serai {
+      if let NetworkId::External(n) = set.network {
+        // If the prior prior set didn't report, emit they're retired now
+        // TODO: we will emit the events 1 session late if there was no call to report_slashes.
```
akildemir: @kayabaNerve no comments for this one?

substrate/primitives/src/networks.rs
akildemir: Second option does seem to make sense to me. Should we do same for the `Coin` and `ExternalCoin` types?

substrate/emissions/pallet/src/lib.rs
```diff
@@ -362,16 +361,30 @@ pub mod pallet {
...
+        let mut total_stake = 0;
+        for n in NETWORKS {
+          total_stake += ValidatorSets::<T>::total_allocated_stake(n).unwrap_or(Amount(0)).0;
+        }
+
+        let stake = ValidatorSets::<T>::total_allocated_stake(network).unwrap_or(Amount(0)).0;
+        let desired_stake = total_stake / (100 / SERAI_VALIDATORS_DESIRED_PERCENTAGE);
+        if stake >= desired_stake {
+          Err(Error::<T>::NetworkHasEconomicSecurity)?;
+        }
```
kayabaNerve, line 382: Hm. I'm fine kicking this to a separate PR if we can't immediately resolve it, yet I do want to discuss this algorithm more. This limits to 20% of the current stake to external networks. That works if and only if Serai is the last network to be staked for. It does not allow Serai to be the first algorithm to be staked for.

From the initial distribution, we can calculate how much the Serai network should have and simply use that value as the limit.

That has the following quirks in behavior:
1) It allows the Serai set to be staked for at an arbitrary point in time, with a much simpler check here
2) If various LPs leave (burning the SRI, reducing the amount needed for economic security), then we may have more SRI stake than targeted

In the case of 2, we simply revert to relying on the emissions distribution to maintain the optimal distribution. I do have a concern where a large amount of people stake to the SRI network *just to swap to SRI at the spot price*. While we're discussing non-trivial slashes re: external networks for offline validators, running a Serai node may only have trivial slashes *and is far less annoying to setup a node for*. Then since the Serai network is perceived as having infinite capacity, validators can immediately unstake after economic security with their spot-price SRI. The first option (currently implemented) does better in that regard by ensuring the amount staked to SRI is only the amount needed.

Also, further edge case, the final stake which causes economic security is unbounded. This is not too likely to be an issue? But likely yet another reason to open an issue for this and kick it to another PR (and if we do that, I'm fine leaving this here as the initial implementation).

substrate/dex/pallet/src/tests.rs
kayabaNerve: Ugh. That's a mess. Probably better to remove a glob import (as they introduce such issues) but it's fine.

PR #629, opened 2024-11-15
Monero: expose merkle root function

Boog900:
This function is needed in the RPC: https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/rpc/core_rpc_server.cpp#L2172

I also made a small change to the function signature, taking all leafs in one arg,

### DIFF
--- a/networks/monero/src/lib.rs
+++ b/networks/monero/src/lib.rs
@@ -7,7 +7,7 @@ pub use monero_io as io;
 pub use monero_generators as generators;
 pub use monero_primitives as primitives;
 
-/// Merkel tree functionality.
+/// Merkle tree functionality.
 pub mod merkle;
 
 /// Ring Signature structs and functionality.
--- a/networks/monero/src/lib.rs
+++ b/networks/monero/src/lib.rs
@@ -7,7 +7,8 @@ pub use monero_io as io;
 pub use monero_generators as generators;
 pub use monero_primitives as primitives;
 
-mod merkle;
+/// Merkel tree functionality.
+pub mod merkle;
 
 /// Ring Signature structs and functionality.
 pub mod ring_signatures;
--- a/networks/monero/src/merkle.rs
+++ b/networks/monero/src/merkle.rs
@@ -2,14 +2,13 @@ use std_shims::vec::Vec;
 
 use crate::primitives::keccak256;
 
-pub(crate) fn merkle_root(root: [u8; 32], leafs: &[[u8; 32]]) -> [u8; 32] {
+/// Calculates the merkel root of the given tree. Equivalent to `tree_hash` in monero-core.
+pub fn merkle_root(leafs: &[[u8; 32]]) -> [u8; 32] {
   match leafs.len() {
-    0 => root,
-    1 => keccak256([root, leafs[0]].concat()),
+    1 => leafs[0],
+    2 => keccak256([leafs[0], leafs[1]].concat()),
     _ => {
-      let mut hashes = Vec::with_capacity(1 + leafs.len());
-      hashes.push(root);
-      hashes.extend(leafs);
+      let mut hashes = leafs.to_vec();
 
       // Monero preprocess this so the length is a power of 2
       let mut high_pow_2 = 4; // 4 is the lowest value this can be
--- a/networks/monero/src/merkle.rs
+++ b/networks/monero/src/merkle.rs
@@ -2,25 +2,29 @@ use std_shims::vec::Vec;
 
 use crate::primitives::keccak256;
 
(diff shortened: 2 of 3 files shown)


networks/monero/src/lib.rs
```diff
@@ -7,7 +7,8 @@ pub use monero_io as io;
 pub use monero_generators as generators;
 pub use monero_primitives as primitives;
 
-mod merkle;
+/// Merkel tree functionality.
```
kayabaNerve: Typo.

networks/monero/src/merkle.rs
```diff
@@ -2,14 +2,13 @@ use std_shims::vec::Vec;
 
 use crate::primitives::keccak256;
 
-pub(crate) fn merkle_root(root: [u8; 32], leafs: &[[u8; 32]]) -> [u8; 32] {
+/// Calculates the merkel root of the given tree. Equivalent to `tree_hash` in monero-core.
```
kayabaNerve: Merkle is a proper noun as it's a name. Also, typo.

monero-core? monero-project/monero? Should we cite a link to tree_hash here?

networks/monero/src/merkle.rs
```diff
@@ -2,14 +2,13 @@ use std_shims::vec::Vec;
...
   match leafs.len() {
-    0 => root,
-    1 => keccak256([root, leafs[0]].concat()),
+    1 => leafs[0],
+    2 => keccak256([leafs[0], leafs[1]].concat()),
     _ => {
-      let mut hashes = Vec::with_capacity(1 + leafs.len());
-      hashes.push(root);
-      hashes.extend(leafs);
+      let mut hashes = leafs.to_vec();
```
kayabaNerve: If we effectively immediately convert to a Vec, we should have the argument be a Vec so there's a chance this is a move, not a clone.

networks/monero/src/merkle.rs
```diff
@@ -2,14 +2,13 @@ use std_shims::vec::Vec;
 
 use crate::primitives::keccak256;
 
-pub(crate) fn merkle_root(root: [u8; 32], leafs: &[[u8; 32]]) -> [u8; 32] {
+/// Calculates the merkel root of the given tree. Equivalent to `tree_hash` in monero-core.
+pub fn merkle_root(leafs: &[[u8; 32]]) -> [u8; 32] {
   match leafs.len() {
-    0 => root,
```
kayabaNerve, line 7: We used to handle no leaves being passed. Now no hashes falls through to the n-leaves case. Is that correct?

I doubt it as our n-leaf code pads to the nearest power of 2 which is assumed to be 4 as we assumed exceptional handling for 0/1/2.

kayabaNerve (changes requested):
ACK to the concept. NACK to the implementation which I believe is faulty.

Also, the monero folder has maintained a code freeze pending audit. How should we handle this? Merge this on Serai's next? Merge this into a new branch without the code freeze? Cuprate vendors a version with select patches we roll in post-audit?

networks/monero/src/merkle.rs
Boog900: Ah yeah it will panic, I guess the best thing to do would be to explicitly panic, `tree_hash` doesn't allow a 0 length input: https://github.com/monero-project/monero/blob/893916ad091a92e765ce3241b94e706ad012b62a/src/crypto/tree-hash.c#L62

kayabaNerve: That would've been so easy for me to check if there was a link to `tree_hash` present :p

*There wasn't such a link prior, I don't legitimately want to hold this against you.

I'd vote we keep the existing explicit argument (potentially renamed from root to first_leaf?) making this unreachable, or return an Option. We can document the panic but I'd consider that a downgrade. Even the Option would be but it isn't such a downgrade I wouldn't agree to it if you believe avoiding these allocations so worthwhile.

Boog900:
> Also, the monero folder has maintained a code freeze pending audit. How should we handle this?

On Cuprate's end we already vendor `monero-serai`, so I think the last option, having us vendor a version with select patches until the audit is complete, would probably be best.

networks/monero/src/merkle.rs
Boog900: I'll change it to return an option, I don't really like having the first leaf as a separate arg.

kayabaNerve (approved):
LGTM, yet won't merge due to code freeze.

kayabaNerve:
Closed due to https://github.com/monero-oxide/monero-oxide/pull/27 achieving this on monero-oxide.

PR #667, opened 2025-08-09
Monero Audit

kayabaNerve:
Responds to the findings from Cypher Stack's audit, as applicable. The notes on the Monero protocol themselves have been ignored, though carefully considered re: documentation. Additionally, the transcripting re: the multisig code has been left as-is. While the audit notes it isn't to preference, and can be hard to follow, the audit confirms its validity. In my opinion, it'd be worse to change harder-to-read, audited-as-correct code than have easy-to-read, unaudited code.

The move away from `unwrap`, overall, has produced better code and is pleasant.


kayabaNerve:
While several of these commits are fixing a CI which has rotted over some months, the plan is to merge _via rebase_ once everything _except_ deny passes.

kayabaNerve:
Processor Tests failed with latest Docker, and started passing with Docker 25. It definitely seems to be Docker version trickery...

kayabaNerve:
After one more attempt to downgrade Docker to v24 and see if that fixes it, I'm removing any edits not verified to actually benefit and no longer trying to fix the decayed CI in this PR, as it's gotten out of hand.

Issue #624, opened 2024-10-18
dkg 2.0


kayabaNerve:
The current `dkg` crate (or the one on the `next` branch) should be published as 1.0. That doesn't change it has flaws we should fix.

1) `dkg` shouldn't actually have any DKGs. It should just be the `ThresholdKeys` type. PedPoP should be moved to its own crate, as should Musig (which I did register the crates for). This would allow anyone to add their own DKG, such as [Chill DKG](https://github.com/BlockstreamResearch/bip-frost-dkg).
2) Only having additive offsets/lagrange interpolation is insufficient. We should support multiplicative offsets, additive offsets, and either constant or lagrange interpolation (as seen with the `dkg` crate on the `next` branch). This was also noted in #195.
3) Replacing `ThresholdKeys` with `BiasedKeyShare` and `UnbiasedKeyShare`. This would allow higher-level protocols to specify which they need (Bitcoin needs a tweaked biased key or an unbiased key, Ethereum needs an unbiased key if the key is fed into keccak160 as addresses are).

Multiplicative offsets will be needed by the time Carrot comes around (the Monero FCMP++ hard fork) yet we shouldn't start working on this before Serai launches.

#195 was closed as it was cool things for a 2.0 yet nothing with concrete benefit. All of these do have concrete benefit, hence this issue existing now and standing.

kayabaNerve:
Abstracted DKG's also may enable including obscure DKGs such as Monero's ECDH-premised DKG. I'm unsure it would in practice as Monero's DKG may not yield verification shares. While we could offer VerifiableBiasedKeyShare/VerifiableUnbiasedKeyShare (or rather `Verifiable + Unbiased + KeyShare`), I'm unsure we actually want to support algorithms without identifiable aborts.

kayabaNerve:
All of this seems resolved with `dkg 0.6`, except:
1) Lack of differentiating biased from unbiased
2) Lack of differentiating with verification shares from without

PR #669, opened 2025-08-18
Merge crypto from next

kayabaNerve:
This does not merge the one-round eVRF-premised DKG from the `next` branch. It checks out the entire folder, removes the eVRF-premised DKG, and then proceeds to reorganize `dkg` into a smashed set of crates. This lets:
1) The changes from `next` to the `ThresholdKeys` struct to be applied
2) Us merge in the one-round eVRF-premised DKG. We only had the next branch as `dkg` is covered by a bug bounty, and the eVRF-premised DKG was written into `dkg`, so merging it would in-scope it under the bug bounty. With this new modular approach to `dkg` (which is optimal regardless as we now have multiple algorithms), we now have that option and can continue further merging `next`.
3) `monero-oxide` update to the new version of `dkg`.
4) The `next` branch replace its inlined `monero-serai` with `monero-oxide`.


kayabaNerve:
Largely resolves #624  .

PR #678, opened 2025-09-22
Move `develop` to `patch-polkadot-sdk`

kayabaNerve:


kayabaNerve:
For some reason, this isn't peering at all with the libp2p backend. It does peer with the `litep2p` backend, but the authority discovery RPC route isn't working...

kayabaNerve:
`libp2p` peering was fixed with https://github.com/serai-dex/patch-polkadot-sdk/commit/e4f65cdde0db2e3719d6fbcfcb17897767b460d6, but authority discovery still isn't working as expected...

Issue #705, opened 2025-12-10
Shells used in testing could be further filled out


kayabaNerve:
`pdksh` is a public-domain alternative to `ksh` from which most of the `*ksh` forked from. While `ksh` refers to `ksh93`, `pdksh` reimplemented `ksh88`.

[`pbosh`](https://codeberg.org/schilytools/schilytools/src/branch/master/pbosh) attempts to be perfectly POSIX, no more and no less. Unfortunately, it isn't packaged seemingly anywhere. While [`mrsh`](https://github.com/emersion/mrsh) has the same goal, [`mrsh` doesn't claim to be complete](https://github.com/emersion/mrsh/issues/8) (`pbosh` should be) and `mrsh` [does not work with our script](https://github.com/serai-dex/serai/issues/704) (for whatever reason).

[`loksh`](https://github.com/dimkr/loksh) is another `ksh`-derivative from OpenBSD, and is packaged for Alpine but not Ubuntu or macOS.

[`GWSH`](https://github.com/hvdijk/gwsh) is allegedly a full shell derived from `dash`, solely not one packaged anywhere.

[(Heirloom) Bourne `sh`](https://heirloom.sourceforge.net/sh.html) isn't POSIX, solely very approximate due to how it was pre-POSIX and inspiring, and would be neat to see if it was usable. It seemingly isn't packaged here.

As I've hunted for shells, it should be noted these shells may themselves be at fault for any errors encountered.

kayabaNerve:
`pdksh` does not support hexadecimal numbers, as identified with the CI for https://github.com/serai-dex/serai/pull/707. While it has an option for POSIX compliance, that option's effects are enumerated and changes to how numeric literals are interpreted is not one of them. This is the same disqualifier as `mrsh` and possibly would be for Bourne sh too.

That leaves `pbosh`, `loksh`, and `GWSH`, with the interesting one being `pbosh` for its behavior (as we already have a distinct BSD's `ksh` fork and already have `dash`).

kayabaNerve:
The workflow now includes `bsh`, `obosh`, `pbosh`, `loksh`, `gwsh`, `nsh`, and `gosh`. The only other idea I have would be to include `toysh` (`toybox sh`), but that's marked experimental so I don't want to solve for building it and effectively testing it at this time.

PR #706, opened 2025-12-10
`pdksh`, `ksh93` shells via MacPorts

kayabaNerve:


kayabaNerve:
`pdksh` has a broken MacPort. `ksh93` is already present via Homebrew (oops). While ideally we would have `pdksh` (#705), I'll close this PR.

PR #710, opened 2025-12-19
Cosign tests

rafael-xmr:
# Changes:

### 1) [bug] wrong usage of .max():

This line:

```rust
// Set a limit of sleeping for two minutes
*current_sleep_before_next_task = new_sleep.max(Self::MAX_DELAY_BETWEEN_ITERATIONS);
```

Had a bug. As the `.max()` typedocs define:

> Compares and returns the maximum of two values.
>
> Returns the second argument if the comparison determines them to be equal.
>
> ### Examples
>
> ```rust
> assert_eq!(1.max(2), 2);
> assert_eq!(2.max(2), 2);
> ```

The intended use of `.max()` here was to, given a `new_sleep` time, and a maximum constant value in `Self::MAX_DELAY_BETWEEN_ITERATIONS`, the `new_sleep` time should never be greater than the constant and if greater, always default the sleeping time to the constant's time instead. Otherwise, a forever increasing `new_sleep` could occur and hold tasks indefinitely.

### 1) [bug] Zero-stake `Event::SetKeys` would panic

Identified a bug would panic every time on a Notable `Event::SetKeys` event where the total stake amounts to 0, and would always come back to panic preventing the cosign task from ever continuing. A 0 total stake notable event scenario is a possibility and this should be handled instead, for example at network's launch as there is a genesis period where liquidity is provided before any coins and any stake yet exist. The solution being adding the following:

```rust
// Handle declarations of the latest set
for event in vset_events.set_keys_events() {
  let Event::SetKeys { set, key_pair } = event else {
    unreachable!("event from `set_keys_events` wasn't `Event::SetKeys`")
  };

  let validators = Validators::take(&mut txn, *set)
    // critical panic:
    // this is a critical issue and will not be solved after re-tries,
    // missing Validators from previous blocks will remain missing until re-indexed
    // if encountered halt the process
    .expect("set which wasn't decided set keys");

  let stake: u64 = validators
    .iter()
    .map(|v| Stakes::get(&txn, set.network, *v).unwrap_or(Amount(0)).0)
    .sum();

  // Sets with 0 stake should be skipped and not considered w.r.t. cosigning
  // for no set with stake then has_events will remain HasEvents::No for this block and ignored
  if stake > 0 {
    has_events = HasEvents::Notable;
    LatestSet::set(
      &mut txn,
      set.network,
      &Set { session: set.session, key: key_pair.0, stake: Amount(stake) },
    );
  } else {
    serai_env::trace!(
      "{block_number}: skipped session {:?} of {:?} with 0 stake from being selected for cosigns",
      set.session,
      set.network
    );
  }
}
```

Where not only the `if stake > 0 {` was added to guard the addition of a `LatestSet` but also `has_events = HasEvents::Notable;` is only set as a Notable event type there to guard against the later condition that initiates a new global session for this event. Events with 0 stake don't need to be considered as Notable and can be skipped w.r.t. the cosigning protocol.

### 2) [feat] Empty validator set from `Event::SetDecided`

Added a sanity check:

```rust
if validators.is_empty() {
    panic!("validator set from Event::SetDecided was empty");
}
```

### 3) [bug] Block indexing start

The current implementation begins indexing by block 1, changed it to:

```rust
let start_block_number = ScanCosignFrom::get(&self.db).unwrap_or(0);
```

Otherwise gets the task permanently stuck with error `node's block #1 doesn't build upon the block #0 prior indexed` if genesis was not indexed.

### 1) [bug] added initial check for a global session because it is not possible to evaluate cosigns for a non-existing global session, so skip.

Now the evaluator initiates a new BlockEvent with:

```rust
// If no session is being evaluated yet, check if this block can be processed
if currently_evaluated_global_session(&txn).is_none() {
  match GlobalSessionsChannel::peek(&txn) {
    // No global session declared yet: this block predates all sessions, skip it
    // this means only HasEvents:No blocks have been consumed so far
    None => {
      commit_evaluated_block(txn, block_number, false);
      made_progress = true;
      continue;
    }
    // Session queued but starts after this block, skip it
    Some(next) if next.1.start_block_number > block_number => {
      commit_evaluated_block(txn, block_number, false);
      made_progress = true;
      continue;
    }
    // Session covers this block: proceed normally
    _ => {}
  }
}
```

### 2) [feat] added has_events to CosignedBlocks so delay can skip no event blocks

Now the db entry looks like:

```rust
db_channel!(
  SubstrateCosignEvaluatorChannels {
    // (cosigned block, time cosign was evaluated, has_events)
    CosignedBlocks: () -> (u64, u64, bool),
  }
);
```

and the addition of the `bool` for `has_events` allows the delay task to not need to sleep on `HasEvents::No` blocks

### 1) [feat] sanity check to avoid index regression

Simply made it consider skipping later blocks if already indexed, following a rule to not regress indexing.

```rust
if block_number <= latest_cosigned_block_number {...}
```

### 2) [feat] skip no event blocks from delay

as explained above with the evaluator, this was added:

```rust
// No events means no cosigns to wait for, mark as cosigned immediately
if !has_events {
  LatestCosignedBlockNumber::set(&mut txn, &block_number);
  txn.commit();
  made_progress = true;
  continue;
}

```

### 3) [bug] fixed the wrong calculation being used for the time to sleep

Now calculates as

```rust
let now_timestamp = now_timestamp().as_secs();
let time_valid_timestamp = time_evaluated + ACKNOWLEDGEMENT_DELAY.as_secs();
```

### 1) [bug] fixed cosign task handles not being held and being dropped

Now uses a:

```rust
// Forget the intend task handle, as dropping the handle would stop the task
// keeps all cosign tasks running in the background
core::mem::forget(intend_task_handle);
```

### 2) [bug] fixed Cosigning::latest_cosigned_block_number() always defaulting to 0, as if block 0 was already cosigned

Returns a `Result<Option<...>>` instead:

```rust
/// The latest cosigned block number.
pub fn latest_cosigned_block_number(getter: &impl Get) -> Result<Option<u64>, Faulted> {
    if FaultedSession::get(getter).is_some() {
      Err(Faulted)?;
    }

    Ok(LatestCosignedBlockNumber::get(getter))
}
```

And is more clear when a block has actually been indexed/cosigned or not

## Misc

- `Cargo.toml`: `unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage)'] }` hides spammy warnings when running tests.
- `coordinator/cosign/README.md`: re-worded a bit the phrasing in an attempt to make it clearer at least to myself when re-reading

---

```
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

```
contribution.
```


coordinator/cosign/Cargo.toml
```diff
@@ -31,3 +31,7 @@ serai-db = { path = "../../common/db", version = "0.1.1" }
 serai-task = { path = "../../common/task", version = "0.1" }
 
 serai-cosign-types = { path = "./types" }
+
+[dev-dependencies]
+env_logger = { version = "0.11", default-features = false }
```
kayabaNerve: The repository currently uses `0.10`.

substrate/abi/src/modules/validator_sets.rs
```diff
@@ -11,6 +11,8 @@ use serai_primitives::{
   validator_sets::*,
 };
 
+pub use serai_primitives::validator_sets::DeallocationTimeline;
```
kayabaNerve: Why re-export here to access as `serai_abi::validator_sets::DeallocationTimeline` compared to accessing as `serai_abi::primitives::validator_sets::DeallocationTimeline`, as everything else is accessed?

coordinator/cosign/src/lib.rs
```diff
@@ -160,53 +216,62 @@ impl IntakeCosignError {
...
       })
-      .continually_run(evaluator_task, vec![delay_task_handle]),
+      .continually_run(evaluator_task, vec![delay_task_handle.clone()]),
     );
     tokio::spawn(
       (delay::CosignDelayTask { db: db.clone() })
         .continually_run(delay_task, tasks_to_run_upon_cosigning),
     );
-    Self { db }
+    Self { db, _task_handles: vec![intend_task_handle, evaluator_task_handle, delay_task_handle] }
```
kayabaNerve: I believe this can replaced with `core::mem::forget(intend_task_handle)`. Cloning these when they're already actively held by a sub-task isn't great if we will not actually potentially use the handle. Storing solely the `delay_task_handle` would also be fine though.

coordinator/cosign/types/Cargo.toml
```diff
@@ -17,6 +17,10 @@ rustdoc-args = ["--cfg", "docsrs"]
 [lints]
 workspace = true
 
+[features]
+default = []
```
kayabaNerve: As `default` is empty, it seems unnecessary to be added, even if without issue.

coordinator/cosign/src/lib.rs
```diff
@@ -35,6 +41,56 @@ mod delay;
...
+#[cfg(test)]
+/// Test helpers and fixtures.
+pub mod tests;
+
+/// The interval at which the cosigning loop runs.
+#[cfg(not(test))]
+pub const COSIGN_LOOP_INTERVAL: Duration = Duration::from_secs(5);
+/// The interval at which the cosigning loop runs (shortened for tests).
+#[cfg(test)]
+pub const COSIGN_LOOP_INTERVAL: Duration = Duration::from_millis(10);
```
kayabaNerve: This _was_ defined in `coordinator/src/main.rs` (outside of this library) and the `tests` module also could've _locally_ defined the constant for usage within its test environment. I don't see why we need the constants here and so configured.

coordinator/src/main.rs
```diff
@@ -407,8 +405,8 @@ async fn main() {
       let mut key_bytes = serai_key.to_bytes();
       // Schnorrkel SecretKey is the key followed by 32 bytes of entropy for nonces
       let mut expanded_key = Zeroizing::new([0; 64]);
-      expanded_key.as_mut_slice()[.. 32].copy_from_slice(&key_bytes);
-      OsRng.fill_bytes(&mut expanded_key.as_mut_slice()[32 ..]);
+      expanded_key.as_mut_slice()[..32].copy_from_slice(&key_bytes);
+      OsRng.fill_bytes(&mut expanded_key.as_mut_slice()[32..]);
```
kayabaNerve: This PR as a whole has a bunch of misc formatting changes. While not fundamentally an issue, would it be possible to revert these out?

coordinator/cosign/types/src/tests/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    block_hash: BlockHash([5u8; 32]),
+    notable: true,
+  };
+
+  let cosign = intent.into_cosign(ExternalNetworkId::Bitcoin);
+
+  assert_eq!(cosign.global_session, [1u8; 32]);
+  assert_eq!(cosign.block_number, 5);
+  assert_eq!(cosign.block_hash, BlockHash([5u8; 32]));
+  assert_eq!(cosign.cosigner, ExternalNetworkId::Bitcoin);
```
kayabaNerve: If you deconstruct the cosign with `let Cosign { global_session, .. } = intent.into_cosign()`, we'd be able to ensure we don't omit any fields here accidentally.

coordinator/cosign/types/src/tests/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+}
+
+#[test]
+fn cosign_intent_to_cosign() {
+  let intent = CosignIntent {
+    global_session: [1u8; 32],
+    block_number: 5,
+    block_hash: BlockHash([5u8; 32]),
+    notable: true,
+  };
```
kayabaNerve: This is fine, but in general, it'd be better to use an RNG in case there's an edge case. Specifically, when serializing `enum`s or _variable-length fields_, it's preferred. Possibly, `random_cosign_intent()` fixtures?

coordinator/cosign/types/src/tests/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    global_session: [1u8; 32],
+    block_number: 5,
+    block_hash: BlockHash([5u8; 32]),
+    cosigner: ExternalNetworkId::Bitcoin,
+  };
+
+  let msg1 = cosign.signature_message();
+  let msg2 = cosign.signature_message();
+
+  assert_eq!(msg1, msg2, "signature_message should be deterministic");
```
kayabaNerve: We likely want to ensure the message isn't just deterministic yet comprehensive. For example, https://github.com/serai-dex/serai/blob/786ba87125ca9205e02bf74f29c49d0e28040a08/substrate/primitives/src/crypto/embedded_elliptic_curve_keys.rs#L464-L486.

coordinator/cosign/types/src/tests/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+    "test precondition: signature bytes should be invalid for schnorrkel"
+  );
+
+  let signed = SignedCosign { cosign, signature: invalid_sig_bytes };
+
+  let keypair = sr25519_fixture();
+  let valid_pubkey = Public(keypair.public.to_bytes());
+
+  assert!(!signed.verify_signature(valid_pubkey), "invalid signature bytes should return false");
+}
```
kayabaNerve, line 155: As these four tests are almost identical, it may be best to use a single test with multiple blocks for each assertion. For example, https://github.com/serai-dex/serai/blob/786ba87125ca9205e02bf74f29c49d0e28040a08/substrate/primitives/src/crypto/mod.rs#L134-L173 uses four blocks to test `RistrettoSignature`, `RistrettoSignature` as SCALE, `Signature`, and `Signature` as SCALE (though I have my own sin of not populating the assertion messages/leaving comments for which blocks test what).

coordinator/cosign/types/src/tests/mod.rs
```diff
@@ -0,0 +1,156 @@
...
+
+/// Creates a SignedCosign using the test fixture keypair (seed [0xff; 32])
+pub fn sign_cosign_with_fixture(cosign: Cosign) -> SignedCosign {
+  sign_cosign(cosign, &sr25519_fixture())
+}
+
+/// Creates a SignedCosign using a keypair derived from the given seed
+pub fn sign_cosign_with_seed(cosign: Cosign, seed: [u8; 32]) -> SignedCosign {
+  sign_cosign(cosign, &sr25519_fixture_from_seed(seed))
+}
```
kayabaNerve, line 25: All of these seem so small, I'm not actually sure they're better to have. I'll have to revisit here once I see how they're used by `coordinator-cosign`.

substrate/client/serai/src/lib.rs
```diff
@@ -228,7 +228,23 @@ impl Serai {
...
 impl Events {
+  /// Create an instance of Events
+  pub fn new() -> Self {
+    Events::default()
+  }
+
+  /// Create an instance of Events
+  pub fn with(events: Vec<Event>) -> Self {
+    Events { events: Arc::new(vec![events]) }
+  }
```
kayabaNerve: Hm. This wasn't public for a reason. It's not that it's unsafe per se, it's just not intended to be publicly consumed. I understand we're doing so with our tests however.

Could we solely offer `fn with` and mark it `doc(hidden)` so it isn't part of our API commitment? It'd either be that or a test flag, but that seems excessive.

coordinator/cosign/src/delay.rs
```diff
@@ -33,17 +44,34 @@ impl<D: Db> ContinuallyRan for CosignDelayTask<D> {
...
+        let now = now_timestamp();
+
+        // drop txn during sleep
+        drop(txn);
 
+        if time_valid > now {
+          // Sleep until then
+          let time_left = time_valid - now;
+          tokio::time::sleep(time_left).await;
+        }
```
kayabaNerve, line 93: Dropping the `txn` during the sleep is polite :)

coordinator/cosign/src/evaluator.rs
```diff
@@ -35,13 +42,16 @@ db_channel!(
...
     let existing = match CurrentlyEvaluatedGlobalSession::get(txn) {
       Some(existing) => existing,
       None => {
-        let first = GlobalSessionsChannel::try_recv(txn)
-          .expect("fetching latest global session yet none declared");
+        let first = GlobalSessionsChannel::try_recv(txn).ok_or_else(|| {
+          format!(
+            "fetching global session for block #{block_number} but none declared in channel yet"
+          )
+        })?;
```
kayabaNerve: This is condition number four for it won't panic unless... Was there a reason to use an error here?

common/log/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "serai-log"
+version = "0.1.0"
+description = "Coverage-gated logging macros for the Serai project"
+license = "MIT"
+repository = "https://github.com/serai-dex/serai/tree/develop/common/log"
+authors = ["Luke Parker <lukeparker5132@gmail.com>", "rafael_xmr <g.pi8ax@slmail.me>"]
```
kayabaNerve: You can identify as the sole author here, if you'd prefer :) I can add myself if/when I make changes. If you want to be clear about the people who be contacted though, it's fine to leave me.

tests/task/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "serai-test-task"
+version = "0.1.0"
+description = "Common test utilities for serai-task ContinuallyRan tasks"
+license = "MIT"
```
kayabaNerve: `serai-task` is `AGPL-3.0-only`, so this likely should be too

tests/task/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+repository = "https://github.com/serai-dex/serai/tree/develop/tests/task"
+authors = ["Luke Parker <lukeparker5132@gmail.com>", "rafael_xmr <g.pi8ax@slmail.me>"]
```
kayabaNerve: Same comment on the authors as `serai-log` :)

common/task/src/test_helpers.rs
```diff
@@ -0,0 +1,35 @@
...
+    serai_log::log::debug!("running task once: {}", core::any::type_name::<T>());
+    assert_eq!(task.run_iteration().await.unwrap(), made_progress);
+  }
+
+  /// Assert that a task iteration fails with an error containing the given string.
+  pub async fn task_runs_and_fails_with<T: ContinuallyRan>(task: &mut T, error: &str) {
+    serai_log::log::debug!("running task (expecting failure): {}", core::any::type_name::<T>());
+    let err = task.run_iteration().await.unwrap_err();
+    let err_str = format!("{err:?}");
+    assert!(err_str.contains(error), "{err_str}");
```
kayabaNerve, line 24: This is the exact issue with using strings for errors :/ If we do want to test these errors, it's likely time to move to structured errors (such as via `thiserror`).

tests/shim-rpc/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "serai-shim-rpc"
+version = "0.1.0"
+description = "A bespoke shim RPC node for testing Serai RPC clients without a real chain"
+license = "MIT"
+repository = "https://github.com/serai-dex/serai/tree/develop/tests/shim-rpc"
+authors = ["Luke Parker <lukeparker5132@gmail.com>", "rafael_xmr <g.pi8ax@slmail.me>"]
```
kayabaNerve, line 7: Same comment on the authors as `serai-log` :)

tests/shim-rpc/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "serai-shim-rpc"
+version = "0.1.0"
+description = "A bespoke shim RPC node for testing Serai RPC clients without a real chain"
+license = "MIT"
```
kayabaNerve: While this only uses MIT-licensed deps, internal testing tools should likely following the practice of using AGPL-3.0-only for anything which isn't directly intended for consumption by other projects (at least, not without them contributing back). Also, due to some functions inherited from `serai-node` to match its behavior (and my comment `serai-node` could be a dep).

tests/shim-rpc/src/rpc.rs
```diff
@@ -0,0 +1,265 @@
...
+  primitives::{
+    BlockHash,
+    network_id::{ExternalNetworkId, NetworkId},
+    validator_sets::{ExternalValidatorSet, Session},
+  },
+};
+
+use crate::state::SharedState;
+
+/// Typed RPC errors mirroring `substrate/node/src/rpc/utils.rs`.
```
kayabaNerve, line 18: In theory, we can add a `lib.rs` to `serai-node` which would allow using it as a library and possibly avoid this. Not something worth doing right now IMO, just a note.

tests/shim-rpc/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+
+  /// Access the underlying shared state directly.
+  pub fn state(&self) -> &SharedState {
+    &self.state
+  }
+
+  /// Stop the shim RPC node server.
+  pub fn stop(&self) {
+    self.handle.stop().expect("failed to stop shim RPC node");
+  }
```
kayabaNerve: I'd assume the handle does this on-`drop`, making an explicit `stop` redundant. Even if the explicit `stop` was necessary, _this_ should be done on `Drop`.

coordinator/cosign/src/lib.rs
```diff
@@ -175,38 +185,47 @@ impl IntakeCosignError {
...
+  pub fn new(db: D) -> Self {
+    Self { db, _task_handles: vec![] }
+  }
+
   /// Spawn the tasks to intend and evaluate cosigns.
   ///
   /// The database specified must only be used with a singular instance of the Serai network, and
   /// only used once at any given time.
-  pub fn spawn<R: RequestNotableCosigns>(
+  pub fn spawn<R: RequestNotableCosigns + Sync>(
```
kayabaNerve: We can introduce the `Sync` bound directly into `RNC`.

coordinator/cosign/src/lib.rs
```diff
@@ -223,12 +242,13 @@ impl<D: Db> Cosigning<D> {
     getter: &impl Get,
     block_number: u64,
   ) -> Result<Option<BlockHash>, Faulted> {
-    if block_number > Self::latest_cosigned_block_number(getter)? {
+    if block_number == 0 || block_number > Self::latest_cosigned_block_number(getter)? {
```
kayabaNerve: Please explicitly distinguish the two clauses of the conditional with parentheses.

coordinator/cosign/src/lib.rs
```diff
@@ -223,12 +242,13 @@ impl<D: Db> Cosigning<D> {
...
       return Ok(None);
```
kayabaNerve, line 287: Why do we report the genesis block _doesn't have a cosigned block hash_? As it's static and final, and block #0 does have a hash, we should be able to return the actual value _unless_ there's some funky internal side effect happening here, but that's not clear.

common/task/src/test_helpers.rs
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+
+/// Trait for test structs that can produce a [`ContinuallyRan`] task.
+pub trait IntoTask {
+  /// The task type produced by this test struct.
+  type Task: ContinuallyRan + 'static;
+  /// Create the task from the current test state.
+  fn into_task(&self) -> Self::Task;
```
kayabaNerve: You can express this as `IntoTask: Into<Self::Task>` and then use `into()` UNLESS:
1) You want to be very explicit about the exact `into` which is occuring
2) You explicitly DO NOT want Rust's type inference to be applied (as `into()` will infer what it is being converted into by how it's used, while `into_task` returns a fixed, defined type). That's a fine opinion, but I wanted to let you know in case you didn't think of it/because I'm unclear that's the opinion being expressed here vs an oversight.

coordinator/cosign/src/delay.rs
```diff
@@ -30,20 +43,43 @@ impl<D: Db> ContinuallyRan for CosignDelayTask<D> {
...
           break;
         };
+
+        let latest_cosigned_block_number = LatestCosignedBlockNumber::get(&mut txn).unwrap_or(0);
+
+        serai_log::debug!(
+          "beginning delay: block_number={block_number}, time_evaluated={time_evaluated}, latest_cosigned_block_number={latest_cosigned_block_number}",
+        );
+
+        if block_number <= latest_cosigned_block_number {
```
kayabaNerve, line 62: Is this defensive or could this happen? It's fine, but it'd be great to have a comment clarifying if this is known to potentially occur or not. Maybe also a `warn` or a `debug` if it actually occurs?

coordinator/cosign/src/delay.rs
```diff
@@ -30,20 +43,43 @@ impl<D: Db> ContinuallyRan for CosignDelayTask<D> {
...
-          .await;
+        let now_timestamp = now_timestamp().as_secs();
+        let time_valid_timestamp = time_evaluated + ACKNOWLEDGEMENT_DELAY.as_secs();
 
+        // drop txn during sleep
+        drop(txn);
+
+        if time_valid_timestamp > now_timestamp {
+          // Sleep until then
+          let time_left = time_valid_timestamp - now_timestamp;
```
kayabaNerve: Nit: `if let Some(time_left) = time_valid_timestamp.checked_sub(now_timestamp)`

The subtraction would panic _if not for the preceding `if`_. This would convert the two into a single statement which cannot panic and has the same effect.

coordinator/cosign/src/evaluator.rs
```diff
@@ -48,7 +56,8 @@ fn currently_evaluated_global_session_strict(
     };
     assert!(
       existing.1.start_block_number <= block_number,
-      "candidate's start block number exceeds our block number"
+      "candidate's start block number {:#?} exceeds our block number {block_number}",
```
kayabaNerve: Nit: `({...})` would be a bit clearer as "candidate's number 5" is not the same in English as "candidate's number (5)". The former is their item, "number 5", the latter is their number with the context it's `5`.

coordinator/cosign/src/evaluator.rs
```diff
@@ -73,36 +82,120 @@ pub(crate) fn currently_evaluated_global_session(getter: &impl Get) -> Option<[u
...
+  }
+
+  *last_request_for_cosigns = Instant::now();
+
+  true
+}
+
+//// Calculate the minimum threshold required for cosigning
+fn cosign_threshold(total_stake: u64) -> u64 {
+  ((total_stake * COSIGN_COMMIT_THRESHOLD) / 100) + 1
```
kayabaNerve: This was a bug in the original code but this formula panics when `total_stake >= u64::MAX.div_ceil(COSIGN_COMMIT_THRESHOLD)`.

`u64::try_from((u128::from(total_stake) * COSIGN_COMMIT_THRESHOLD_NUMERATOR) / COSIGN_COMMIT_THRESHOLD_DENOMINATOR).expect("threshold < 1") + 1`

would solve this. The definition of a denominator just follows how the numerator was made a constant, where `100` is only valid as the denominator _because_ of the way the numerator (now opaque behind the constant) is expressed.

coordinator/cosign/src/evaluator.rs
```diff
@@ -112,9 +205,10 @@ impl<D: Db, R: RequestNotableCosigns> ContinuallyRan for CosignEvaluatorTask<D,
           // supermajority of the prior block's validator sets
           HasEvents::Notable => {
             let mut weight_cosigned = 0;
+
             for set in global_session_info.sets {
               // Check if we have the cosign from this set
-              if NetworksLatestCosignedBlock::get(&txn, global_session, set.network)
+              if NetworksLatestCosignedBlock::get(&mut txn, global_session, set.network)
```
kayabaNerve: This doesn't have to be promoted to mutable AFAIK.

coordinator/cosign/src/evaluator.rs
```diff
@@ -73,36 +82,120 @@ pub(crate) fn currently_evaluated_global_session(getter: &impl Get) -> Option<[u
...
+) -> Result<(), String> {
+  if weight_cosigned >= cosign_threshold(total_stake) {
+    return Ok(());
+  }
+
+  if should_request_cosigns(last_request_for_cosigns) {
+    request
+      .request_notable_cosigns(global_session)
+      .await
+      .map_err(|e| format!("RPC error fetching notable cosigns: {e:?}"))?;
```
kayabaNerve: It isn't an "RPC error" but just an "error". This is piped to our own P2P network and other nodes on the network are queried over a gossip layer.

coordinator/cosign/src/evaluator.rs
```diff
@@ -125,23 +219,17 @@ impl<D: Db, R: RequestNotableCosigns> ContinuallyRan for CosignEvaluatorTask<D,
...
-                  .await
-                  .map_err(|e| format!("{e:?}"))?;
-              }
-              // We return an error so the delay before this task is run again increases
-              return Err(format!(
-                "notable block (#{block_number}) wasn't yet cosigned. this should resolve shortly",
-              ));
-            }
 
-            log::info!("marking notable block #{block_number} as cosigned");
```
kayabaNerve, line 144: It seems we lost this log

coordinator/cosign/src/evaluator.rs
```diff
@@ -157,52 +245,23 @@ impl<D: Db, R: RequestNotableCosigns> ContinuallyRan for CosignEvaluatorTask<D,
...
-                  self
-                    .request
-                    .request_notable_cosigns(global_session)
-                    .await
-                    .map_err(|e| format!("{e:?}"))?;
-                }
-                // We return an error so the delay before this task is run again increases
-                return Err(format!(
-                  "block (#{block_number}) wasn't yet cosigned. this should resolve shortly",
-                ));
```
kayabaNerve, line 204: We lost the comments in this block

coordinator/cosign/src/evaluator.rs
```diff
@@ -212,29 +271,19 @@ impl<D: Db, R: RequestNotableCosigns> ContinuallyRan for CosignEvaluatorTask<D,
               */
               known_cosign = lowest_common_block;
             }
-
-            log::debug!("marking non-notable block #{block_number} as cosigned");
```
kayabaNerve, line 216: We lost this log.

coordinator/cosign/src/evaluator.rs
```diff
@@ -212,29 +271,19 @@ impl<D: Db, R: RequestNotableCosigns> ContinuallyRan for CosignEvaluatorTask<D,
...
-              .as_secs(),
-          ),
-        );
+        CosignedBlocks::send(&mut txn, &(block_number, now_timestamp().as_secs()));
         txn.commit();
 
+        // Roughly ~1 hour, no need for repetitive logging
         if (block_number % 500) == 0 {
-          log::info!("marking block #{block_number} as cosigned");
+          serai_log::debug!("marking block #{block_number} as cosigned");
```
kayabaNerve: This should be `info`. It informs the operator every hour that this task is functioning. The use of `debug` for every single block is fair, as that debugs if a halt happened at a specific point.

coordinator/cosign/src/intend.rs
```diff
@@ -102,66 +76,127 @@ impl<D: Db> ContinuallyRan for CosignIntendTask<D> {
...
+        .latest_finalized_block_number()
+        .await
+        // Ephemeral RPC Err: task to re-run and continue trying
+        .map_err(|e| format!("RPC error fetching latest finalized block number: {e}"))?;
+
+      serai_log::debug!(
+        "beginning scan: start={start_scan_block_number}, latest={latest_serai_block_number}"
+      );
+
+      if latest_serai_block_number < start_scan_block_number {
```
kayabaNerve: This `if` check is completely redundant as the following loop only executed under this condition.

coordinator/cosign/src/intend.rs
```diff
@@ -102,66 +76,127 @@ impl<D: Db> ContinuallyRan for CosignIntendTask<D> {
...
+            // safe to skip this deallocation event
+            continue;
+          };
+
+          let existing = Stakes::get(&txn, network, *validator)
+            // critical panic:
+            // this is a critical issue and will not be solved after re-tries,
+            // missing Stakes from previous blocks will remain missing until re-indexed
+            // if encountered halt the process
+            .expect("unable to deallocate with no prior existing stake");
```
kayabaNerve: No, actually, but I see why you did this.

If a validator has never staked, they will have `0` allocated as stake.

If they then try to deallocate `0` stake, the amount they're deallocating is less than or equal to the amount allocated, so the system allows it.

It's silly, but it's something I've learned to do over the years. Specifically, it descends from rejecting transferring `0`, which is pointless. The reason _not_ to reject transferring `0` is because sometimes, you want to always issue a transfer on a regular basis (one a week), and some weeks, there may not be any value to transfer. When the regular operation occurs, it still occurs, it just transfers `0`. If it errored, it may bork the regular operation and screw up the caller.

With that mind, as my personal practice, I write code to never bork on any legitimate amount, even if a silly amount.

coordinator/cosign/src/intend.rs
```diff
@@ -102,66 +76,127 @@ impl<D: Db> ContinuallyRan for CosignIntendTask<D> {
...
+            unreachable!("event from `set_decided_events` wasn't `Event::SetDecided`")
           };
 
           let Ok(set) = ExternalValidatorSet::try_from(*set) else { continue };
+
+          if validators.is_empty() {
+            // Maybe ephemeral: event blocks from RPC returned empty set list
+            // could resolve after retry. or will get forever stuck.
+            Err(format!("validator set from Event::SetDecided was empty"))?;
+          }
```
kayabaNerve: This we can panic on as there never should be no validators within a set. They may have no stake, but never not exist at all.

kayabaNerve (commented):
This is all exceptionally clear and straighforward tests (and bug fixes) for the cosign module I can say I'm incredibly happy with. While I've left a littany of comments, they're largely just review nits and style as we work together moving forward. The dedicated shim RPC really seems to be the way to go for testing re: the Serai node's behavior, even if I'm not entirely sold on all of the designs of fixtures/abstractions here. Thankfully, because they're test files, I can accept they work and be happy with them in that regard. The usage of coverage also really seems to have been great at clearly establishing all the test cases, and I appreciated the fuzz test re: intend.

kayabaNerve:
Per https://github.com/serai-dex/serai/blob/next-polkadot-sdk/LICENSE.md (which this caused me to add), please update the `LICENSE` files for the new crates and attach the requested DCO :)

Also, should the utilities to test a task simply be part of `serai-task`? Just a stray thought I had.

coordinator/cosign/src/intend.rs
rafael-xmr: Yes but here it is checking for a previous _indexed_ Stakes object, so it's just a sanity check in the case the task did not succesfully index a stake any size it is, but I added tests here: 

https://github.com/rafael-xmr/serai/blob/108754e16a950bf7aa1372e296d0072502ed2cd9/coordinator/cosign/src/tests/intend.rs#L608-L644

To verify that given a stake event, and the existing **Stake** index, deallocate always works even if 0-denominated, or either stake and deallocate are 0.

But given you say

> If a validator has never staked, they will have 0 allocated as stake.

should the stake _just be considered 0_,when hitting this case on deallocate, even if not indexed in a db, and pass, or should it be **indexed as 0**, for every existing validator beforehand to then also allow this to pass?

kayabaNerve:
Ugh. When I merged `next` in, I merged my local `next` and not the upstream's `next`. That butchered the history a bit. I've corrected it by manually squash merging 95ea47d569b5f0dc446bfb2e0f76ec3dbc0aedaf into `next` and force pushing to `next`, but a bit of weirdness for anyone reviewing the exact Git history there...

PR #722, opened 2026-01-10
fix: remove duplicate docsrs cfg_attr lines

phessophissy:
Removes duplicate `#![cfg_attr(docsrs, feature(doc_cfg))]` lines in:
- `common/env/src/lib.rs`
- `substrate/in-instructions/pallet/src/lib.rs`

These appear to be copy-paste errors. The duplicate attribute has no effect but is unnecessary.

### DIFF
--- a/common/env/src/lib.rs
+++ b/common/env/src/lib.rs
@@ -1,5 +1,4 @@
 #![cfg_attr(docsrs, feature(doc_cfg))]
-#![cfg_attr(docsrs, feature(doc_cfg))]
 
 // Obtain a variable from the Serai environment/secret store.
 pub fn var(variable: &str) -> Option<String> {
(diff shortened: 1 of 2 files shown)


kayabaNerve:
ACK, but FYI, `develop` isn't necessarily a maintenance target. Current HEAD would be within `next-polkadot-sdk` which should be collapsed to the `next` branch in a next month or so. The awkward structure is just due to certain branches offering certain features where the other branches have yet to replace _all_ features within `develop`.

Thank you for the linting however!

phessophissy:
> ACK, but FYI, `develop` isn't necessarily a maintenance target. Current HEAD would be within `next-polkadot-sdk` which should be collapsed to the `next` branch in a next month or so. The awkward structure is just due to certain branches offering certain features where the other branches have yet to replace _all_ features within `develop`.
> 
> Thank you for the linting however!

Oh, I see. Thanks for the clarification, that makes sense. Appreciate the context around the branch structure and roadmap, and glad the linting feedback was helpful.

kayabaNerve:
Feel free to ask any questions moving forward! Best place is our Matrix/Discord (which are bridged, so it's solely about how you want to connect) from the README :)

PR #725, opened 2026-01-14
fix(bitcoin-serai): include OP_RETURN output in weight calculation

### LINKED ISSUE #691: `bitcoin-serai` miscalculates weight when there's an OP_RETURN output present
This technically can lead to breaking the maximum weight established by policy.

EricGrill:
## Summary
- Fixes weight calculation in `bitcoin-serai` to include OP_RETURN outputs when computing transaction weight/vbytes
- Previously, the `calculate_weight_vbytes` function did not account for OP_RETURN data, causing underestimation of transaction weight
- This could lead to transactions exceeding `MAX_STANDARD_TX_WEIGHT` when OP_RETURN data was present

## Changes
- Modified `calculate_weight_vbytes` to accept an optional `data` parameter
- Added the OP_RETURN output to the weight calculation transaction when data is present
- Restructured the code to calculate weight before consuming the data ownership
- Added regression test `test_op_return_weight_calculation` to verify correct fee calculation with OP_RETURN

## Test plan
- [x] Code compiles with `cargo check`
- [x] Added `test_op_return_weight_calculation` test that verifies:
  - Fee calculation accounts for OP_RETURN output
  - Transaction weight stays within policy limits
- [ ] Run full test suite with `cargo test` (requires bitcoind for integration tests)

Closes #691

---
Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/networks/bitcoin/src/wallet/send.rs
+++ b/networks/bitcoin/src/wallet/send.rs
@@ -63,6 +63,7 @@ impl SignableTransaction {
     inputs: usize,
     payments: &[(ScriptBuf, u64)],
     change: Option<&ScriptBuf>,
+    data: Option<&[u8]>,
   ) -> (u64, u64) {
     // Expand this a full transaction in order to use the bitcoin library's weight function
     let mut tx = Transaction {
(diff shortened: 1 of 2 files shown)


kayabaNerve:
Thank you for the interest and for declaring your usage of an LLM!

I don't believe we can accept this at this time however. Claude is not an open model and lacks attribution for its weights. Its unclear what code has been transformed into the contribution here and what licenses were intended. Even if this can be legally considered fair use in the relevant jurisdictions (itself a question), I don't believe it's respectful to the original authors who enabled such models via their work being scraped.

While I plan to close this PR, if you need this issue fixed as a user (and not solely as a volunteer looking to contribute), please let me know. Of course, you may always open a PR which wasn't performed via usage of unattributed origins if you still want to fix this yourself in Serai's upstream, or use your fork with your LLM-derived patches.

PR #732, opened 2026-01-20
Ensure start_block is finalized in IndexTask::new and add tests

j-berman:
The main implementation change in here is in `IndexTask::new`. Without this change, it may be setting a start_bock that is not actually finalized according to the ScannerFeed logic. This change makes sure the provided start_block is actually finalized in the chain before setting it as "finalized" in the db.

In order to test the IndexTask implementation, I did some restructuring:

1) Moved ScannerFeed into processor primitives.
2) Exposed the Monero ScannerFeed impl in serai-monero-processor.

This enabled me to write a test directly in the IndexTask definition file, which I think is a safe/sane approach to make sure it's behaving as expected. It also improves code structure & organization. E.g. the monero processor now does not depend on scanner, and instead implements the ScannerFeed trait that it gets from the serai-processor-primitives crate. This cleanly separates impl from interface.

______

The new IndexTask tests are very close to chain agnostic, and would be very simple to update for each chain. Their only dependency is that they require a live daemon to be running, which appears to be the case for XMR and BTC CI currently. I figure this is saner than requiring the entire suite of Serai containers to be running in order for these more isolated unit tests to run.

I imagine there may be an alternative desired approach, but this got me off the ground for now.

______

If this approach looks good, I can continue task-by-task in similar fashion, and extend the tests for each chain.

### DIFF
--- a/processor/scanner/Cargo.toml
+++ b/processor/scanner/Cargo.toml
@@ -38,3 +38,16 @@ serai-primitives = { path = "../../substrate/primitives", default-features = fal
 
 primitives = { package = "serai-processor-primitives", path = "../primitives" }
 scheduler-primitives = { package = "serai-processor-scheduler-primitives", path = "../scheduler/primitives" }
+
+[dev-dependencies]
+tempfile = "3"
+serai-db = { path = "../../common/db", features = ["rocksdb"] }
+serai-monero-processor = { path = "../monero", features = ["rocksdb"] }
+
+monero-simple-request-rpc = { git = "https://github.com/monero-oxide/monero-oxide", rev = "cae84b7b16051ac8bf92ed151aad49b9e6f4fcae", default-features = false }
+monero-wallet = { git = "https://github.com/monero-oxide/monero-oxide", rev = "cae84b7b16051ac8bf92ed151aad49b9e6f4fcae", default-features = false, features = ["std"] }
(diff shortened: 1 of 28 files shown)


processor/scanner/Cargo.toml
```diff
@@ -38,3 +38,16 @@ serai-primitives = { path = "../../substrate/primitives", default-features = fal
...
+serai-db = { path = "../../common/db", features = ["rocksdb"] }
+serai-monero-processor = { path = "../monero", features = ["rocksdb"] }
+
+monero-simple-request-rpc = { git = "https://github.com/monero-oxide/monero-oxide", rev = "cae84b7b16051ac8bf92ed151aad49b9e6f4fcae", default-features = false }
+monero-wallet = { git = "https://github.com/monero-oxide/monero-oxide", rev = "cae84b7b16051ac8bf92ed151aad49b9e6f4fcae", default-features = false, features = ["std"] }
+monero-ed25519 = { git = "https://github.com/monero-oxide/monero-oxide", rev = "cae84b7b16051ac8bf92ed151aad49b9e6f4fcae", version = "0.1.0", default-features = false, features = ["std"] }
+
+zeroize = { version = "1", default-features = false, features = ["std"] }
+
+serial_test = "*"
```
kayabaNerve: I'd prefer we didn't add an external crate for this when we already have a handwritten shim somewhere in tree. I'd also asked if this is required due to the overlap on a single external daemon instead of provisioning with Docker.

If we do add an external crate, we definitely shouldn't accept any version of it. We should know which one works for us and then not blindly accept breaking changes (this will accept downgrading from 1.0.0 to 0.2.1 and then upgrading to 3.8.5).

PR #740, opened 2026-02-02
fix build process for M1 MacOS

akildemir:
I was getting bunch of errors when I try to build the `next-polkadot-sdk` branch on M1 macOS Tahoe 26.2 with `cargo build --all-features` command. This PR fixes couple of issues on the runtime  build  script that causes build to fail. This is only tested on said operating system. Its effect on other operating systems yet to be tested.  Here what it fixes;

1. Moved setting the `PATH` env variable directly within `command` function instead of only setting it for `build_command` since other commands also need the `PATH` variable. Specifically, `tree_command` also needed to find the `rustc`.
2. Include  `RUSTUP_HOME` and `RUSTUP_TOOLCHAIN` in the env variables we are setting so that it is still resolved after we create fresh `CARGO_HOME`. 
3. Put `rustflags` command inside a block to properly apply the `cfg` attribute.
4. Import `alloc::vec::Vec` to `signals/benchmarking`

### DIFF
--- a/orchestration/src/main.rs
+++ b/orchestration/src/main.rs
@@ -304,7 +304,9 @@ fn build_serai_service(
 
     // LLVM's MemTagSanitizer
     #[cfg(target_feature = "mte")]
-    rustflags += " -C target-feature=+mte -Z sanitizer=memtag";
+    {
+      rustflags += " -C target-feature=+mte -Z sanitizer=memtag";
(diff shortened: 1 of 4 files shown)


kayabaNerve:
Closing as this was manually merged in via the above commit ^ Thank you!

PR #767, opened 2026-03-19
Enable various CIs for `next-polkadot-sdk`

kayabaNerve:
It definitely is time to re-enable these. Some likely will fail out of the box, such as due to bit rot, due to `machete` being having a couple outstanding items, and due to MSRV declarations not having been maintained for months.


kayabaNerve:
`macos-26-intel` seems to have some spurious failures, but all the CIs this enables now pass :tada: I'll merge this accordingly.

PR #778, opened 2026-03-31
Extend runtime RPC

akildemir:
This PR adds various RPC endpoints to serai node and support for them to  serai client. Tried to follow existing RPC design but happy to update endpoint, function and/or parameter names as well as the documentation for the said functions as well.

-----------

Version 1.1

    indicated in the file; or

    in the file; or

    it.

contribution.

-----------

### DIFF
--- a/substrate/node/src/rpc/core.rs
+++ b/substrate/node/src/rpc/core.rs
@@ -5,7 +5,7 @@ pub(crate) fn module(
 ) -> Result<RpcModule<impl 'static + Send + Sync>, Box<dyn std::error::Error + Send + Sync>> {
   let mut module = RpcModule::new(client);
 
-  module.register_method("system/next-nonce", |params, client, _ext| -> Result<_, Error> {
+  module.register_method("core/next-nonce", |params, client, _ext| -> Result<_, Error> {
     let Some(block_hash) = block_hash(client, &params)? else { Err(Error::InvalidStateReference)? };
 
     #[derive(sp_core::serde::Deserialize)]
--- a/substrate/node/src/rpc/dex.rs
+++ b/substrate/node/src/rpc/dex.rs
@@ -1,0 +1,22 @@
+use super::*;
+
+pub(crate) fn module(
+  client: Arc<FullClient>,
+) -> Result<RpcModule<impl 'static + Send + Sync>, Box<dyn std::error::Error + Send + Sync>> {
+  let mut module = RpcModule::new(client);
+
+  module.register_method("dex/pool-reserves", |params, client, _ext| -> Result<_, Error> {
+    let Some(block_hash) = block_hash(client, &params)? else { Err(Error::InvalidStateReference)? };
+
+    let coin = coin(&params)?;
+    let external_coin: ExternalCoin =
+      coin.try_into().map_err(|()| Error::InvalidRequest("coin is not external coin"))?;
+
+    let Ok(reserves) = client.runtime_api().pool_reserves(block_hash, external_coin) else {
+      Err(Error::Internal("couldn't fetch the reserves for the coin"))?
+    };
+    Ok(hex::encode(borsh::to_vec(&reserves).unwrap()))
+  })?;
+
+  Ok(module)
+}
--- a/substrate/node/src/rpc/dex.rs
+++ b/substrate/node/src/rpc/dex.rs
@@ -12,10 +12,24 @@ pub(crate) fn module(
     let external_coin: ExternalCoin =
       coin.try_into().map_err(|()| Error::InvalidRequest("coin is not external coin"))?;
 
-    let Ok(reserves) = client.runtime_api().pool_reserves(block_hash, external_coin) else {
-      Err(Error::Internal("couldn't fetch the reserves for the coin"))?
+    let address = serai_abi::dex::address(external_coin);
+    let Ok(sri_balance) = client.runtime_api().balance(block_hash, address, Coin::Serai) else {
+      Err(Error::Internal("couldn't fetch the sri balance of the pool"))?
     };
-    Ok(hex::encode(borsh::to_vec(&reserves).unwrap()))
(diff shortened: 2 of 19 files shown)


substrate/runtime/src/wasm/mod.rs
```diff
@@ -506,6 +508,42 @@ sp_api::impl_runtime_apis! {
...
+    }
+    fn genesis_liquidity_supply(coin: ExternalCoin) -> Amount {
+      GenesisLiquidityTokens::supply(coin)
+    }
+
+    fn genesis_completed() -> bool {
+      GenesisLiquidity::completed()
+    }
+
+    fn pool_reserves(coin: ExternalCoin) -> Reserves {
```
kayabaNerve: This _could_ be trivially moved to the RPC, as it's just fetching two balances. That wouldn't require the RPC peek under the hood and would keep this slimmer.

substrate/node/src/rpc/mod.rs
```diff
@@ -22,8 +22,13 @@ use sc_network::config::MultiaddrWithPeerId;
 mod utils;
 use utils::*;
 
+mod dex;
+mod system;
```
kayabaNerve: `core`

substrate/node/src/rpc/core.rs
```diff
@@ -0,0 +1,30 @@
...
+    #[serde(crate = "sp_core::serde")]
+    struct Address {
+      address: String,
+    }
+    let Ok(address) = params.parse::<Address>() else {
+      Err(Error::InvalidRequest(r#"missing `string` "address" field"#))?
+    };
+    let Ok(address) = SeraiAddress::from_str(&address.address) else {
+      Err(Error::InvalidRequest(r#"invalid serai address"#))?
+    };
```
kayabaNerve, line 21: I think we can deserialize from a string directly into a `SeraiAddress`, though that may require always considering the address as a string from the PoV of `serde` (when in general, it's considered a 32-byte array).

Since Serai doesn't use `serde`, maybe it'd be better to overfit the `serde` derivations for JSON and consider most of our items as such strings. That'd simplify a good bit of our RPC.

(this doesn't have to be addressed now, it's just thoughts on the topic)

substrate/node/src/rpc/dex.rs
```diff
@@ -0,0 +1,22 @@
...
+    let Some(block_hash) = block_hash(client, &params)? else { Err(Error::InvalidStateReference)? };
+
+    let coin = coin(&params)?;
+    let external_coin: ExternalCoin =
+      coin.try_into().map_err(|()| Error::InvalidRequest("coin is not external coin"))?;
+
+    let Ok(reserves) = client.runtime_api().pool_reserves(block_hash, external_coin) else {
+      Err(Error::Internal("couldn't fetch the reserves for the coin"))?
+    };
+    Ok(hex::encode(borsh::to_vec(&reserves).unwrap()))
```
kayabaNerve: Do we want to return hex-encoded amounts or a JSON `{ "sri": 0, "sriXYZ": 0 }`, where "sriXYZ" is filled in with the associated coin (`Coin::serai_symbol`?)?

substrate/node/src/rpc/utils.rs
```diff
@@ -87,6 +87,31 @@ pub(super) fn network(params: &Params) -> Result<NetworkId, Error> {
   network_from_str(network.network)
 }
 
+pub(super) fn coin_from_str(coin: impl AsRef<str>) -> Result<Coin, Error> {
+  Ok(match coin.as_ref().to_uppercase().as_str() {
+    "SRI" => Coin::Serai,
+    "BTC" => Coin::External(ExternalCoin::Bitcoin),
+    "ETH" => Coin::External(ExternalCoin::Ether),
+    "DAI" => Coin::External(ExternalCoin::Dai),
+    "XMR" => Coin::External(ExternalCoin::Monero),
```
kayabaNerve: These should be the Serai symbols ("sriBTC", "sriETH"...).

kayabaNerve (approved):
Overall, LGTM :) There's probably some nits on naming but for now, I've only left comments on anything worth commenting about.

I'm holding off on merging till after the current audit resolves, which also gives time to respond to any comments you want to. I think the only explicit blocker would be the Serai symbols. Rest are fine for now.

Also, please add the [DCO statement and other declaration](https://github.com/serai-dex/serai/blob/next-polkadot-sdk/LICENSE.md#contributions).

substrate/node/src/rpc/mod.rs
akildemir: initially named it "core" but it is conflicting with rust "core" library which imports FromStr trait. Should we use std::str::FromStr instead of core::str::FromStr in this file?

substrate/node/src/rpc/dex.rs
akildemir: I was thinking the same thing. Went with this since when we return a type we use this approach in validator-sets and that would be consistent. So I would say being consistent is for returns from RPC is best. So we would always either return hex-encoded bytes - that client deserializes - or string encoded json.

substrate/node/src/rpc/mod.rs
kayabaNerve: Ha, sorry, I didn't consider that.

`::core::str::FromStr` should be the unambiguous way to refer to FromStr` within Rust's `core` when we have `mod core`. If any issue still exists, `serai_core` or `core_pallet` work as the name for the module. I objected to `system` as that's something else entirely 😅

substrate/node/src/rpc/dex.rs
kayabaNerve: I agree we should be consistent, which here is presumably best if we go all-in on JSON and don't yield `borsh`-encoded blobs with JSON.

kayabaNerve:
Thank you!

PR #783, opened 2026-04-09
coordinator/tributary tests

rafael-xmr:
### 1) [bug] Identified in `impl ReadWrite for Transaction`:

Replaced the usage of:

```rust
borsh::from_reader(reader)
```

for:

```rust
borsh::BorshDeserialize::deserialize_reader(reader)
```

And added the tests `sequential_reads_from_shared_reader` and `borsh_from_reader_rejects_shared_reader_with_trailing_bytes` to verify the behavior, as using the previous method would fail on the scenario of multiple transactions on reader.

### 1) [feat] Added an assertion for the commented case:

Here:

```rust
// This function will only be called once for a (validator, topic) tuple due to how we handle
// nonces on transactions (deterministically to the topic)
assert!(
  txn.get(Accumulated::<D>::key(set, topic, validator)).is_none(),
  "accumulate called twice for the same (validator, topic) tuple: \
   the nonce system should have prevented this"
);
```

Not entirely necessary just defensive.

### 2) [bug] Bad usage of ::get(): replaced by ::key()

Found a bug which, given the nature of the accumulate function using a Generic type D for the data, when an Accumulate entry is added for a Preprocess ([u8; 64]) type, later a type of Share ([u8; 32]) would panic with a "Not all bytes read" error trying to deserialize the DB entry for the preceding topic's type. This is fixed by not trying to `get` the `Accumulated` entry with D as a type to deserialize, but rather getting its preceding topic `key` and using `txn.get()` instead.

From:

```rust
if Accumulated::<D>::get(txn, set, preceding_topic, validator).is_none() {
```

To:

```rust
if txn.get(Accumulated::<D>::key(set, preceding_topic, validator)).is_none() {
```

### 3) [bug] attempt * BASE_REATTEMPT_DELAY overflow:

From:

```rust
let blocks_till_reattempt = u64::from(attempt * BASE_REATTEMPT_DELAY);
```

To:

```rust
let blocks_till_reattempt = u64::from(attempt) * u64::from(BASE_REATTEMPT_DELAY);
```

Before, a too high attempt * BASE_REATTEMPT_DELAY could overflow the u32::max, now convert each value to u64 first, before doing the operation which avoids the u32 overflow.

### 1) [bug] panic on slash_report's added with a value of 0

By not filtering the 0 points reports, `slash_report` panics when expected to be less or equal than `f`

```rust
            let mut slash_report = vec![];
            for points in amortized_slash_report {
              // TODO: Natively store this as a `Slash`
              if points == u32::MAX {
                slash_report.push(Slash::Fatal);
              } else {
                slash_report.push(Slash::Points(points));
              }
            }
            assert!(slash_report.len() <= f);
```

The fix is:

```rust
              } else if points > 0 {
                slash_report.push(Slash::Points(points));
```
---

```
Version 1.1

    indicated in the file; or

    in the file; or

    it.

```

```
contribution.
```


kayabaNerve:
FYI, some of the above CI failures are because they're misconfigured. For anything `on: push: branches: - next`, the CI should pass and should continue to pass however. For the current HEAD of `next`, https://github.com/serai-dex/serai/commit/970f86d9f245ab803bda81b50ba955b11e78ccea, 22 were triggered and 22 passed (where some more may have triggered if the PR modified more files).

With this PR,
- Lint/[fmt, clippy, machete] fail
- Lint/vet fails because the `Cargo.lock` is dirty (and may independently fail even if it wasn't dirty)

I don't assume this PR is final, so no worries on the failures, I just wanted to clarify which lights are expected to be red and which lights should be green. If any CIs are being a problem, feel free to poke me :) For new crates, there's some basic definitions which have to be added for it to work as expected, and now, even for updates to/new dependencies, `cargo vet` will need to be poked (presumably by me).

kayabaNerve:
That's disgusting re: `borsh::from_reader`. It isn't even documented behavior. Can we add it to the `clippy.toml`'s disallowed methods for the footgun of requiring encapsulation, please? Though I'll note we do have the opposite DoS concern, where someone submits valid data with arbitrary junk after it, and once we have read everything, we do need to check the reader is _then_ empty ourselves. I don't think achieving that by adopting `borsh::from_reader` is the correct solution at this time as it's unreliable and we likely need a more encompassing solution. A reader over a slice that on drop, checks if it was empty or not, could be interesting if there was a way to handle an error on drop...

Defensive assertions are good. I have a note the processor really should be flush with them.

I have yet to review the PR itself, these are just a pair of comments on the summary :)

PR #785, opened 2026-04-20
Update contracts re: Prague, Osaka

kayabaNerve:
The router is updated to handle `floor_gas`, which is the only non-trivial change that affects us.

This adopts `solc 0.8.34`, which is old enough to be considered sufficiently tested, and which includes a bug fix the authors deemed important. The non-interface contracts now explicitly specify this version as required, not allowing compilation with any other version.

`build-solidity-contracts` is updated to specify the EVM it targets.

`slither-analyzer` has its version bumped, as does `svm-rs` which had a release cut but didn't post a release via GitHub, hence why it wasn't updated to earlier.

### DIFF
--- a/.github/actions/build-dependencies/action.yml
+++ b/.github/actions/build-dependencies/action.yml
@@ -53,18 +53,18 @@ runs:
     - name: Install solc
       shell: bash
       run: |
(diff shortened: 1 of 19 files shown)


kayabaNerve:
Manually merged into `next` after confirming the relevant CIs passed.

PR #1255, opened 2026-07-25
Perform MSRV checks against a minimal lockfile

### LINKED ISSUE #1246: MSRVs are over-stated
They're dependent on our current `Cargo.lock`, not a theoretical satisfying set of dependencies.

kayabaNerve:
Fixes #1246.


kayabaNerve:
This should be ready, barring two caveats:
1) I updated `monero-oxide` to https://github.com/monero-oxide/monero-oxide/pull/190, which isn't merged
2) I didn't update the `fcmp++` libraries, and we have some patches which should be upstreamed to `generalized-bulletproofs`
3) The `fcmp++` branch uses the old `multiexp` and should be updated to the new one, causing us to have both in our tree

kayabaNerve:
CI passes :tada: The only concern is if we should 'correct' the `monero-oxide` commits updated to. That'd also allow removing some bounds which were specified here, but have now been specified upstream.

Also, on the next release of `hyper-util`, we should be able to remove our `tracing` bound in `simple-request`.

kayabaNerve:
The `monero-oxide` crates now use the published releases. Its solely the FCMP++ libraries, which don't have releases, we need their commits updated.

